PR #66, opened 2022-02-14
More verbose error messages

FedericoV:
Makes the error messages a bit more verbose in a few places, with information about types and values.

### DIFF
--- a/diffrax/integrate.py
+++ b/diffrax/integrate.py
@@ -599,7 +599,8 @@ def diffeqsolve(
 
     # Error checking
     if dt0 is not None:
-        error_if(lambda: (t1 - t0) * dt0 <= 0, "Must have (t1 - t0) * dt0 > 0")
+        msg = f"Must have (t1 ({t1, type(t1)}) - t0  ({t0, type(t0)})) * dt0  ({dt0, type(dt0)})> 0"
+        error_if(lambda: (t1 - t0) * dt0 <= 0, msg)
 
     # Error checking
     term_leaves, term_structure = jax.tree_flatten(
--- a/diffrax/integrate.py
+++ b/diffrax/integrate.py
@@ -599,10 +599,12 @@ def diffeqsolve(
 
     # Error checking
     if dt0 is not None:
-        msg = ("Must have (t1 - t0) * dt0 > 0, we instead got: "
-        f"t1, with value: {t1} and type {type(t1)}, "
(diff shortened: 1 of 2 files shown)


diffrax/integrate.py
```diff
@@ -599,7 +599,8 @@ def diffeqsolve(
 
     # Error checking
     if dt0 is not None:
-        error_if(lambda: (t1 - t0) * dt0 <= 0, "Must have (t1 - t0) * dt0 > 0")
+        msg = f"Must have (t1 ({t1, type(t1)}) - t0  ({t0, type(t0)})) * dt0  ({dt0, type(dt0)})> 0"
```
patrick-kidger, line 602: So I find this a little hard to read. Can the `f"{}"` values be collected at the end of the error message?
When printing out a type, can we use plain English to denote what's going on, e.g. `Got value {t1} of type {type(t1)}.`

diffrax/misc/errors.py
```diff
@@ -57,7 +57,7 @@ def raises(_arg):
         raises((pred, index))
     else:
         msg = (
-            "`pred` must either be a `bool`, a NumPy array, a JAX array, or a "
-            "zero-argument callable that returns a `bool` or JAX array."
+            f"`pred` {pred} must either be a `bool`, a JAX array, or a zero-argument callable "
+            f"that returns a `bool` or JAX array, instead of {type(pred)}."
```
patrick-kidger, line 61: I'd don't hold to this as a hard-and-fast rule, but for something like this where an error here can almost only ever occur due to something totally crazy: I'd recommend putting the `f"{}"` values at the end of the docstring. That way if they're something with a huge `__str__` they won't blat the rest of the error message that comes after them.

patrick-kidger (commented):
I apologise for doing [this](https://twitter.com/iamdevloper/status/397664295875805184) ;)

diffrax/misc/errors.py
FedericoV: All good :)

Let me know if this looks better.

patrick-kidger:
LGTM! I'll merge once the tests pass and we'll fold this into the 0.0.3 release.

patrick-kidger:
...*if* the tests pass, haha.

It looks you haven't set up the pre-commit hooks. See [CONTRIBUTING.md](https://github.com/patrick-kidger/diffrax/blob/main/CONTRIBUTING.md).

PR #91, opened 2022-04-06
[JAX] Add MHLO lowerings in preparation for xla.lower_fun() removal

hawkinsp:
JAX is transitioning from building XLA HLO to MLIR MHLO. As part of
that work, xla.lower_fun() will be removed in a subsequent JAX release.
This PR adds direct MHLO lowerings using its cousin mhlo.lower_fun()
that should continue to work after xla.lower_fun() is removed.

### DIFF
--- a/diffrax/misc/ad.py
+++ b/diffrax/misc/ad.py
@@ -4,6 +4,7 @@ import equinox as eqx
 import jax
 import jax.interpreters.ad as ad
 import jax.interpreters.batching as batching
+import jax.interpreters.mlir as mlir
 import jax.interpreters.xla as xla
 import jax.lax as lax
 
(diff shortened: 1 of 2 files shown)


patrick-kidger:
Nice! Thank you, I appreciate you downstreaming this.
(Ignore the test failure, this is a known issue recently introduced into Black: https://github.com/psf/black/issues/2964)

What's the timeline for dropping XLA? (i.e. is it worth making a new release for this change?)
Separately: I've not completely kept track of what's going on in the XLA->MLIR transition. Is XLA being dropped from JAX entirely? Or even larger, the entire XLA project being deprecated in favour of MLIR? (If so, does this mean we get backpropagate-able while loops? :D )

hawkinsp:
@patrick-kidger The change here is that JAX will generate the MHLO compiler intermediate-language as its IR. It will still use XLA to compile that IR for the foreseeable future, although the switch does make it easier to plug in other compilers if that's interesting (e.g., there's a prototype IREE integration).

I would expect that we will remove the function `xla.lower_fun` in the next JAX release or so. (Note it's an internal API and so not subject to the deprecation policy.)

patrick-kidger:
Sounds good, thanks! No worries about the lack of normal deprecation.

In going through my other code to prepare it for this change: I notice that `jax.experimental.host_callback` is still using `xla` rather than `mlir`. Will this be updated? (a) I use this extensively (in the expected way), and hope that this subpackage won't be culled as a result of this change; (b) I have some code explicitly working with its XLA translation rule that I'd like to update to use the MLIR lowering rule instead.

hawkinsp:
Things in the `jax.interpreters.xla` package will mostly go away or be moved *eventually*, but right now I'm just trying to remove uses of `xla.lower_fun` and its cousins. This will allow me to start removing XLA translation rules where we have MHLO equivalents rather than having lots of duplicated logic. I wouldn't too much worry about migrating anything else yet.

patrick-kidger:
Sounds good! Thanks a lot for your help.

PR #99, opened 2022-05-02
switch error_if to python if (regarding google/jax/issues/10047)

amir-saadat:
Hey @patrick-kidger,

This is causing some failures in the internal tests and switching the if statement fixes them. Does this make sense, or did you have other intentions to keep this `error_if`?
This issue is relavant to[ your recent changes](https://github.com/patrick-kidger/diffrax/commit/23a1c199ac5e50d86a500cddd612b3b06ab45c50)  and also [google/jax/issues/10047](https://github.com/google/jax/issues/10047).

Thank you

### DIFF
--- a/diffrax/nonlinear_solver/newton.py
+++ b/diffrax/nonlinear_solver/newton.py
@@ -8,7 +8,7 @@ import jax.numpy as jnp
 import jax.scipy as jsp
 
 from ..custom_types import Bool, Int, PyTree, Scalar
-from ..misc import error_if, rms_norm
+from ..misc import rms_norm
 from ..solution import RESULTS
 from .base import AbstractNonlinearSolver, LU_Jacobian, NonlinearSolution
 
@@ -81,8 +81,8 @@ class NewtonNonlinearSolver(AbstractNonlinearSolver):
     tolerate_nonconvergence: bool = False
 
     def __post_init__(self):
-        if self.max_steps is not None:
-            error_if(self.max_steps < 2, "max_steps must be at least 2.")
+        if self.max_steps is not None and self.max_steps < 2:
+            raise ValueError("max_steps must be at least 2.")
 
     def _solve(
         self,


diffrax/nonlinear_solver/newton.py
```diff
@@ -82,7 +82,8 @@ class NewtonNonlinearSolver(AbstractNonlinearSolver):
 
     def __post_init__(self):
         if self.max_steps is not None:
-            error_if(self.max_steps < 2, "max_steps must be at least 2.")
+            if self.max_steps < 2:
```
patrick-kidger, line 85: These two `if` statements can be combined.

patrick-kidger (commented):
So this `error_if` was deliberately left as-is on the basis that this would allow `max_steps` to be a `Tracer`; in contrast your PR will instead raise a concretisation error. We could obviously imagine some more complicated machinery that keeps the `error_if` *and* avoids using JAX functionality at import time, but that's a bit of a faff.

Until someone has a use-case otherwise, and if (over both Diffrax and Equinox) this is the sole case of "JAX functionality used at import time", then I'm inclined to just accept this PR. Can you confirm that this really is the only case?

Do you have a way to test this? Without a test, I can't promise this won't silently regress in future versions.

If it'd be convenient for this change to have a corresponding version bump, then feel free to update this PR to bump the patch number.

diffrax/nonlinear_solver/newton.py
mahdi-shafiei: Thanks Patrick. We were thinking that you wanted to keep `max_steps` being a `Tracer` value and that makes sense. I can verify though that this PR solves tests breaking for our usecase.

amir-saadat:
Yes, I agree with you, maybe we can switch back to `error_if` when we have a practical case for it. We couldn't think of any but figured that might be your intention. Thanks for the quick reply.

FedericoV:
Hey Patrick:

re: test to catch regression.  I think that might have to wait until the issue Amir referenced (https://github.com/google/jax/issues/10047) is solved.

Another thing about error_if and plumbing mechanism to catch errors.  It's still experimental, but, have you seen the new checkify functionality in jax.experimental (https://jax.readthedocs.io/en/latest/_autosummary/jax.experimental.checkify.check.html) - eventually once the API / functionality is stabilized I expect that to be the standard way to plumb errors through JAX code.

diffrax/nonlinear_solver/newton.py
```diff
@@ -81,8 +81,8 @@ class NewtonNonlinearSolver(AbstractNonlinearSolver):
     tolerate_nonconvergence: bool = False
 
     def __post_init__(self):
-        if self.max_steps is not None:
-            error_if(self.max_steps < 2, "max_steps must be at least 2.")
+        if self.max_steps and self.max_steps < 2:
```
patrick-kidger, line 84: This should be `if self.max_steps is not None and self.max_steps < 2`. With this PR, `max_steps==0` will not raise an error when it should.

To avoid exactly these kinds of errors, I recommend avoiding implicit casts to bool.

patrick-kidger (commented):
Tests: I figured that would be your answer.

Checkify: yep, I've seen this. Once it's feature-complete -- in particular once it supports being staged out -- then I'd be happy to revisit the mechanism for raising errors.

diffrax/nonlinear_solver/newton.py
amir-saadat: sounds good. applied.

patrick-kidger:
Checks have failed - please follow [CONTRIBUTING.md](https://github.com/patrick-kidger/diffrax/blob/main/CONTRIBUTING.md).

patrick-kidger:
And merged. Thanks for the contribution.

Issue #120, opened 2022-06-29
Should `ts` in diffrax.StepTo be inclusive of `t0` and `t1`?


jatentaki:
I suppose not but it may be better to ask :)

patrick-kidger:
`StepTo(ts=...)` should/must include `t0` and `t1` as passed to `diffeqsolve`.

(The error raised if you don't is here, incidentally:)

https://github.com/patrick-kidger/diffrax/blob/ebfa6a82b884f6f188e24f37a396d05bc3e06daa/diffrax/step_size_controller/constant.py#L122

PR #121, opened 2022-06-30
Clarify in docs that StepTo checks correctness of `ts`.

jatentaki:
Adds the answer to #120 in `StepTo` docstring. The JAX ecosystem is full of silent errors so I think it's useful to tell the user explicitly they don't need to be paranoid about misusing the feature (like I was).


patrick-kidger:
Thanks!

PR #144, opened 2022-08-09
Adding support for complex dtypes

golanor:
Hi,

I finally have the time to actually finish this PR. I've been trying to start from scratch instead of going over all of the changes, and currently the solvers are able to return a solution, it's just the wrong solution. 
I'm not sure how to debug this, any suggestions?

### DIFF
--- a/diffrax/integrate.py
+++ b/diffrax/integrate.py
@@ -781,7 +781,10 @@ def diffeqsolve(
     save_index = 0
     made_jump = False if made_jump is None else made_jump
     ts = jnp.full(out_size, jnp.inf)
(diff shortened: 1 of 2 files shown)


patrick-kidger:
Some suggestions:
- Go through the guts of the integration code, adding `jax.experimental.host_callback.id_print` statements to see what value each array takes, and where things go wrong.
- Try using e.g. `Tsit5().step` directly instead of the `diffeqsolve` interface. Check if that works as you expect.
- Try using only constant step sizes, and not adaptive step sizing, if you aren't already. Just trying to get the simplest possible thing working first.

golanor:
It seems that `jax.experimental.host_callback.id_print` is doing nothing when I run in debug mode. Should I use it directly in a print statement?

PR #147, opened 2022-08-14
fix warning

### LINKED ISSUE #146: Still get warnings on `tree_util`
Hi @patrick-kidger ,
I find recent version of `diffrax` has updated to `jax 0.3.15`, but there's still a place you use `jax.tree_leaves` intead of `jtu.tree_leaves` in `diffax/misc/ad.py` in line 32 which leads to warnings, this is a small mistake, maybe u have noticed it. : )

patrick-kidger:
Fixes #146

### DIFF
--- a/diffrax/misc/ad.py
+++ b/diffrax/misc/ad.py
@@ -29,7 +29,7 @@ def is_perturbed(x: Any) -> bool:
 
 
 def nondifferentiable_input(x: PyTree, name: str) -> None:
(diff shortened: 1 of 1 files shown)


JiaYaobo:
hi @patrick-kidger , actually I think every api you use `jax.tree_leaves` has not been replaced by `jtu.tree_leaves`, since I find other files are still generating warnings (eg, term.py and integrate.py), maybe you just replace all `jax.tree_map` to `jtu.tree_map` and ignore `jax.tree_leaves` :)

PR #149, opened 2022-08-16
jax -> jtu for relevant functions

jacobusmmsmit:
There were still cases of deprecation warnings coming from jtu functions being called from jax directly. This PR fixes these cases by replacing `jax` with `jtu` for all functions in [`jax.tree_util`](https://jax.readthedocs.io/en/latest/jax.tree_util.html). These were `tree_leaves`, `tree_transpose` and `tree_unflatten`.

### DIFF
--- a/diffrax/adjoint.py
+++ b/diffrax/adjoint.py
@@ -2,7 +2,6 @@ import abc
 from typing import Any, Dict
 
 import equinox as eqx
-import jax
 import jax.lax as lax
 import jax.numpy as jnp
 import jax.tree_util as jtu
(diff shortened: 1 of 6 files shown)


patrick-kidger:
Looks like pre-commit checks are failing. (Should be an easy enough fix.)

jacobusmmsmit:
Apologies for not running the tests myself. I did attempt to follow the guidelines but the compute terminals I was using weren't playing well with `pytest`. I figured that because the changes were so small there wouldn't be any problems. I hope that's the case now.

patrick-kidger:
Merged - thank you!

PR #159, opened 2022-09-14
added new kalman-filter example

SimiPixel:
I wrote a little additional example that showcases `diffrax` in a maybe not so obvious way. It also showcases `equinox` and the ability to freeze parameters. 
Let me know what you think (and what needs to be changed).
Greetings


patrick-kidger:
This is quite nice.

A few high-level comments to begin with. Can you:
- tweak the introduction to follow the same style as the current examples? In particular a bit of discussion, some references etc? Atm terms like `LTISystem` show up without explanation.
- add this to the list of examples in `mkdocs.yml`?
- run the pre-commit hooks to autoformat this?
- Import `jax.random` as `jr`? (This is what I've settled on going forward.)
- Follow the Equinox style for docstrings?
- To tidy things up a bit: run the notebook afresh so that the cells increment 1-2-3-..., and delete the empty last cell?

patrick-kidger:
Yep, checking if there was interest before polishing this definitely makes a lot of sense.

Regarding the pre-commit hooks: see [CONTRIBUTING.md](https://github.com/patrick-kidger/diffrax/blob/main/CONTRIBUTING.md). These are a set of scripts that run when you commit (and which the tests here will also check) that autoformat your code, check it for common errors, etc. Super useful for keeping things automatically tidy.

SimiPixel:
Sorry this took so long.

- pre-commit hooks pass (i.e. notebook is autoformatted)
- added some introductory words on Kalman-Filter and LTI systems
- add this to the list of examples in mkdocs.yml? done

SimiPixel:
Also, i could not run neither `pytest` nor `mkdocs` as they both throw some error. 

For `pytest` this is weird because i did not change any source files.

For `mdocs` i would have ran it to see the notebook after nbconvert. However, it gives following error
`
ERROR    -  Error reading page 'examples/steady_state.ipynb': Additional properties are not allowed ('id' was unexpected)
`

patrick-kidger:
For `pytest` -- I think this is due to an upstream change in JAX. If you rebase/merge the current latest version of Diffrax, do you still get an error? (And if so what error?)

For `mkdocs`: again I think this is due to an upstream change. Try installing the specifically pinned versions in `docs/requirements.txt`.

SimiPixel:
`pytest`
Works now. Took about ~30 Minutes.

`mkdocs`
Works with pinned versions of nbformat / nbconvert. 

Also fixed some typography in the notebook.

examples/kalman_filter.ipynb
```diff
@@ -0,0 +1,490 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Problem Formulation\n",
+    "\n",
+    "Assume that there exists some *unknown* dynamical system of the form\n",
+    "\n",
+    "$\\frac{dx}{dt} = f(x(t), u(t), t)$\n",
+    "\n",
+    "$y(t) = g(x(t)) + \\epsilon$ \n",
```
patrick-kidger, line 57: Nit: we could make the time dependence explicit with `\frac{dx}{dt}(t)` and `\epsilon(t)`.

examples/kalman_filter.ipynb
```diff
@@ -0,0 +1,490 @@
...
+    "\n",
+    "where \n",
+    "- $u(t)$ denotes the time-dependent input to the system\n",
+    "- $y(t)$ denotes the time-dependent output / measurement to the system\n",
+    "- $x(t)$ denotes the time-dependent state of the system (which is not directly measureable in general)\n",
+    "- $f,g$ denote the time-dependent dynamics- and measurement-function, respectively\n",
+    "- $\\epsilon$ denotes random measurement uncertainty\n",
+    "\n",
+    "*The goal is to infer $x(t)$ from $y(1:t)$ even though $f,g,\\epsilon$ are unkown.*\n",
```
patrick-kidger, line 66: I don't think `y(1:t)` is standard notation. I think it'd be technically correct to write "The goal is to infer $x$ from $y$ even though", or to use $\\{x(t)\\}_t$ and $\\{y(t)\\}_t$ if one wished to emphasise the time-dependence.

examples/kalman_filter.ipynb
```diff
@@ -0,0 +1,490 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "class LTISystem(eqx.Module):\n",
+    "    A: jnp.ndarray\n",
+    "    B: jnp.ndarray\n",
+    "    C: jnp.ndarray\n",
+    "\n",
+    "    def __iter__(self):\n",
+    "        return (x for x in (self.A, self.B, self.C))"
```
patrick-kidger, line 119: Can we avoid defining this `__iter__` method? It's neat, but it's only used once below.

(Also generator comprehensions are probably worth avoiding in introductory examples; someone will think it's a tuple comprehension.)

examples/kalman_filter.ipynb
```diff
@@ -0,0 +1,490 @@
...
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "WARNING:absl:No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)\n"
+     ]
+    }
+   ],
+   "source": [
+    "def _interpolate_us(ts, us, B):\n",
+    "    if us:\n",
```
patrick-kidger, line 164: `if us is None`; I prefer to avoid implicit-cast-to-bool.

examples/kalman_filter.ipynb
```diff
@@ -0,0 +1,490 @@
...
+    "        u_t = dfx.LinearInterpolation(ts=ts, ys=us)\n",
+    "    else:\n",
+    "        m = B.shape[-1]\n",
+    "        u_t = SimpleNamespace(evaluate=lambda t: jnp.zeros((m,)))\n",
+    "    return u_t\n",
+    "\n",
+    "\n",
+    "def _diffeqsolve(\n",
```
patrick-kidger, line 172: Let's remove the underscore at the start of the name; I'd prefer simplicity instead of declaring public vs private here.

examples/kalman_filter.ipynb
```diff
@@ -0,0 +1,490 @@
...
+    "    rhs,\n",
+    "    ts: jnp.ndarray,\n",
+    "    y0: jnp.ndarray,\n",
+    "    solver: dfx.AbstractSolver = dfx.Dopri5(),\n",
```
patrick-kidger, line 176: A better default solver these days is usually `Tsit5`!

examples/kalman_filter.ipynb
```diff
@@ -0,0 +1,490 @@
...
+    "    return xs, ys"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Here we define the Kalman-Filter.\n",
+    "\n",
+    "Note how we use `equniox` to combine the Kalman-Filter logic in `__call__` and the Kalman-Filter parameters `Q` , `R` in one object."
```
patrick-kidger, line 220: Typo `equniox`

examples/kalman_filter.ipynb
```diff
@@ -0,0 +1,490 @@
...
+    "                A @ P\n",
+    "                + P @ A.transpose()\n",
+    "                + self.Q\n",
+    "                - P @ C.transpose() @ jnp.linalg.inv(self.R) @ C @ P\n",
+    "            )\n",
+    "\n",
+    "            # eq 3.23 of Ref [1]\n",
+    "            dxdt = A @ x + B @ u_t.evaluate(t) + K @ (y_t.evaluate(t) - C @ x)\n",
+    "\n",
+    "            return jnp.concatenate((dxdt, dPdt.reshape(n**2)))\n",
```
patrick-kidger, line 270: Diffrax supports using PyTree-valued state! We don't need to concatenate/reshape/slice here.

`y0 = (self.x0, self.P0)` etc.

patrick-kidger (commented):
Alright, I've left the usual gamut of various nitty comments. But I have to say this is very cleanly written, and I really like what you've done with this example. Thank you for contributing, and I look forward to merging this.

examples/kalman_filter.ipynb
SimiPixel: Sure

examples/kalman_filter.ipynb
SimiPixel: Hmm, so i just wanted to highlight that we are doing filtering and not smoothing, i.e. not $x(1:T)$ from $y(1:T)$. 

But these are again maybe just details and then often simpler is better. So, sure!

examples/kalman_filter.ipynb
SimiPixel: Sure

examples/kalman_filter.ipynb
SimiPixel: I will keep that in mind ;)

examples/kalman_filter.ipynb
SimiPixel: Heh did not know that. That's nice

patrick-kidger:
And merged! Thank you for contributing.

Issue #180, opened 2022-10-24
Make brownian motion path classes accept shape: PyTree[Tuple[int]]


ciupakabra:
Suppose we want to integrate the following SDE ( $X_t$ is $2d$ dimensional)

$$
dX_t = f(X_t, t)\mathrm{d}t + 
\begin{pmatrix} 
    0_{d \times d}\\
    1_{d \times d}
\end{pmatrix}\mathrm{d}W_t
$$

For the diffusion term it would be wasteful to use the full `diffrax.ControlTerm` term, instead we'd like to make use of `diffrax.WeaklyDiagonalControlTerm`. `diffrax.WeaklyDiagonalControlTerm` allows our vector fields and control paths to output PyTrees, so we could start by defining `vector_field = lambda _ : (jnp.zeros(d), jnp.ones(d))`. However, correct me if I'm wrong but at the moment there's no clean way to create a Brownian path which outputs PyTrees, since `diffrax.VirtualBrownianTree` accepts the argument `shape` only as a `Tuple[int]`. I see no reason why it couldn't be `PyTree[Tuple[int]]`, which allows much more freedom modelling the diffusion term efficiently.

Since it seems to be quite a small change I'm happy to write a PR for this, if the authors like the suggestion.

patrick-kidger:
Yep, that makes sense to me. A PR would be welcomed.

I think `shape` should probably be of type `Union[Tuple[int, ...], PyTree[jax.ShapeDtypeStruct]]`. This would additionally allow for specifying dtype if required.

patrick-kidger:
Closing as now implemented in Diffrax 0.2.2.

PR #183, opened 2022-10-29
Brownian motion classes accept pytrees for shape and dtype arguments

ciupakabra:
This changes the argument `shape` for classes `VirtualBrownianTree` and `UnsafeBrownianPath`, and adds an additional argument `dtype` as per the dicussion in #180.

- I decided upon `shape: Pytree[Tuple[int, ...]` instead of `shape: Union[Tuple[int, ...], PyTree[jax.ShapeDtypeStruct]]`. It's unclear what to do with `named_shape` in `jax.ShapeDtypeStruct` -- I don't know if there is a way to sample Brownian motion via named shapes. But if you feel strongly about this and give me some pointers, I can reimplement.
- To allow specifying dtypes, `dtype` argument specifies them as a pytree and has to be a prefix tree of `shape`.
- I added `__init__` methods to both classes since I was not sure how to have `dtype=None` without it.
- Added some helper functions that I use in `misc.py`, hope that's the right location to place them.
- Used `jtu.tree_map` instead of `jax.vmap` -- was not sure how to supply `is_leaf` to `jax.vmap`. Happy to change this as well, with some pointers.
- Changed the `test_brownian.py:test_shape` to test pytree shapes and dtypes. Just noticed that formatting made it look pretty bad, not sure if that's a big deal.
- Tests pass locally.

Let me know what you think. Thanks!


patrick-kidger:
- I think it should be fine to just ignore the `named_shape` field. Most of the time a `jax.ShapeDtypeStruct` is just meant as a way to hold a shape and a dtype together, and it can be called as just. `ShapeDtypeStruct(shape, dtype)`. In particular I think you should find that this approach means that many of the new helper functions are unnecessary.
- I'm not too worried about the funny-looking formatting. It's not that elegant, but the occasional oddity is the price of using autoformatters.
- We should probably add a check that the dtypes are floating-point. (`jnp.issubdtype(dtype, jnp.inexact`)

ciupakabra:
I've changed to `shape: Union[Tuple[int, ...], PyTree[jax.ShapeDtypeStruct]]` and added a check for floating point dtypes.

diffrax/brownian/path.py
```diff
@@ -29,11 +37,28 @@ class UnsafeBrownianPath(AbstractBrownianPath):
...
+    ):
+        self.shape = (
+            jax.ShapeDtypeStruct(shape, None) if is_tuple_of_ints(shape) else shape
+        )
+        error_if(
+            not jtu.tree_all(
+                jtu.tree_map(lambda x: jnp.issubdtype(x.dtype, jnp.inexact), self.shape)
+            ),
+            "UnsafeBrownianPath dtypes all have to be floating-point.",
+        )
```
patrick-kidger, line 58: This should just be a regular `ValueError`. The `error_if` function is for raising runtime errors, but this is an error we'd like to catch at trace time.

I.e.
```python
if any(not jnp.issubdtype(x.dtype, jnp.inexact) for x in jtu.tree_leaves(self.shape)):
     raise ValueError(...)
```

diffrax/brownian/path.py
```diff
@@ -43,20 +68,29 @@ def t1(self):
...
+            shape.dtype
+        )
 
 
 UnsafeBrownianPath.__init__.__doc__ = """
 **Arguments:**
 
-- `shape`: What shape each individual Brownian sample should be.
+- `shape`: What shape each individual Brownian sample should be. Can be 
+specified by a pytree.
```
patrick-kidger, line 94: Can this be a little more explicit?

```
Should be a PyTree of `jax.ShapeDtypeStruct`s, representing the shape, dtype, and PyTree structure of the output. For simplicity, `shape` can also just be a tuple of integers, describing the shape of a single JAX array.
```

diffrax/brownian/tree.py
```diff
@@ -1,14 +1,15 @@
 from dataclasses import field
-from typing import Optional, Tuple
```
patrick-kidger, line 2: (Same two nitty comments in this file.)

patrick-kidger (commented):
I've left a couple of nitty comments but otherwise I'm happy with this.

diffrax/brownian/path.py
ciupakabra: That makes sense -- I've changed this

diffrax/brownian/path.py
ciupakabra: You're right. Also added

```
... For simplicity, `shape` can also just be a tuple of integers, describing the shape of a single JAX array. In that case the dtype is chosen to be `float64` if `JAX_ENABLE_X64=True` and `float32` otherwise.
```

which I _think_ is correct

ciupakabra:
Thanks for the review! I added the changes.

One last thing -- the tests I said pass were run on python 3.10. I've tried this PR on another server which has python 3.8 and throws the following error when testing:

```
___________________________________________________________________________________________________________ ERROR collecting test/test_brownian.py ____________________________________________________________________________________________________________
test/test_brownian.py:3: in <module>
    import diffrax
diffrax/__init__.py:1: in <module>
    from .adjoint import (
diffrax/adjoint.py:9: in <module>
    from .misc import implicit_jvp, nondifferentiable_output, ω
diffrax/misc/__init__.py:9: in <module>
    from .misc import (
diffrax/misc/misc.py:167: in <module>
    def split_by_tree(key, tree, is_leaf: Optional[Callable[PyTree, bool]] = None):
/usr/lib/python3.8/typing.py:813: in __getitem__
    raise TypeError(f"Callable[args, result]: args must be a list."
E   TypeError: Callable[args, result]: args must be a list. Got <class 'diffrax.custom_types.PyTree'>
```

which I guess is either difference in python or `typing` versions. I can fix this error, but would also want to test this PR more systematically. I'm not sure how to do this -- i.e. what versions of python or other packages should I test this with?

patrick-kidger:
The error is that it should be `Callable[[PyTree], bool]`. The first argument to `Callable` should be a list of the arguments it accepts.

ciupakabra:
Yes, my question was more about how to test this more systematically, but I didn't notice that the tests are run on 3 different versions of python. Should be good now.

ciupakabra:
Not sure why it's failing -- that test passes on my machine. Package versions agree as well

```bash
(env) [ao464@csic41 diffrax]$ pip list
Package             Version Location                                  
------------------- ------- ------------------------------------------
absl-py             1.3.0   
attrs               22.1.0  
chex                0.1.5   
diffrax             0.2.1   /home/ao464@ad.eng.cam.ac.uk/repos/diffrax
dm-tree             0.1.7   
equinox             0.9.0   
etils               0.9.0   
exceptiongroup      1.0.1   
importlib-resources 5.10.0  
iniconfig           1.1.1   
jax                 0.3.24  
jaxlib              0.3.24  
jaxtyping           0.2.7   
numpy               1.23.4  
opt-einsum          3.3.0   
optax               0.1.3   
packaging           21.3    
pip                 20.0.2  
pkg-resources       0.0.0   
pluggy              1.0.0   
psutil              5.9.3   
pyparsing           3.0.9   
pytest              7.2.0   
scipy               1.9.3   
setuptools          44.0.0  
tomli               2.0.1   
toolz               0.12.0  
typeguard           2.13.3  
typing-extensions   4.4.0   
wheel               0.38.1  
zipp                3.10.0  
(env) [ao464@csic41 diffrax]$ JAX_PLATFORM_NAME=cpu pytest test/test_adjoint.py::test_adjoint_seminorm
===================================================================================================================== test session starts =====================================================================================================================
platform linux -- Python 3.8.10, pytest-7.2.0, pluggy-1.0.0
rootdir: /home/ao464@ad.eng.cam.ac.uk/repos/diffrax
plugins: typeguard-2.13.3, jaxtyping-0.2.7
collected 1 item                                                                                                                                                                                                                                              

test/test_adjoint.py .                                                                                                                                                                                                                                  [100%]

====================================================================================================================== 1 passed in 0.96s ======================================================================================================================
```

ciupakabra:
From https://docs.python.org/3/library/sys.html#sys.modules

```
sys.modules
[...]. If you want to iterate over this global dictionary always use sys.modules.copy() or tuple(sys.modules) to avoid exceptions as its size may change during iteration as a side effect of code or activity in other threads.
```

Should I add the change to this PR?

patrick-kidger:
Ah! I've recently had to fix this locally on my own (as-yet-unreleased) PRs. I'm not sure why it ever ran before.
Go ahead and add that in here.

ciupakabra:
This is ready to be tested again

patrick-kidger:
Alright, LGTM! Thank you for contributing.

Issue #96, opened 2022-04-26
Solving with complex initialization


brosand:
When trying to run a basic example: 

```
from diffrax import diffeqsolve, ODETerm, Dopri5
import jax.numpy as jnp
from jax.lax import cond


Array.set_default_backend('jax')
def f(t, y, args):
    return -y

term = ODETerm(f)
solver = Dopri5()
y0 = jnp.array([2., 3.], dtype=complex)
solution = diffeqsolve(term, solver, t0=0, t1=1, dt0=0.1, y0=y0)
```

I receive this warning: `ComplexWarning: Casting complex values to real discards the imaginary part`, which goes away if I change the type of `y0` to `float`.

Moreover, when running something else with a complex `y0`, there was an error 'TypeError: nextafter does not accept dtype complex64 at position 0. Accepted dtypes at position 0 are subtypes of floating.'

Does diffrax not support complex `y0`s? If so is there any plan to enable this? Although we can convert the complex elements to 2d arrays, it would make the process a lot smoother if we could just use complex elements.

patrick-kidger:
Diffrax doesn't support complexes at the moment. This was a concious choice on my part, so that other aspects of Diffrax could be prioritised.

My expectation is that there's probably just a few lines of code, that are currently assuming reals, that need tweaking.
I'd welcome a PR on this.

golanor:
From what I looked, it seems that this line: https://github.com/patrick-kidger/diffrax/blob/cec091c5e4cc4311f64ae3aa09a371db5fe766ee/diffrax/integrate.py#L771

```python
ys = jax.tree_map(lambda y: jnp.full((out_size,) + jnp.shape(y), jnp.inf), y0)
```
is what causing complex numbers to be cast as floats.

What do you think @patrick-kidger ? Are there other places that need changing? If it is only a matter of changing the dtype here to be complex, I'll open a PR.

patrick-kidger:
I haven't looked into this so I don't know how many places might need changing. If you're happy to dig into this I'd still be happy to accept a PR on this. (Including tests, to be sure that this doesn't revert in the future.) My expectation is that it's probably just a few lines that need tweaking.

Phionx:
What's the status of supporting complex dtypes in `diffrax`? I see that PR #112 was closed before merging. 

---

BTW (in case it helps anyone looking at this thread), as a temporary workaround, I used the following isomorphism [1] between complex-valued and real matrices/vectors:

[...]

PR #197, opened 2022-11-29
Complex dtype support

TimonHoess:
Hi, so I have tweaked the code such that complex dtypes should now be handled correctly as discussed in #96 -- at least for the explicit solvers.

As for the implicit solvers, the issue is the jacobian which could be made to work without errors by passing the `holomorphic=True` to `jax.jacfwd`, however that would assume of course that the functions need to be holomorphic. Also I am not sure if there are other things one would need to pay attention to when using Newton's method on complex valued functions.

I am not sure if the tests that I added are sufficiently comprehensive, so if there are more cases that should be covered I would be happy about advice.

Additionally I found that in the case that `y0` is given with single precision dtype e.g. `float32` but all 'timelike' variables are given as weak types it would lead to a `TypeError`, which should now also be fixed.


patrick-kidger:
This pretty much LGTM!

I'd suggest also adding a test that compares the complex solve against the equivalent thing using two separate real numbers for real and imaginary part.

---

Fwiw, it's not enough to ensure that it just runs without crashing. For example see this line:

https://github.com/patrick-kidger/diffrax/blob/5f5a1217131f8d2d55ff14f4e895432a42167236/diffrax/misc/misc.py#L95

Which happens to be written as `jnp.real(x * jnp.conj(x))` rather than `x**2`. (In anticipation of complex support someday.) But `x**2` would still have run and produced silently the wrong result.

So I'd suggest going over the code to check that you're not running afoul of anything like this.

---

It's worth noting that right now, ensuring complex support throughout all of Diffrax is a non-goal. I'm happy to accept PRs making appropriate tweaks, but anything not covered by tests is something that may break in the future. (It's just too difficult to get this consistently correct right now.)

---

As an alternate plan to all of the above: perhaps if `y0` is complex then we should:
- split it into separate real and imaginary parts at the very start of `diffeqsolve`, and solve the diffeq using real numbers;
- wrap the vector field in something that recombines the real and imaginary part into a complex number, calls the user-provided vector field, then splits apart the result;
- combine the real and imaginary parts again before returning the solution.

This would mean we could support complex numbers ubiquitously, without needing to carefully plumb them through every operation.

I don't know if this approach has performance concerns though.

---

WDYT?

Randl:
FWIW, you can just use `np.iscomplexobj` to pass the holomorphic parameter:
```
return jsp.linalg.lu_factor(
            jax.jacfwd(curried, holomorphic=jnp.iscomplexobj(flat))(flat)
        )
```

As for testing, I think that adding complex inputs to tests should be relatively easy for many tests, and the fact that it passes `test_basic` probably means that it is not obviously wrong at least

Randl:
Also note that diffrax currently silently fails if passed complex input, so this is not regression in this sense

Randl:
In fact, it turns out there is a problem with `rms_norm`, the jvp is complex instead of being real, see
https://github.com/Randl/diffrax/commit/f610b45febfa74af01234beafba6ced769a93bb9

patrick-kidger:
Right! As you can see, complex support is a can of worms.

Agreed that having Diffrax silently fail isn't great, though. I'd be happy to accept a PR that raises a warning if any of the inputs are complex.

I'd need to brush up on my complex autodiff to be sure that we do the right thing here, especially around calls to `jacfwd` , linear solves, etc. For example PyTorch uses the [conjugate Wirtinger derivative](https://pytorch.org/docs/stable/complex_numbers.html#autograd) but JAX uses the [nonconjugate Wirtinger derivative](https://pytorch.org/docs/stable/notes/autograd.html#complex-autograd-doc). The former is what you actually want when doing gradient descent, but the latter agrees with the holomorphic derivative.

TimonHoess:
Hi, sorry for not replying for that long time.
I have looked into the code of Diffrax and to my understanding these problems apply only to a limited set of solvers, or am I missing something here? 

`rms_norm` only appears in the implicit solver types (through the newton method) and in the adaptive step sizing, also `jvp` seems to only appear in the **implicit** adjoint method for backpropagation and in the Milstein solver. 

I don't know if there are some other hidden appearances that I missed, but wouldn't that mean that just some functionality is affected? I think if that is the case, one could -- for now -- warn only if those solvers that are incompatible with complex dtypes are used and otherwise let it run (and with further assurance through tests that show it works, of course). To me it looks like at least forward solving with non-implicit solvers and fixed stepsize works fine with complex numbers, I am not sure about backpropagating but it also seems to work for the same limited cases.

Randl:
I tend to agree. Since these changes do not break existing functionality and don't even deal will complex number s per SE, I do not see a good reason not to merge them. Adding a warning that complex number support is WIP would allow people to report bugs since after this pr at least basic functionality with complex input is not wrong.

diffrax/integrate.py
```diff
@@ -719,6 +719,7 @@ def diffeqsolve(
             dt0 = jnp.asarray(dt0, dtype=dtype)
         if saveat.ts is not None:
             saveat = eqx.tree_at(lambda s: s.ts, saveat, saveat.ts.astype(dtype))
+        timelikes.append(dtype)
```
patrick-kidger, line 722: Nit: I don't think this line is necessary?

patrick-kidger (commented):
Okay, I'm convinced! Let's merge this.
Adding a warning and then gradually building out support for complex numbers sounds like a good plan to me.

diffrax/integrate.py
TimonHoess: It is not necessary for complex input, but I came across this issue with dtypes while trying to figure out the complex-type issue. Consider:
```
import jax.numpy as jnp
from jax.config import config
import diffrax
config.update("jax_enable_x64", True)

term = diffrax.ODETerm(lambda t, y, args: -1.0 * y)
y0 = jnp.array([1.0, 1.0], dtype=jnp.float32)
solver = diffrax.Euler()
sol = diffrax.diffeqsolve(term, solver, t0=0.0, t1=1.0, dt0=0.01, y0=y0)
```

https://github.com/patrick-kidger/diffrax/blob/9280c3a39c0c96851f38e4e38ab14777050f6032/diffrax/integrate.py#L593

In this case `jnp.result_type(*timelikes)` apparently will return the highest possible precision of the weak dtypes; `dtype('float64')`  if you have `jax_enable_x64`set to `True`. 

https://github.com/patrick-kidger/diffrax/blob/9280c3a39c0c96851f38e4e38ab14777050f6032/diffrax/integrate.py#L609

In this line `jnp.result_type(yi, *timelikes)` will return `dtype('float32')`  and you will have different dtypes for your timelike variables and your state and get a TypeError.

However, I see that this is a rather unrelated issue and the way I worked around it is not really ideal (if you specify your input as float32 and all other as weak types it should probably return float32 not float64). Should I open this as a new issue and suggest this workaround as a separate pull request?

patrick-kidger: Ah -- I understand. that's a subtle issue!

I think probably the neatest fix might be to adjust [this line](https://github.com/patrick-kidger/diffrax/blob/9280c3a39c0c96851f38e4e38ab14777050f6032/diffrax/integrate.py#L609) to `jnp.result_type(yi, dtype)`?
(And either way include a comment linking to this discussion, for posterity.)

I'd be happy to take this fix alongside this PR; no need for a separate one.

On which note -- apologies for taking so long to get back around to this! If you're happy to update this to the latest version of Diffrax, then I would be happy to merge this.

Randl:
You may want to cherry pick this one: https://github.com/Randl/diffrax/commit/f610b45febfa74af01234beafba6ced769a93bb9

Randl:
@patrick-kidger @TimonHoess  any updates on this one?

patrick-kidger:
Ah, this one slipped through the cracks!
I think the changes look good as-is. If you can rebase on to the latest main I'd be happy to merge this now.

Randl:
If @TimonHoess is unavailable, I can try to rebase his PR on master

TimonHoess:
Please feel free to do so @Randl , otherwise I can probably do it in about a month. I'm quite busy for the upcoming weeks.

patrick-kidger:
Closing in favour of #330.

connorblake1:
@patrick-kidger Not sure if this is still an issue, but equinox/_jit.py:49 still raises a warning linking to this page.

patrick-kidger:
Ah, we should probably remove the reference here. (But keep the warning, complex support is now better but still needs more work.)

PR #198, opened 2022-11-30
Fix minor typo in rectilinear interpolation docs

slishak:
Just fixing a tiny mistake I noticed while reading the docs!


patrick-kidger:
LGTM! Thank you for the fix.

Issue #113, opened 2022-06-07
Saving metrics during solve


joglekara:
Hi, thanks again for the library.

we have a numerical integration we perform using `lax.scan`. We'd like to port it over to `diffrax`. I wanted to get your thoughts on the shortest path to this implementation. 

From https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.scan.html 

```python
def scan(f, init, xs, length=None):
  if xs is None:
    xs = [None] * length
  carry = init
  ys = []
  for x in xs:
    carry, y = f(carry, x)
    ys.append(y)
  return carry, np.stack(ys)
```

we rely quite a bit on the implicit `append` and `stack` calls that happen inside `lax.scan`. I think the core of my question is, is there a way to avoid the pre-allocation as you suggest in https://github.com/patrick-kidger/diffrax/issues/60#issuecomment-1034768386 . My guess is the answer is no but figured I'd see if you had any ideas.

TY in advance!

patrick-kidger:
Do I understand that:

1. As in #60, you're interested in logging metrics during the solve, and want to save something at every step?
2. As you're currently using a `lax.scan`, you're probably using a fixed step size rather than an adaptive step size?

If this is the case then this is totally doable. Follow the approach as in #60 to create a wrapper solver that stores your metrics. `metric_state` should be a buffer of the appropriate size (equal to the number of steps you'll be taking), which you update using "in-place updates". (`buffer = buffer.at[step].set(value)`.) Now the analogous behaviour to a `lax.scan` in Diffrax is to pass `stepsize_controller=ConstantStepSize(compile_steps=True)`, and if you do this then you should find that your "in-place updates" really do get complied down to efficient in-place updates!

patrick-kidger:
Can you give me some examples of the sort if thing you're trying to save, by the way? That this has come up twice now suggests that it may be worth creating an easy way to save user-specified metrics. (And in particular without needing to worry about this business of the efficiency of in-place updates.)

joglekara:
> Do I understand that:
> 
> 1. As in [Logging metrics during an ODE solve #60](https://github.com/patrick-kidger/diffrax/issues/60), you're interested in logging metrics during the solve, and want to save something at every step?
> 2. As you're currently using a `lax.scan`, you're probably using a fixed step size rather than an adaptive step size?
> 
[...]

PR #220, opened 2023-01-31
Add `func` method to `SaveAt`

joglekara:
See https://github.com/patrick-kidger/diffrax/issues/113

### DIFF
--- a/diffrax/integrate.py
+++ b/diffrax/integrate.py
@@ -1,6 +1,6 @@
 import functools as ft
 import warnings
-from typing import Optional
+from typing import Optional, Callable
 
 import equinox as eqx
 import equinox.internal as eqxi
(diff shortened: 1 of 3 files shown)


diffrax/integrate.py
```diff
@@ -283,7 +283,7 @@ def _body_fun(_state, _inplace):
                         jnp.where(_pred, __saveat_y, __ys[_save_index])
                     ),
                     _ys,
-                    _saveat_y,
+                    saveat.func(_ts[_save_index], _saveat_y, args)
```
patrick-kidger, line 286: Nit: `_ts[save_index]` is set to `jnp.where(_pred, _saveat_y, __ys[_save_index])` above, so we can just pull that our as an intermediate value and access it directly, rather than re-indexing.

patrick-kidger: You're also missing a couple of other spots in this file that need a `func` wrapper: the `saveat.steps` branch, and the `saveat.t0` and `saveat.t1` branch.

patrick-kidger (commented):
Good idea adding `func` to `SaveAt`.

Could you add some tests, as well?

diffrax/integrate.py
joglekara: - Not sure what you mean by the intermediate value bit
- Done

joglekara:
Added a (set of) test(s?) as a parameterization to the existing test

joglekara:
> Nit aside, this LGTM!

Great. Looks like the checks are failing bc of some package manager but LMK if you need something from my PR too

patrick-kidger:
Yup, looks like a broken upstream update. I'll take it from here.

Thank you for contributing!

PR #222, opened 2023-01-31
Add support for complex variables in ODEs

Randl:


Randl:
Hmm I've noted https://github.com/patrick-kidger/diffrax/pull/197

PR #233, opened 2023-02-22
Testing a possible upstream JAX fix

patrick-kidger:
~~Do not merge.~~


patrick-kidger:
Merging as a hopefully-temporary measure to get v0.3.1 out whilst hopefully JAX can get fixed.
This is only affecting the tests so it's not too bad to have all the private imports.

Issue #223, opened 2023-02-07
How does diffrax handle state variables becoming inf?


virajpandya:
I'm trying to port some code from pure python to JAX. The code is a bit complicated with time-varying interpolated inputs but essentially involves solving a system of 8 coupled ODEs. The state variables and independent time variable span a large dynamic range so the system is solved in logarithmic form (i.e., read in 10^logt and 10^logy, do calculations in physical units, return dlogy/dlogt). The system is stiff but my pure python implementation based on scipy.integrate.solve_ivp leads to reasonable solutions for multiple solvers (the implicit methods BDF and Radau being the best, but RK23 and DOP853 also finish and give similar solutions though require lower tolerances and more steps; all of these agree with my manual Euler integrations with extremely small fixed timesteps, increasing my confidence in the solutions). 

However, all of the adaptive diffrax solvers I've tried fail and JAX's odeint also fails. Both packages only get 4 steps in before they return NaN's (and diffrax hits up against max_steps=10K whereas scipy only needs 300 for BDF and 3000 for RK23). I used print/jax.debug.print() statements inside my integrator function to print the values of the state variables and their associated derivatives for both the pure python and JAX versions of my code. I find that early on, the derivatives for one or two of my state variables blows up (dlogy/dlogt=10^6) so in the next step, the value of that state variable becomes inf, and then that leads to NaN's. solve_ivp is able to deal with this by repeatedly going back to an earlier time and taking ever smaller timesteps, eventually getting over this initial hump. But both jax-odeint and diffrax remain stuck with NaN's. I was hoping diffrax's implicit solvers and adaptive step-size controller would help compared to the default jax-odeint one, but it has the same problem.

Should diffrax/jax-odeint be able to adapt to state variables becoming inf? Why is solve_ivp successful? 

I can probably share a self-contained github repo with two example Jupyter notebooks comparing solve_ivp to diffrax/jax-odeint if that would help.

patrick-kidger:
This sounds like an interesting problem! Diffrax should definitely be able to handle this kind of thing (at least, anything that `solve_ivp` can handle), and I've definitely had Diffrax detect `NaNs`, and go back and make smaller time steps until we get past the problematic region. [C.f. this line in the integrator](https://github.com/patrick-kidger/diffrax/blob/5c0bfb3fd58b4726060bb04e9bdb09f9cb8e6e63/diffrax/integrate.py#L150).

[...]

Issue #200, opened 2022-12-04
Forcing solver to stay in given region


grfrederic:
I'm solving a system of ODEs that simulates concentrations of some substances. I know that the concentrations have to be real numbers in the range [0, 1]. Normally, after each step, I would simply clamp the values to be in that region (since values outside of that range can mess with the simulation). Is there a way to achieve this nicely with diffrax?

I'm trying out a pretty wide range of parameters (since I'm using the simulations for MCMC), so ramping up the accuracy and step sizes to handle the outliers seems wasteful.

patrick-kidger:
There's a couple of ways you could do this. The first, as you say, is to just clamp the values to the desired region. You can do this by wrapping the solver:

```python
class ClampSolver(dfx.AbstractSolver):
    solver: dfx.AbstractSolver
    clamp: Callable

    def step(...):
        y1, ... = self.solver.step(step)
        y1 = self.clamp(y1)
        return y1, ...

    # also forward any other methods to `self.solver` as needed

clamp = lambda y: jnp.clip(y, a_min=0, a_max=1)
solver = ClampSolver(dfx.Tsit5(), clamp)
```

Another way (that will give more accurate solutions) is to have the step size controller reject any steps that are out of bounds (which we do by setting the error estimate to infinity):
```python
class InBoundsSolver(dfx.AbtractSolver):
    solver: dfx.AbstractSolver
    out_of_bounds: Callable

    def step(...):
        y1, y_error, ... = self.solver.step(...)
        oob = self.out_of_bounds(y1)
        keep = lambda y: jnp.where(oob, jnp.inf, y)
        y_error = jax.tree_util.tree_map(keep, y_error)
        return y1, y_error, ...

    # also forward any other methods to `self.solver` as needed

out_of_bounds = lambda y: (y < 0) | (y > 1)
solver = InBoundsSolver(dfx.Tsit5(), out_of_bounds)
```

Eventually I intend to provide built-in support for this kind of thing, so I'm also going to mark this as a feature request.

grfrederic:
Thank you! I really like the second solution.

PR #235, opened 2023-02-22
use jnp.nanmin and jnp.nanmax to compute new stepsize factor 

virajpandya:
…instead of of jnp.clip in diffrax.step_size_controller.adaptive.adapt_step_size()

Care was taken to make sure that the order of jnp.nanmin and jnp.nanmax reflects the actual behavior of jnp.clip. According to https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.clip.html, using jnp.(nan)min and jnp.(nan)max may be a bit slower than  jnp.clip but at least it will be robust against NaN's in y_error.

This solves https://github.com/patrick-kidger/diffrax/issues/223 and is similar to my proposed bugfix / pull request for jax.experimental.ode.optimal_step_size() here: https://github.com/google/jax/issues/14612 and https://github.com/google/jax/pull/14624 

I confirmed that this works with different explicit/implicit diffrax solvers and I get the expected correct solution to my non-autonomous ODE system vs. scipy.integrate.solve_ivp, manual non-adaptive Euler integration with extremely small timesteps, and manual adaptive RK23 (Bogacki-Shampine) solver in both pure python and JAX.

### DIFF
--- a/diffrax/step_size_controller/adaptive.py
+++ b/diffrax/step_size_controller/adaptive.py
@@ -496,11 +496,7 @@ class PIDController(AbstractAdaptiveStepSizeController):
         factor2 = 1 if _zero_coeff(coeff2) else prev_inv_scaled_error ** coeff2
         factor3 = 1 if _zero_coeff(coeff3) else prev_prev_inv_scaled_error ** coeff3
         factormin = jnp.where(keep_step, 1, self.factormin)
-        factor = jnp.clip(
-            self.safety * factor1 * factor2 * factor3,
-            a_min=factormin,
-            a_max=self.factormax,
-        )
+        factor = jnp.nanmin(jnp.array([self.factormax, jnp.nanmax(jnp.array([self.safety * factor1 * factor2 * factor3, factormin]))]))
         # Once again, see above. In case we have gradients on {i,p,d}coeff.
         # (Probably quite common for them to have zero tangents if passed across
         # a grad API boundary as part of a larger model.)


patrick-kidger:
Okay, sorry for taking so long to get back around to this!

One thing I would like to understand is why [this approach](https://github.com/patrick-kidger/diffrax/issues/223#issuecomment-1437723913) failed. It looks like both possible fixes involve detecting NaNs.

In particular you mentioned wanting to autodiff through these solvers. If one generates a NaN and then removes it later, then this can sometimes still re-appear on the backward pass. Being robust to such issues usually means catching the NaN as soon as possible.

Are you able to track down how the NaN still sneaks by in the previous approach? I'd definitely like to get some version of this fix in.

virajpandya:
Sorry for the delay -- I'm finally getting back to this. Thanks for all the help! So the good news is that with my fix, the system can both be successfully solved and forward-mode autodiff'd (jax.jacfwd and jax.jvp). I verified the resulting Jacobian and JVP with finite-difference (using atol and rtol << parameter perturbations). That was with your old v0.2.2 using NoAdjoint(). Switching to your latest v0.3.1 without either my fix or yours, the system and forward-mode autodiff also are both successful (using DirectAdjoint()). I don't yet know exactly what changed in v0.3.1 to have this work out of the box now but it's great! 

What doesn't work is reverse-mode autodiff and I would like to know why. RecursiveCheckpointAdjoint() gives NaN gradients and BacksolveAdjoint() leads to a 'max_steps reached' error, even though the system is solved in 290 steps.

**Where in diffrax do you recommend putting jax.debug.print statements so I can see if something wonky is happening in solving my ODE system backwards in time?** I checked ad.py, adjoint.py and integrate.py and I see a lot of equinox calls, but are there specific places where, e.g., I can print the value of the 8 state variable values and their time derivatives along the backward pass? (My ODE function isn't designed to deal with negative times so that's the first thing I'd check.) 

FWIW, the standard jax.experimental.ode.odeint has the same problem -- the system gets solved with my NaN fix, but because it is only designed to do reverse-mode autodiff with its custom VJP, jax.jacrev and jax.vjp fail. Its internals seem easier to add jax.debug.print statements to so I'll see what's happening in their _odeint_rev function.

Randl:
Any updates on this one?

patrick-kidger:
Getting NaNs in reverse-mode was expected here I think -- papering over them in the forward pass doesn't remove them from the computation graph, so it's common to expect that they should reappear in the backward pass. I think the goal here should be to find where the NaNs arose in the forward pass in the first place.

Randl:
So well, my use case has a fastly growing function, and for too large step size the result is Nan. I think what happens is that currently this propagates to the step size and kills the solver. Applying this PR locally fixed the problem for me, that's why I'm asking

patrick-kidger:
Right! So I think the correct fix is not this PR, but instead to locate where the NaNs arise, and prevent them from ever being created.

Randl:
So lets say I have `dy/dt = f(y)` where `f` is not defined for whole R but rather for some subset `U`. Of course the dynamics are such that `y` is never outside of `U`. Yet nothing prevents the solver (and specifically step size controller) from trying values outside `U`, especially if the solution approaches the boundary. What would be the solution (except praying that the step size is not too large of course)?

patrick-kidger:
This is a good question. See also #200 for another take on this.

Right now the answer is really to:
(a) write your vector field such that it returns non-NaN values for all possible inputs [as above, the moment you get a NaN then it might all be over]
(b) use the `InBoundsSolver` of #200 to ensure you stay within the valid region.

I think this answer is a little dissatisfying. We should probably add `InBoundsSolver` into Diffrax itself.
I am also trying to figure out a way to make (a) happen automatically -- or maybe some way that we can be robust to NaNs in all cases. Getting that right is a tricky thing for a user to do.

Randl:
Ok but what are the arguments against this PR? It will never break working code and may help in cases like mine. Propagating nans is also not the best way of debugging, the user is better of checking for nans in user functions if that is a concern. Minor slowdown? Maybe make this one optional then?

patrick-kidger:
I can see that this PR improves forward integration of systems only defined in some domain.

However, my concern for this PR is that it when backpropagating through a solve which produces NaNs, then it delays the error from the forward pass to the backward pass. NaNs arising only on the backward pass are incredibly tricky to debug!

PR #253, opened 2023-05-14
Initial step size now uses a scan trick to reduce compilation time

patrick-kidger:


patrick-kidger:
Closing in favour of #257.

PR #254, opened 2023-05-14
General RK tidy-ups

patrick-kidger:
This does also fix one bug, in which the predictor tableau wasn't meaningfully being used for implicit methods.


patrick-kidger:
Closing in favour of #257.

Issue #282, opened 2023-07-26
Neural ODE example `eqxi` error ? 


thibmonsel:
Hi Patrick, 

It seems that the new equinox version and `eqxi.error_if` makes the equinox version make the Neural ODE in the documentation crash : 

```py
Traceback (most recent call last):
  File "/home/monsel/Desktop/test2.py", line 157, in <module>
    ts, ys, model = main()
  File "/home/monsel/Desktop/test2.py", line 106, in main
    ts, ys = get_data(dataset_size, key=data_key)
  File "/home/monsel/Desktop/test2.py", line 71, in get_data
    ys = jax.vmap(lambda key: _get_data(ts, key=key))(key)
  File "/home/monsel/Desktop/test2.py", line 71, in <lambda>
    ys = jax.vmap(lambda key: _get_data(ts, key=key))(key)
  File "/home/monsel/Desktop/test2.py", line 61, in _get_data
    sol = diffrax.diffeqsolve(
  File "/home/monsel/miniconda3/envs/diffrax_debug/lib/python3.9/site-packages/diffrax/integrate.py", line 572, in diffeqsolve
    dt0 = eqxi.error_if(dt0, pred, msg)
ValueError: No arrays to thread error on to.
```

A quick fix is to change the float `dt0` to an array `jnp.array(dt0)`. Not sure how you intent `eqxi.error_if` to work with float.

My package versions are : 
```
absl-py==1.4.0
chex==0.1.82
contourpy==1.1.0
cycler==0.11.0
diffrax==0.4.0
equinox==0.10.11
fonttools==4.41.1
importlib-metadata==6.8.0
importlib-resources==6.0.0
jax==0.4.13
jaxlib==0.4.13
jaxtyping==0.2.20
kiwisolver==1.4.4
matplotlib==3.7.2
ml-dtypes==0.2.0
numpy==1.25.1
opt-einsum==3.3.0
optax==0.1.5
packaging==23.1
Pillow==10.0.0
pyparsing==3.0.9
python-dateutil==2.8.2
scipy==1.11.1
six==1.16.0
toolz==0.12.0
typeguard==4.0.0
typing_extensions==4.7.1
zipp==3.16.2
```

One option would be to change in `integrate.py`: 
`dt0 = eqxi.error_if(dt0, pred, msg)` to ` dt0 = eqxi.error_if(jnp.array(dt0), pred, msg)`

patrick-kidger:
[...]

PR #283, opened 2023-07-27
bug fixed for init dt0

thibmonsel:
Added small fix in `integrate.py` mentionned in https://github.com/patrick-kidger/diffrax/issues/282.
`float` and `jnp.array` work for `dt0` now.


patrick-kidger:
LGTM! Thank you.

Issue #297, opened 2023-08-25
Interactively step through a solve - Example


VIVelev:
The example on [manual-stepping](https://docs.kidger.site/diffrax/usage/manual-stepping/) is not working as it is, resulting in an error:
```cmd
  File "/Users/vivelev/micromamba/envs/diff/lib/python3.11/site-packages/diffrax/solver/runge_kutta.py", line 698, in <lambda>
    lambda x: jnp.zeros((num_stages,) + x.shape, x.dtype), y0
                                        ^^^^^^^
AttributeError: 'int' object has no attribute 'shape'
```

One needs to change the line
```python
    y0 = 1
```
to
```python
    y0 = jnp.array([1.0])
```
to give `y0` a `shape` and a `dtype`.

I was wondering if maybe that could be done automatically, so that diffrax can work just as nicely with plain types.

Thanks for the great library btw!

patrick-kidger:
Ah, thanks for the report! I think I'll probably update the example rather than the implementation -- for simplicity I generally assume that everything after the `diffeqsolve` API is an array unless declared otherwise. (Just handling stuff like  `.shape` is easy to convert, of course, but there are several other edge cases where the distinction is actually important, often when using raw Python values to make things compile faster.)

I'd be happy to take a PR on this / if not I'll fix the example myself in a bit!

VIVelev:
I'll be happy to submit a PR!

patrick-kidger:
Closing as fixed in #298. Thank you for the contribution! The online docs will pick up the change and update shortly.

PR #298, opened 2023-08-25
Fix manual-stepping example (#297)

VIVelev:
#297


patrick-kidger:
Marvellous! Thank you for the fix.

PR #305, opened 2023-09-01
Patched edge-case in `LinearInterpolation`

packquickly:
Fixes an edge case where `dfx.LinearInterpolation` will return a `nan` value when input arrays have a repeated leading constant value. ie. the following example will return a `nan` value in current Diffrax:

```python
import diffrax
import jax.numpy as jnp

ts = jnp.array([1., 1., 1., 2., 3.])
ys = jnp.array([2., 2., 2., 3., 4.])

interpolator = diffrax.LinearInterpolation(ts, ys)
nan_result = interpolator.evaluate(1.)
```

Note the interpolator will only get `nan` evaluated on the repeated value.


patrick-kidger:
Right now this is deliberate behaviour in Diffrax. Having repeated times doesn't make a lot of sense in general?

packquickly:
I find either behavior reasonable. To me, multiple consistent values like this seems fine from the interpolation perspective, but maybe not from the diffeq perspective.

This type of data came up for me when using the Diffrax interpolators outside of standard diffeq solves.

Do note though that either way something like
```python
ts = jnp.array([0., 1., 1., 1., 2., 3.])
ys = jnp.array([1., 2., 2., 2., 3., 4.])
```
will give a non-nan value for `interpolator.evaluate(1.)`.

packquickly:
Okay!

patrick-kidger:
Actually, I think the numerics here are slightly wrong. From the docstring for `linear_rescale`: `"Calculates (t - t0) / (t1 - t0)"`. But we want `fractional_part / (t_next - t_prev)`.

Issue #317, opened 2023-10-02
[Question] Torchsde to diffrax conversion


Sinestro38:
I'm trying to convert this simple SDE that converges in torchsde to run on diffrax and produce corroborating results but I'm seeing very different results on diffrax with all of the solvers available.

The torchsde code:
```
import numpy as np
import torch
import torchsde

def arbitrary_force_fn(grid_x, l_bar = 1, rho = 1.5, lambda_th = 1, phi_ext = np.pi/16 ):
    return 2 * l_bar/(lambda_th**2) * (grid_x - lambda_th * rho * torch.sin(grid_x/lambda_th + phi_ext))

batch_size = 100000
init_states = torch.broadcast_to(torch.Tensor([0.0, -1]), (batch_size, 2))
times = torch.linspace(start=0, end=20, steps=1000)

class MyParticle(torch.nn.Module):
    def __init__(self, force_fn):
        super().__init__()
        self.force_fn = force_fn
        self.noise_type = "diagonal"
        self.sde_type = "ito"
        self.noise_vec = torch.nn.Parameter(torch.tensor([0, np.sqrt(2)]))

    def f(self, _, y):
        x = y[..., 0]
        p = y[..., 1]
        return torch.stack([p, -self.force_fn(x) - p], dim=-1)

    def g(self, _, y):
        return torch.broadcast_to(self.noise_vec, y.shape)

sde = MyParticle(arbitrary_force_fn)
device_cuda = torch.device("cuda")
with torch.no_grad():
    solns = torchsde.sdeint(sde.to(device_cuda), init_states.to(device_cuda), times.to(device_cuda), adaptive=True)
solns = torch.swapaxes(solns, 0, 1) # of shape (batch_size, times, 2) after swapping
```

And averaging over the batch size dimension produces the following correct result:
![image](https://github.com/patrick-kidger/diffrax/assets/70229100/55059e99-9810-4e5a-8a18-6a86c0a9f40d)

However, converting the same SDE to diffrax appears to produce very different results, and takes much more time.
![image](https://github.com/patrick-kidger/diffrax/assets/70229100/e1d619d0-b3b7-4523-bede-3f6285467484)

This was the diffrax code I ran to implement the same SDE:
```
import jax.random as jr
import jax.numpy as jnp
from diffrax import diffeqsolve, ReversibleHeun, ControlTerm, MultiTerm, ODETerm, SaveAt, VirtualBrownianTree, PIDController

[...]

PR #320, opened 2023-10-09
Fixed SDEs being unnecessarily slow to solve when max_steps!=None.

patrick-kidger:
The main changes needed to make this happen are in
https://github.com/patrick-kidger/equinox/pull/548
and as such this commit is fairly small -- it declares a dependency on
a new (as yet unreleasd) version of Equinox, and removes the
compatibility shim that was there before.

**How things used to be.**

To explain what's going on here a little more carefully: JAX only
recently added support for communicating to a `custom_vjp` which
input arguments were being perturbed, and which output cotangents were
symbolic zeros. Prior to that, a `custom_vjp` basically just had to
differentiate all inexact arrays.

However, there are some nondifferentiable inexact arrays, in the
sense that attempting to differentiate them will raise an error.
Solving SDEs has one such array: the nondifferentiable input to a
VirtualBrownianTree, which is guarded by a `eqxi.nondifferentiable` to
reflect the fact that Brownian motion is nondifferentiable.

So it used to be the case that the `custom_vjp` underlying
`RecursiveCheckpointAdjoint` would differentiate the overall
make-a-step function with respect to all inexact arrays,
including the time variable, hit the `nondifferentiable` guard, and
crash.

One (unsafe) fix would have just been to remove the `nondifferentiable`
guard. In practice I previously took the slower, safer option: silently
switch out `RecursiveCheckpointAdjoint` for a `DirectAdjoint`. The
latter is much less efficient, but uses no `custom_vjp`, and thus used
the perturbation and symbolic-zero propagation rules already present
in JAX's AD machinery, and was thus safe to use here.

**How things are now.**

And so, what has now changed: JAX has now added support for tracking
which inputs are perturbed, and which cotangents are symbolic zeros.

The Equinox PR above uses this functionality to determine which parts
of the carry need to be differentiated; no longer is it just "all
inexact arrays".

And thus this Diffrax PR removes the compatibility shim that is no
longer needed.

**Implications.**

SDE solving should now be much faster. In particular this fixes the
speed issue reported in #317.

### DIFF
--- a/diffrax/integrate.py
+++ b/diffrax/integrate.py
@@ -10,7 +10,7 @@ import jax.numpy as jnp
 import jax.tree_util as jtu
 from jax.typing import ArrayLike
 
-from .adjoint import AbstractAdjoint, DirectAdjoint, RecursiveCheckpointAdjoint
+from .adjoint import AbstractAdjoint, RecursiveCheckpointAdjoint
 from .custom_types import Array, Bool, Int, PyTree, Scalar
 from .event import AbstractDiscreteTerminatingEvent
 from .global_interpolation import DenseInterpolation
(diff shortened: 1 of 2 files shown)


patrick-kidger:
(Ignore the failing CI, it just doesn't know about the as-yet-unreleased version of Equinox.)

PR #325, opened 2023-10-20
Update comment on adaptive step-sizing

packquickly:
I noticed the description of the comment on adaptive step-sizing defined δ_{n, n} as the inverse of what it looks to be in code. If I'm wrong, just close this pr


patrick-kidger:
Agreed, this looks right to me.

This gets me thinking, what happens if we swap the inverse and the norm? I feel like I've thought about that at some point but I don't recall if there's a good reason to do it one way rather than the other. (This way around does avoid a potential div-by-zero error.)

PR #326, opened 2023-10-29
Complex type support

Randl:
Rebase of https://github.com/patrick-kidger/diffrax/pull/197

### DIFF
--- a/diffrax/integrate.py
+++ b/diffrax/integrate.py
@@ -654,6 +654,7 @@ def diffeqsolve(
     t1 = jnp.asarray(t1, dtype=dtype)
     if dt0 is not None:
         dt0 = jnp.asarray(dt0, dtype=dtype)
+    timelikes.append(dtype)
 
     def _get_subsaveat_ts(saveat):
         out = [s.ts for s in jtu.tree_leaves(saveat.subs, is_leaf=_is_subsaveat)]
@@ -779,7 +780,9 @@ def diffeqsolve(
             solver.step, terms, tprev, tnext, y0, args, solver_state, made_jump
         )
         dense_ts = jnp.full(max_steps + 1, jnp.inf)
-        _make_full = lambda x: jnp.full((max_steps,) + jnp.shape(x), jnp.inf)
+        _make_full = lambda x: jnp.full(
+            (max_steps,) + jnp.shape(x), jnp.inf, dtype=x.dtype
+        )
         dense_infos = jtu.tree_map(_make_full, dense_info)
         dense_save_index = 0
     else:
--- a/diffrax/integrate.py
+++ b/diffrax/integrate.py
@@ -760,7 +760,9 @@ def diffeqsolve(
         save_index = 0
         ts = jnp.full(out_size, jnp.inf)
         struct = eqx.filter_eval_shape(subsaveat.fn, t0, y0, args)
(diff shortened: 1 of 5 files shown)


diffrax/integrate.py
```diff
@@ -654,6 +667,7 @@ def diffeqsolve(
     t1 = jnp.asarray(t1, dtype=dtype)
     if dt0 is not None:
         dt0 = jnp.asarray(dt0, dtype=dtype)
+    timelikes.append(dtype)
```
patrick-kidger, line 670: Nit: this line is unnecessary, as `timelikes` isn't used after this point.

diffrax/nonlinear_solver/base.py
```diff
@@ -106,4 +106,6 @@ def jac(fn: Callable, x: PyTree, args: PyTree) -> LU_Jacobian:
         if not jnp.issubdtype(flat, jnp.inexact):
             # Handle integer arguments
             flat = flat.astype(jnp.float32)
-        return jsp.linalg.lu_factor(jax.jacfwd(curried)(flat))
+        return jsp.linalg.lu_factor(
+            jax.jacfwd(curried, holomorphic=jnp.iscomplexobj(flat))(flat)
+        )
```
patrick-kidger, line 111: Given the guard against implicit solvers earlier, is this change necessary?

(I think we'd want a lot of tests to be sure that the nonlinear solvers handle complex numbers correctly, what with all the linear solves and such going on.)

patrick-kidger (commented):
I think there's a couple of unnecessary changes. Other than that, LGTM!

diffrax/nonlinear_solver/base.py
Randl: With this change, implicit solvers are working for me locally. If anything, I'd remove the guard?

patrick-kidger: So as it happens, I'm currently in the process of upgrading Diffrax to use [Lineax](https://github.com/google/lineax) and [Optimistix](https://github.com/patrick-kidger/optimistix) to handle the linear and nonlinear solving. This is going to be great from a feature point of view, but it does also mean that any changes we make here will be short-lived.

Other than that, I think we'd need some fairly comprehensive tests to check that things are working. There's subtleties here that can easily go wrong, e.g. transpose vs conjugate-transpose. (JAX does the former, PyTorch does the latter.)

To justify the above, btw -- the position I've been taking with the JAX sciML ecosystem is that it's better not to do something than it is to do it unreliably. Hence why complex support has been such a slow-moving thing to have happen!

Randl: I agree with you regarding taking slow part. But this code otherwise will just crash, and it passes basic tests. If you insist I'll remove it, or reinstate the guard

patrick-kidger: Yup, I'd reinstate the guard for now. (As I say, the current implementation will be deleted in ~2 weeks anyway.)
I think getting complex support into Lineax will be the main thing needed to get proper support here.

Randl: ok I've put it back

patrick-kidger:
Closing in favour of #330, which fixes the pre-commit failure present here.

Thank you to all of @TimonHoess @Randl @andyElking for your efforts getting this in!

PR #327, opened 2023-11-02
fixing bugs 323 and 324

JadM133:
I fixed both bugs 323 and 324 and tested linear, rectilinear and cubic interpolations with both empty and random matrices. Tests are passing but I am getting 4 warnings (which I assume are not related to the changes since they're related to some deprecation.

### DIFF
--- a/test/test_global_interpolation.py
+++ b/test/test_global_interpolation.py
@@ -394,3 +394,49 @@ def test_dense_interpolation_vmap(solver, getkey):
         diffrax.Ralston: 1e-3,
     }.get(type(solver), 1e-6)
     assert shaped_allclose(derivs, true_derivs, atol=deriv_tol, rtol=deriv_tol)
+
+
+@pytest.mark.parametrize("mode", ["linear", "rectilinear", "cubic"])
+@pytest.mark.parametrize(
+    "nans, expected",
+    [
+        (
+            jnp.array([0, 3, 4, 6, 9]),
+            jnp.array([20.0, 1.0, 2.0, 23.0, 24.0, 5.0, 26.0, 7.0, 8.0, 29.0]),
+        ),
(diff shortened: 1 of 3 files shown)


test/test_global_interpolation.py
```diff
@@ -394,3 +394,49 @@ def test_dense_interpolation_vmap(solver, getkey):
...
+            jnp.array([20.0, 1.0, 2.0, 23.0, 24.0, 5.0, 26.0, 7.0, 8.0, 29.0]),
+        ),
+        (jnp.arange(0, 10, 1), jnp.arange(20, 30, 1)),
+    ],
+)
+@pytest.mark.parametrize("init_nan", [True, False])
+def test_replace_nans_at_start(mode, nans, expected, init_nan):
+    ts = jnp.linspace(0, 1, 15)
+    if init_nan:
+        ys = jnp.empty((15, 10)).at[:].set(jnp.nan)
```
patrick-kidger, line 414: I think this can be simplified to `ys = jnp.full((15, 10), jnp.nan)`.

patrick-kidger (commented):
It looks like the autoformatter isn't happy -- can you take a look at [CONTRIBUTING.md](https://github.com/patrick-kidger/diffrax/blob/main/CONTRIBUTING.md)?

Other than this, this looks great to me, I think it solves both issues perfectly. Thank you for taking the time to tackle this!

JadM133:
Thank you for the comment! I agree, using jnp.full is a better way. I fixed this and followed the instructions in CONTRIBUTING.md.

patrick-kidger:
Hmm, this is a little odd -- we seem to be seeing a flake in an unrelated test. Your changes themselves seem fine.
I've not seen this issue before, so possibly this is because of a change somewhere upstream in JAX?

PR #328, opened 2023-11-03
Add RATTLE solver 

krzysztofrusek:
This PR adds RATTLE solver for constrained Hamiltonian dynamics. It's very similar to #303 with the addition of implicit constrain. 
I use this solver for HMC on implicitly defined manifolds in [jax_chmc](https://github.com/krzysztofrusek/jax_chmc) and it may be useful to have it in diffrax.

This is the first solver with constraints so I am not sure if the abstraction is correct.
Since the constrain function is used in the solver step, maybe it's better to introduce another type of `Term`?

### DIFF
--- a/diffrax/__init__.py
+++ b/diffrax/__init__.py
@@ -71,6 +71,7 @@ from .solver import (
     Midpoint,
     MultiButcherTableau,
     Ralston,
+    Rattle,
     ReversibleHeun,
     SemiImplicitEuler,
     Sil3,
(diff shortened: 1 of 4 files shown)


patrick-kidger:
Thank you for the contribution!

The introduction of a constraint here is pretty interesting. As it happens, I'm actually aiming to have Diffrax support semi-explicit DAEs at the `diffeqsolve` level -- for all kinds of solvers. (I'm hoping to get that in in a couple of releases, maybe around the new year.) Given that, I'm thinking it might make most sense to hold off on merging this until I've got a draft of that going -- I can tag you on that to get your thoughts once it's done.

WDYT?

krzysztofrusek:
Ok,  lets wait for you to finish and I would love to give it a try - epscially for symplectic integrators.

PR #331, opened 2023-11-08
Pr branch

andyElking:


diffrax/solver/ansr.py
```diff
@@ -0,0 +1,220 @@
...
+        assert self.ch.shape[0] == self.b.shape[0]
+        for i, (a_i, c_i) in enumerate(zip(self.a, self.c)):
+            assert np.allclose(sum(a_i), c_i)
+        assert np.allclose(sum(self.b), 1.0)
+
+        # TODO: add checks for whether the method is FSAL
+
+
+class ANSR(AbstractItoSolver):
+    r"""Additive-Noise Stochastic Runge-Kutta method.
```
patrick-kidger, line 76: By the way, what's the reason for restricting to only additive noise? I think the SRK schemes themselves don't change that much.

diffrax/solver/ansr.py
```diff
@@ -0,0 +1,220 @@
...
+
+        h = t1 - t0
+        drift, diffusion = terms.terms
+        check_diffusion_has_stla(diffusion)
+
+        # compute the Brownian increment and space-time Levy area
+        levy: LevyVal = diffusion.levy_contr(t0, t1)
+        w = levy.W
+        hh = levy.H
+        sigma = diffusion.vf(t0, y0, args)
```
patrick-kidger, line 151: If you decide to stick with specifically additive noise, then that's actually an invariant we can check:
```python
sigma, t_sigma = eqx.filter_jvp(lambda y: diffusion.vf(t0, y, args), y0)
if t_sigma is not None:
    raise ValueError
```
The above checks that `t_sigma` is returning a symbolic zero tangent when differentiated, i.e. it doesn't depend on the `y0` input.

diffrax/solver/ansr.py
```diff
@@ -0,0 +1,220 @@
...
+            hk_j = drift.vf_prod(t0 + c_j * h, z_j, args, h)
+            hks_j = jtu.tree_map(
+                lambda ks_leaf, k_j_leaf: ks_leaf.at[j].set(k_j_leaf), hks_j, hk_j
+            )
+            # note that carry will already contain the whole stack of
+            # k_js, so no need for second return value
+            return (j + 1, hks_j), None
+
+        a = self.embed_a_lower(jnp.dtype(y0))
+        c = jnp.insert(jnp.asarray(self.tableau.c, dtype=jnp.dtype(y0)), 0, 0.0)
```
patrick-kidger, line 186: You can make this potentially slightly more efficient by doing the insert in `np` before casting to `jnp`. (Essentially this will force the insert to happen at compile time.)

diffrax/solver/ansr.py
```diff
@@ -0,0 +1,220 @@
...
+        b = jnp.asarray(self.tableau.b, dtype=jnp.dtype(y0))
```
patrick-kidger, line 187: These `jnp.dtype` calls will fail for pytree-valued `y0`.

diffrax/solver/ansr.py
```diff
@@ -0,0 +1,220 @@
...
+        cw_last = jnp.asarray(self.tableau.cw_last, dtype=jnp.dtype(y0))
+        ch_last = jnp.asarray(self.tableau.ch_last, dtype=jnp.dtype(y0))
+        # hks is a PyTree of the same shape as y0, except that the arrays inside have
+        # an additional batch dimension of size len(b) (i.e. num stages)
+        hks = jtu.tree_map(
+            lambda leaf: jnp.zeros(shape=(len(b),) + leaf.shape, dtype=leaf.dtype), y0
+        )
+        carry = (0, hks)
+
+        # output of lax.scan is ((num_stages, hks), [None, None, ... , None])
```
patrick-kidger, line 199: Nit: it will just be a `None`, not a `[None, ...]`.

diffrax/solver/ansr.py
```diff
@@ -0,0 +1,220 @@
...
+        assert self.cw.shape[0] == self.b.shape[0]
+        assert self.ch.shape[0] == self.b.shape[0]
+        for i, (a_i, c_i) in enumerate(zip(self.a, self.c)):
+            assert np.allclose(sum(a_i), c_i)
+        assert np.allclose(sum(self.b), 1.0)
+
+        # TODO: add checks for whether the method is FSAL
+
+
+class ANSR(AbstractItoSolver):
```
patrick-kidger, line 75: Btw, for consistency I try to name abstract classes beginning with `Abstract`.

diffrax/solver/ansr.py
```diff
@@ -0,0 +1,220 @@
...
+
+    term_structure = MultiTerm[Tuple[ODETerm, ControlTerm]]
+    interpolation_cls = LocalLinearInterpolation
+    tableau: StochasticButcherTableau
+
+    def order(self, terms):
+        return 1  # should be modified depending on tableau
+
+    def strong_order(self, terms):
+        return 0.5  # should be modified depending on tableau
```
patrick-kidger, line 108: I'd suggest leaving these out here if they get overriden on a per-solver basis.

diffrax/term.py
```diff
@@ -413,6 +426,10 @@ def is_vf_expensive(
         _t1 = jnp.where(self.direction == 1, t1, -t0)
         return self.term.is_vf_expensive(_t0, _t1, y, args)
 
+    def levy_contr(self, t0: Scalar, t1: Scalar) -> (PyTree, PyTree):
+        assert isinstance(self.term, _ControlTerm)
+        return self.term.levy_contr(t0, t1)
```
patrick-kidger, line 431: I think I'd suggest removing these extra methods and instead just have `contr` pipe any `**kwargs` forward.

patrick-kidger (commented):
I like this! I think you've done things very cleanly overall.

I've left a fair few comments (it's a large PR); sometimes they apply to multiple places. E.g. I haven't called out every place where a method should be private.

diffrax/brownian/tree.py
```diff
@@ -34,12 +42,61 @@ class _State(eqx.Module):
...
+def wj_to_wh_diff(x0: LevyVal, x1: LevyVal) -> LevyVal:
+    """
+    Computes (W_{s,u}, H_{s,u}) from (W_s, J_s) and (W_u, J_u),
+    where J_u = ∫_0^u W_t dt
+    Args:
+        x0: LevyVal(W_s, J_s)
+        x1: LevyVal(W_u, J_u)
+    """
+    h = (x1.h - x0.h).astype(x0.W.dtype)
+    inverse_h = jnp.nan_to_num(1 / h).astype(x0.W.dtype)
```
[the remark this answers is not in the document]
andyElking: I think without this it wasn't passing one of the tests. But I think that after I added interval normalisation, I was more principled about the types, so yes, this should no longer be required.

diffrax/brownian/tree.py
```diff
@@ -70,10 +128,13 @@ def __init__(
         tol: Scalar,
         shape: Union[Tuple[int, ...], PyTree[jax.ShapeDtypeStruct]],
         key: "jax.random.PRNGKey",
+        compute_stla: bool = False,
     ):
-        self.t0 = t0
-        self.t1 = t1
-        self.tol = tol
+        self.t0 = jnp.minimum(t0, t1)
+        self.t1 = jnp.maximum(t0, t1)
```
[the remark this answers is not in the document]
andyElking: Should be fine without, but I think it is more transparent this way. Additionally, we use sqrt(t1 - t0) at various points, and we do not unnecessary imaginary values. I could add abs(t1 - t0), but that would probably only make the code more confusing.

diffrax/brownian/tree.py
```diff
@@ -88,69 +149,158 @@ def __init__(
...
+
+        def mult(z):
+            return (z * self._interval_len).astype(z.dtype)
+
+        return LevyVal(
+            h=jtu.tree_map(mult, x.h),
+            W=jtu.tree_map(sqrt_mult, x.W),
+            J=jtu.tree_map(mult, x.J),
+            H=jtu.tree_map(sqrt_mult, x.H),
+        )
```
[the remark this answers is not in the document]
andyElking: self.t1 and self.t0 can be of any type. The values in the PyTree however, must have dtypes as specified in self.shape. The latter is itself a PyTree, and each leaf can have different dtypes. So, regardless of the dtype of self.t1, self.t0, each leaf must keep the dtype it had before.

diffrax/brownian/path.py
```diff
@@ -72,14 +79,38 @@ def evaluate(self, t0: Scalar, t1: Scalar, left: bool = True) -> PyTree[Array]:
...
+        w = jrandom.normal(key_w, shape.shape, shape.dtype) * jnp.sqrt(t1 - t0).astype(
             shape.dtype
         )
+        if not use_levy:
+            return w
+        hh = None
+        if compute_stla:
+            hh = jrandom.normal(key_hh, shape.shape, shape.dtype) * jnp.sqrt(
+                (t1 - t0) / 12
+            ).astype(shape.dtype)
```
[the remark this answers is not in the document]
andyElking: > I think `use_levy=True` and `compute_stla=False` should probably just raise an error?

I don't think that's necessary. I have made the improvement you suggested in the comment up above and now the code is very clean for all 4 combinations of use_levy and spacetime_levyarea.

diffrax/term.py
andyElking: I wanted to make it explicit that a given method will always return LevyVal. But I suppose we can get rid of that. I'll just add the following check to all solvers requiring STLA:
```levy = diffusion.contr(t0, t1, use_levy=True)
        assert isinstance(levy, LevyVal) and (levy.H is not None), \
            ("The diffusion should be a ControlTerm controlled by either a"
             "VirtualBrownianTree or an UnsafeBrownianPath with"
             "`spacetime_levyarea` set to True.")
```

diffrax/solver/align.py
```diff
@@ -0,0 +1,300 @@
...
+        # jax.debug.print("y0 {a}", a=y0.dtype)
+        γ, u, f = args  # f is in fact grad(f)
+        h = t1 - t0
+
+        tay_cfs = comp_taylor_coeffs(γ, u)
+        coeffs = self.recompute_coeffs(h, γ, u, tay_cfs)
+
+        assert y0.ndim == 1
+        dim = int(y0.shape[0] / 2)
+        assert y0.shape[0] == 2 * dim
```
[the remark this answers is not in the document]
andyElking: If I want to separate the Langevin variables into a tuple (x, v), then the Brownian control must also be a tuple, with both entries being d-dimensional (where d=dim(x)). But, x does not interact with the BM at all (i.e. it is only influenced by the drift, not the diffusion), so half of the Brownian dimensions would just be doing unnecessary computation. 

The possible resolutions I saw were filling out the control corresponding to x with a dummy constant path, or introducing a new kind of _ControlTerm. I opted for the latter and introduced a LangevinTerm, as you suggested above. I will update the PR shortly and you can see if you agree with that approcah.

Additionally, in ALIGN I do not intend to allow y to be an arbitrary PyTree. That has no meaningful physical or mathematical interpretation. Unlike ANSR, which supports arbitrary PyTrees, ALIGN is not designed to be a general method, and as such 1) the user should not be allowed to treat it as one and 2) it is a waste of time trying to make it accept general inputs, if nobody would use it that way.

diffrax/solver/align.py
```diff
@@ -0,0 +1,300 @@
...
+    ) -> _SolverState:
+        """
+        Precompute _SolverState which carries the Taylor coefficients and the
+        ALIGN coefficients (which can be computed from h and the Taylor coeffs).
+        This method is FSAL, so _SolverState also carries the previous evaluation
+        of grad_f.
+        """
+        # print(y0.dtype)
+        # jax.debug.print("y0 {a}", a=y0.dtype)
+        γ, u, f = args  # f is in fact grad(f)
```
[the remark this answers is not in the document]
andyElking: Sounds good!

diffrax/solver/ansr.py
```diff
@@ -0,0 +1,220 @@
...
+from ..solution import RESULTS
+from ..term import AbstractTerm, ControlTerm, MultiTerm, ODETerm, WrapTerm
+from .base import AbstractItoSolver
+
+
+_ErrorEstimate = None
+_SolverState = None
+
+
+def check_diffusion_has_stla(diffusion):
```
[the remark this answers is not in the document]
andyElking: Thanks, sounds like a good suggestion.

diffrax/solver/ansr.py
andyElking: I Intend to add that eventually, but for now this will have to do. The way Rößler defines his tableaus is incredibly convoluted, and from what James told me, they might be a bit of a dead end, so it isn't the highest priority on my list right now.

diffrax/__init__.py
```diff
@@ -40,6 +41,7 @@
 from .solution import is_event, is_okay, is_successful, RESULTS, Solution
 from .solver import (
     AbstractAdaptiveSolver,
+    AbstractANSR,
```
patrick-kidger, line 44: I'd suggest switching from `AbstractANSR` to `AbstractAdditiveSRK`, which I think is probably a clearer name.

diffrax/brownian/base.py
```diff
@@ -1,14 +1,49 @@
...
+    """Helper that takes a PyTree of LevyVals and transposes
+    into a LevyVal of PyTrees.
+
+    Args:
+        tree_shape:
+        spacetime_levyarea:
+        tree:
+
+    Returns:
+
```
patrick-kidger, line 22: Can you fill in the docstring here? (And using the same style as the rest of Diffrax, i.e. markdown, rather than the Google-style stub currently here.)

diffrax/brownian/base.py
```diff
@@ -1,14 +1,49 @@
...
+        ),
+        pytree_to_transpose=tree,
+    )
+
+
 class AbstractBrownianPath(AbstractPath):
-    "Abstract base class for all Brownian paths."
+    """Abstract base class for all Brownian paths."""
+
+    spacetime_levyarea: bool = eqx.field(static=True)
```
patrick-kidger, line 37: I think it's generally an antipattern to put fields on abstract base classes.

Are you looking to declare that there must be a `spacetime_levyarea` attribute? (E.g. that the following should typecheck:
```python
def f(x: AbstractBrownianPath):
    y = x.spacetime_levyarea
```
)
or something else?

diffrax/brownian/base.py
```diff
@@ -1,14 +1,49 @@
...
-from ..custom_types import Array, PyTree, Scalar
+import equinox as eqx
+import jax
+from jax import tree_util as jtu
+
+from ..custom_types import Array, LevyVal, PyTree, Scalar
 from ..path import AbstractPath
 
 
+def _levy_tree_transpose(tree_shape, spacetime_levyarea, tree):
```
patrick-kidger, line 12: Remove the leading underscore, as this is imported into other files, so it isn't private.

diffrax/brownian/base.py
```diff
@@ -1,14 +1,49 @@
...
+
+    Args:
+        tree_shape:
+        spacetime_levyarea:
+        tree:
+
+    Returns:
+
+    """
+    hh_default_val = 0.0 if spacetime_levyarea else None
```
patrick-kidger, line 24: Should this maybe just always be a `None`?

diffrax/brownian/tree.py
```diff
@@ -34,12 +42,58 @@ class _State(eqx.Module):
...
+
+
+def _wj_to_wh_diff(x0: LevyVal, x1: LevyVal) -> LevyVal:
+    r"""Computes $(W_{s,u}, H_{s,u})$ from $(W_s, J_s)$ and
+    $(W_u, J_u)$, where $J_u = \int_0^u W_t dt$
+
+    **Arguments:**
+
+        - `x0`: LevyVal$(W_s, J_s)$
+        - `x1`: LevyVal$(W_u, J_u)$
```
patrick-kidger, line 57: Nits: dedent these two lines. Also, I'm not sure what's going on with the partial use of LaTeX. I think I'd suggest just writing it as code, `LevyVal(W_s, J_s)`, if it's annoying to express in LaTeX.

diffrax/brownian/tree.py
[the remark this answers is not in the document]
patrick-kidger: I think passing `t1 < t0` should probably just be an error.

You could do that at runtime using an `eqx.error_if`:
```python
t0, t1 = eqx.error_if((t0, t1), t1 < t0, "`VirtualBrownianTree(t0, t1, ...)` must satisfy `t0 < t1`")
```

diffrax/brownian/tree.py
```diff
@@ -70,10 +124,12 @@ def __init__(
...
         shape: Union[Tuple[int, ...], PyTree[jax.ShapeDtypeStruct]],
         key: "jax.random.PRNGKey",
+        spacetime_levyarea: bool = False,
     ):
-        self.t0 = t0
-        self.t1 = t1
-        self.tol = tol
+        self.t0 = jnp.minimum(t0, t1)
+        self.t1 = jnp.maximum(t0, t1)
+        self.tol = tol / (self.t1 - self.t0)
```
patrick-kidger, line 131: This is a change from previous behaviour?

diffrax/brownian/tree.py
```diff
@@ -88,69 +144,173 @@ def __init__(
...
+
+        **Arguments:**
+            - `t0`: Start of interval
+
+            - `t1`: End of interval
+
+            - `left`: ignored since Brownian motion is continuous
+
+            - `use_levy`: If true, then the return type is a LevyVal encapsulating
+            the Brownian increment and the corresponding Lévy area if it was computed.
```
patrick-kidger, line 186: I don't think you need a docstring here; it should inherit it from its parent class.

diffrax/brownian/tree.py
```diff
@@ -88,69 +144,173 @@ def __init__(
...
+        assert isinstance(levy_out, LevyVal)
+        return levy_out if use_levy else levy_out.W
+
+    def _evaluate(self, r: Scalar) -> PyTree[LevyVal]:
+        """Maps the _evaluate_leaf function at time τ using self.key onto self.shape"""
+        eqxi.error_if(
+            r,
+            r < 0,
+            "Cannot evaluate VirtualBrownianTree outside of its range [t0, t1].",
+        )
```
patrick-kidger, line 227: Here you need to use the return value: `r = eqx.error_if(r, ...)`. The first argument is returned with the same value, but with the error "threaded on to it". (That is, the error condition is checked at some point between the input `r` being computed and the output `r` being used. And if the output is unused then this check is never performed.)

diffrax/custom_types.py
```diff
@@ -131,3 +133,10 @@ def __class_getitem__(cls, item):
 DenseInfo = Dict[str, PyTree[Array]]
 DenseInfos = Dict[str, PyTree[Array["times", ...]]]  # noqa: F821
 sentinel: Any = eqxi.doc_repr(object(), "sentinel")
+
+
+class LevyVal(eqx.Module):
+    h: Scalar = field(default=0.0)
+    W: PyTree[Array] = field(default=None)
+    J: Optional[PyTree[Array]] = field(default=None)
+    H: Optional[PyTree[Array]] = field(default=None)
```
patrick-kidger, line 142: `field` isn't necessary if you're only specifying default values. Just set them as the default directly.

That said, are default values even needed here? I like to say that "default values are the root of unexpected behaviour", and as this is a type that is only instantiated internally, we can always just pass in every value and I think that shouldn't be too much work?

diffrax/path.py
```diff
@@ -39,16 +39,26 @@ def evaluate(self, t0, t1=None, left=True):
     t0: Scalar = field(init=False, repr=False)
     t1: Scalar = field(init=False, repr=False)
 
+    def eval_levy(
+        self, t0: Scalar, t1: Optional[Scalar] = None, left: bool = True
+    ) -> LevyVal:
+        return self.evaluate(t0, t1, left, use_levy=True)
```
patrick-kidger, line 45: I think we agreed this method shouldn't exist?

diffrax/path.py
```diff
@@ -39,16 +39,26 @@ def evaluate(self, t0, t1=None, left=True):
...
+
     @abc.abstractmethod
     def evaluate(
-        self, t0: Scalar, t1: Optional[Scalar] = None, left: bool = True
+        self,
+        t0: Scalar,
+        t1: Optional[Scalar] = None,
+        left: bool = True,
```
patrick-kidger, line 52: Spurious change?

diffrax/path.py
```diff
@@ -39,16 +39,26 @@ def evaluate(self, t0, t1=None, left=True):
...
     ) -> PyTree:
         r"""Evaluate the path at any point in the interval $[t_0, t_1]$.
 
         **Arguments:**
 
         - `t0`: Any point in $[t_0, t_1]$ to evaluate the path at.
+
         - `t1`: If passed, then the increment from `t1` to `t0` is evaluated instead.
+
```
patrick-kidger, line 61: Spurious change?

diffrax/brownian/base.py
andyElking: Yes, that was exactly the intention. It is not strictly needed except in some assert statements, but I think it would be useful to have. I suppose I can also use hasattr() first.

diffrax/brownian/tree.py
```diff
@@ -52,15 +109,16 @@ class VirtualBrownianTree(AbstractBrownianPath):
...
-        corrects a small bias in the generated samples.)
+        (The implementation here is a slight improvement on the reference implementation
+        by using an interpolation method which ensures all the 2nd moments are correct.)
     """
 
     t0: Scalar = field(init=True)
     t1: Scalar = field(init=True)  # override init=False in AbstractPath
-    tol: Scalar
+    _interval_len: Scalar
+    tol: Scalar = field(init=True)
```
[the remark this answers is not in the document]
andyElking: Oh sorry I missed this comment last time. Yes, I will get rid of that adjustment. Also see my reply below.

diffrax/brownian/tree.py
andyElking: Yes, because of normalising the interval. This ensures that the behaviour is the same in fact, but avoids computing tol/(t1-t0) each time in the main tree loop.

diffrax/brownian/tree.py
andyElking: Fair.

diffrax/brownian/tree.py
andyElking: Ohh, I see, thanks.

diffrax/custom_types.py
andyElking: That's fair. The main idea was that as new types of Levy area are added, there is no need to change all the code that uses LevyVals. But then again, that's pretty minor, so I can easily get rid of these defaults if you think that's the way to go.

diffrax/path.py
andyElking: Yes, we did, I apparently forgot to delete it. It is certainly not used anymore, I'm pretty sure.

diffrax/path.py
andyElking: This one was on purpose. It looked a bit funny in the docs. But yes, I can get rid of it.

diffrax/solver/align.py
```diff
@@ -0,0 +1,300 @@
...
+
+class ALIGN(AbstractItoSolver):
+    """The Adaptive Langevin via Interpolated Gradients and Noise method
+    designed by James Foster. Only works for Underdamped Langevin Diffusion
+    of the form
+    d x_t &= v_t dt
+    d v_t &= - γ v_t dt - u ∇f(x_t) dt + (2γu)^(1/2) dW_t
+    where v is the velocity, f is the potential, γ is the friction, and
+    W is a Brownian motion.
+    """
```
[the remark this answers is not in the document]
andyElking: I'm not entirely sure what you mean here. There aren't any args or returns, so I don't know what should be reformatted.

diffrax/solver/ansr.py
andyElking: In fact I am in the proccess of generalising this, expect an update of that together with the space-time-time Levy area sometime next month.

diffrax/solver/ansr.py
andyElking: I did add that check above.

diffrax/term.py
andyElking: Hmm I thought I removed this...

diffrax/brownian/base.py
andyElking: I don't think so, it should match the inner tree structure. And there is a leaf here (e.g. 0), whereas a None has no leaves. If I understand correctly at least? I could be wrong, you're the JAX expert.

diffrax/brownian/base.py
andyElking: I removed this. In addition, in VBT and UBP I swapped this out for levy_area: str, which can be either "", "space-time", or eventually "space-time-time".

diffrax/brownian/base.py
andyElking: Sounds good! I also moved it to custom_types right under the definition of LevyVal, since any changes to LevyVal (e.g. adding space-time-time Levy area) will have to be reflected in levy_tree_transpose, so it's nice to have them close.

diffrax/brownian/tree.py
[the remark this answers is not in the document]
andyElking: In fact I think this can be done in python, not in jaxpr, right?

diffrax/brownian/path.py
```diff
@@ -30,9 +30,14 @@ class UnsafeBrownianPath(AbstractBrownianPath):
...
     motion. Hence the restrictions above. (They describe the general case for which the
     correlation structure isn't needed.)
+
+    Depending on the `levy_area` argument, this can also be used to generate Levy area.
+    `levy_area` can be "", "space-time" or "space-time-time".
+
     """
 
     shape: PyTree[jax.ShapeDtypeStruct] = eqx.field(static=True)
+    levy_area: str = eqx.field(static=True)
```
patrick-kidger, line 40: The type annotation here should be `Literal["", "space-time", "space-time-time"]`. This helps to be much more precise than simply `str`.

diffrax/brownian/tree.py
```diff
@@ -164,64 +317,99 @@ def _cond_fun(_state):
...
-        # too small.)
+
+        # Based on the values of (W, H) at s<t<u (where t = (s+u)/2), we interpolate
+        # to obtain approximate values of (W_r, H_r) for all r ∈ [s,u]. This is done
+        # in a way that gives (W_r, H_r) all the correct first and second moments
+        # conditional on (W_s, H_s), and (W_u, H_u), where (W_t, H_t) is treated as
+        # the source of randomness.
+        # NOTE: this gives a different result than the original implementation of the
+        # VirtualBrownianTree by Patrick Kidger.
+
```
patrick-kidger, line 361: Let's remove this note, since this code lives under `patrick-kidger/diffrax` ;)

patrick-kidger (commented):
Okay, another partial review! I think we're most of the way there for the changes to the Brownian motions. I've also started to leave a few comments on the solvers.

diffrax/brownian/path.py
```diff
@@ -41,13 +46,20 @@ def __init__(
...
         key: "jax.random.PRNGKey",
+        levy_area: str = "",
     ):
         self.shape = (
             jax.ShapeDtypeStruct(shape, jax.dtypes.canonicalize_dtype(None))
             if is_tuple_of_ints(shape)
             else shape
         )
         self.key = key
+        if levy_area not in ["", "space-time"]:
```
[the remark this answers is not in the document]
andyElking: Yes, you are right. The issue was that I was simultaneously working on adding space-time-time, and I merged some changes from the branch that was further ahead, and forgot to remove this reference to "space-time-time".
As things stand now, there should be no more references to "space-time-time" in docstrings that will be displayed in the docs, but the code is structured with "space-time-time" in mind, because I don't think it's worth reverting all the changes just to add them back in later. Still I made sure to delete all the code that is "space-time-time" only. I hope that's okay.

diffrax/brownian/tree.py
andyElking: Yep.

diffrax/brownian/tree.py
```diff
@@ -88,69 +137,173 @@ def __init__(
...
         if t1 is None:
-            return self._evaluate(t0)
+
+            if self.levy_area == "space-time":
+                # set tH_t to None
+                levy_out = jtu.tree_map(
+                    lambda x: dataclasses.replace(x, tH_t=None), levy_0
+                )
+            else:
+                levy_out = levy_0
```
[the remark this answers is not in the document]
andyElking: Fair.

diffrax/solver/additive_srk.py
```diff
@@ -0,0 +1,284 @@
...
+        assert all(i + 1 == a_i.shape[0] for i, a_i in enumerate(self.a))
+        assert (self.b_error is None) or self.b_error.shape[0] == self.b_sol.shape[0]
+        assert self.c.shape[0] + 1 == self.b_sol.shape[0]
+        assert self.cW.shape[0] == self.b_sol.shape[0]
+        assert self.cH.shape[0] == self.b_sol.shape[0]
+        for i, (a_i, c_i) in enumerate(zip(self.a, self.c)):
+            assert np.allclose(sum(a_i), c_i)
+        assert np.allclose(sum(self.b_sol), 1.0)
+
+        # TODO: add checks for whether the method is FSAL
```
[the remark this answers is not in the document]
andyElking: ALIGN, for example is FSAL, and perhaps it could appear. But yes, you're right, FSAL is just not supported. This is a little remnent of when I thought I might implement it soon, but I realised it is not really relevant for the solvers that I currently have, so no point adding it. Anyway, I just deleted that.

But you'll be happy to know that I implemented embedded error estimates for both ShARK and SRA1, so adaptive time stepping works with both. And same for ALIGN.

diffrax/solver/additive_srk.py
```diff
@@ -0,0 +1,284 @@
...
+    \, (c^W_j W_{t_0, t_1} + c^H_j H_{t_0, t_1})$
+
+    where $W_{t_0, t_1}$ is the increment of the Brownian motion and
+    $H_{t_0, t_1}$ is its corresponding space-time Lévy Area.
+
+    The values $( a_{i,j} ) , b_j, c_j, c^W_j, c^H_j$ are provided
+    in the `StochasticButcherTableau`.
+    """
+
+    term_structure = MultiTerm[Tuple[ODETerm, _ControlTerm]]
```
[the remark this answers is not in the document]
andyElking: Fair enough :)

diffrax/solver/additive_srk.py
```diff
@@ -0,0 +1,284 @@
...
+        # Check that the diffusion has levy_area="space-time"
+        _, diffusion = terms.terms
+        is_bm = lambda x: isinstance(x, AbstractBrownianPath)
+        leaves = jtu.tree_leaves(diffusion, is_leaf=is_bm)
+        paths = [x for x in leaves if is_bm(x)]
+        for path in paths:
+            if not path.levy_area == "space-time":
+                raise ValueError(
+                    "The Brownian path controlling the diffusion "
+                    "should be initialised with `compute_stla=True`"
```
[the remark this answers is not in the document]
andyElking: Yep, good point. I tried to replace it everywhere, but this one flew under the radar.

diffrax/solver/additive_srk.py
```diff
@@ -0,0 +1,284 @@
...
+
+        # check that the vector field of the diffusion term is constant
+        sigma, (t_sigma, y_sigma) = eqx.filter_jvp(
+            lambda t, y: diffusion.vf(t, y, args), (t0, y0), (t0, y0)
+        )
+        if (t_sigma is not None) or (y_sigma is not None):
+            raise ValueError(
+                "Vector field of the diffusion term should be constant, "
+                "independent of t and y."
+            )
```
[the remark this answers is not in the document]
andyElking: The way it is currently set up, time dependence is not supported. I will modify this in the next PR accordingly.

diffrax/solver/align.py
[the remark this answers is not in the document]
andyElking: Sorry, but I'm still unsure what you meant here.

diffrax/brownian/path.py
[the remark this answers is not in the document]
patrick-kidger: That sounds perfect!

diffrax/brownian/tree.py
```diff
@@ -70,10 +111,18 @@ def __init__(
         tol: Scalar,
         shape: Union[Tuple[int, ...], PyTree[jax.ShapeDtypeStruct]],
         key: "jax.random.PRNGKey",
+        levy_area: str = "",
     ):
+        if t0 >= t1:
+            raise ValueError("t0 must be strictly less than t1.")
```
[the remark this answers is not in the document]
patrick-kidger: If they're not JAX arrays then I the `t0 >= t1` condition should be evaluated statically, in which case the `error_if` will eagerly raise at trace-time rather than runtime. Thus it should always work :)

diffrax/solver/additive_srk.py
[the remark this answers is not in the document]
patrick-kidger: Great!

diffrax/solver/additive_srk.py
```diff
@@ -0,0 +1,284 @@
...
+
+class AbstractAdditiveSRK(AbstractStratonovichSolver):
+    r"""Additive-Noise Stochastic Runge-Kutta method.
+
+    The second term in the MultiTerm must be a `ControlTerm` with
+    `control=VirtualBrownianTree(levy_area="space-time")`, since this method
+    makes use of space-time Lévy area.
+
+    Given the SDE
+    $dX_t = f(t, X_t) dt + σ dW_t$
```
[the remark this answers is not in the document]
patrick-kidger: Just for my own education: is there actually any scenario in which time-dependent and time-independent additive noise need to be treated differently?

As for a general SRK -- if this is landing later, perhaps we should remove `AbstractAdditiveSRK` from the public interface now, on the assumption it will be removed later?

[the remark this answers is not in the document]
patrick-kidger: On your first paragraph: no strong feelings; up to you.

But yes, definitely do remove it from the docs / public API if it's not going to be long for this world!

---

Do you think it would it be worth splitting this PR into two pieces? Keep just the BM changes here, and then do the solvers separately? The BM changes are now very nearly ready, and I'm thinking it might not be worth ironing out the details of the solvers if they're going to be changing soon anyway. (The bottleneck here is clearly my review time, so anything that minimises that is good :) )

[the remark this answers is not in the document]
andyElking: Yes, I can try to split it, that's probably a good idea. What's the best way to accomplish this? Should I just start from a fresh branch (up to date with just main Diffrax) and then add in all the brownian-related files, and not the solvers? There are some files (e.g. term.py) which have changes relating to both one and the other, and I think it would be a mess trying to separate all of that out. What do you suggest (mostly in terms of the git procedure + useful commands)?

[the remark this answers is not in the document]
patrick-kidger: > Should I just start from a fresh branch (up to date with just main Diffrax) and then add in all the brownian-related files, and not the solvers? 

Yup, exactly.

For `term.py`, let's leave that out for now as it isn't reviewed yet.

In terms of git commands: the simplest way is probably to copy-paste the browian files somewhere else, then create a new branch, then copy them back!

PR #337, opened 2023-12-07
added space-time levy area (for PR)

andyElking:
Hi Patrick,

This is a new PR with just the Brownian things. I tried to keep it as contained as possible.


diffrax/brownian/tree.py
```diff
@@ -25,21 +25,75 @@
...
+
+    `LevyVal(W_su, H_su)`
+    """
+
+    su = (x1.dt - x0.dt).astype(x0.W.dtype)
+    w_su = x1.W - x0.W
+    if x0.H is None or x1.H is None:  # BM only case
+        return LevyVal(dt=su, W=w_su, H=None, bar_H=None, K=None, bar_K=None)
+
+    # levy_area == "space-time"
```
patrick-kidger, line 69: ?

diffrax/brownian/tree.py
```diff
@@ -25,21 +25,75 @@
...
+    _su = jnp.where(jnp.abs(su) < jnp.finfo(su).eps, jnp.inf, su)
+    inverse_su = 1 / _su
+    u_bb_s = x1.dt * x0.W - x0.dt * x1.W
+    bhh_su = x1.bar_H - x0.bar_H - 0.5 * u_bb_s  # bhh_su = H_{s,u} * (u-s)
+    hh_su = inverse_su * bhh_su
+
+    return LevyVal(dt=su, W=w_su, H=hh_su, bar_H=None, K=None, bar_K=None)
+
+
+def split_interval(_cond, x_stu, x_st_tu):
```
patrick-kidger, line 79: If this is a helper private to this file then begin it with an underscore: `_split_interval`.

diffrax/brownian/tree.py
```diff
@@ -88,70 +152,224 @@ def __init__(
...
+            "Cannot evaluate VirtualBrownianTree outside of its range [t0, t1].",
+        )
+        r = eqxi.error_if(
+            r,
+            r > 1,
+            "Cannot evaluate VirtualBrownianTree outside of its range [t0, t1].",
+        )
+        # Clip because otherwise the while loop below won't terminate, and the above
+        # errors are only raised after everything has finished executing.
+        r = jnp.clip(r, 0, 1)
```
patrick-kidger, line 226: I know you were copying this from what's already here, but actually this is no longer true! Let's fix this up whilst we're here. We can remove this line.
Likewise, we can actually improve efficiency a little bit by having just a single `error_if` combining both conditions:
```python
  r = eqxi.error_if(
      r,
      (r < 0) | (r > 1),
      "Cannot evaluate VirtualBrownianTree outside of its range [t0, t1].",
  )
```

diffrax/brownian/tree.py
```diff
@@ -88,70 +152,224 @@ def __init__(
...
+
+        **Returns:**
+            - `t`: midpoint time
+            - `w_t`: value of BM at t
+            - `w_st_tu`: $(W_{s,t}, W_{t,u})$
+            - `bhh`: (optional) $(\bar{H}_s, \bar{H}_t, \bar{H}_u)$
+            - `bhh_st_tu`: (optional) $(\bar{H}_{s,t}, \bar{H}_{t,u})$
+            - `bkk_t`: (optional) $(\bar{K}_s, \bar{K}_t, \bar{K}_u)$
+            - `bkk_st_tu`: (optional) $(\bar{K}_{s,t}, \bar{K}_{t,u})$
+
```
patrick-kidger, line 272: Nit: these should be formatted as
```python
**Arguments:**

- `s`: ...
- `u`: ...
```
i.e. with a newline and without an indent.

diffrax/brownian/tree.py
```diff
@@ -88,70 +152,224 @@ def __init__(
...
+        su = jnp.power(jnp.asarray(2.0, dtype), -level)
+        st = su / 2
+        t = s + st
+        u_minus_s = u - s
+        su = eqxi.error_if(
+            su,
+            jnp.abs(u_minus_s - su) > 1e-17,
+            "VirtualBrownianTree: u-s is not 2^(-tree_level)",
+        )
```
patrick-kidger, line 283: I don't think we actually need this test at runtime? IIRC this was just for checking things whilst developing.

diffrax/brownian/tree.py
```diff
@@ -161,67 +379,105 @@ def _cond_fun(_state):
...
-            _u = jnp.where(_cond, _state.u, _state.t)
-            _w_s = jnp.where(_cond, _state.w_t, _state.w_s)
-            _w_u = jnp.where(_cond, _state.w_u, _state.w_t)
+            _s, _t, _u = _state.stu
+            _cond = r > _t
+            _s = jnp.where(_cond, _t, _s)
+            _u = jnp.where(_cond, _u, _t)
+
+            _w = split_interval(_cond, _state.w_stu, _state.w_st_tu)
+            if self.levy_area in ["space-time", "space-time-time"]:
```
patrick-kidger, line 395: Nit: let's remove `space-time-time` here.

diffrax/brownian/tree.py
```diff
@@ -88,70 +152,224 @@ def __init__(
...
+            - `w_t`: value of BM at t
+            - `w_st_tu`: $(W_{s,t}, W_{t,u})$
+            - `bhh`: (optional) $(\bar{H}_s, \bar{H}_t, \bar{H}_u)$
+            - `bhh_st_tu`: (optional) $(\bar{H}_{s,t}, \bar{H}_{t,u})$
+            - `bkk_t`: (optional) $(\bar{K}_s, \bar{K}_t, \bar{K}_u)$
+            - `bkk_st_tu`: (optional) $(\bar{K}_{s,t}, \bar{K}_{t,u})$
+
+        """
+
+        su = jnp.power(jnp.asarray(2.0, dtype), -level)
```
patrick-kidger, line 275: I think this can just be `2**-level`.

diffrax/brownian/tree.py
```diff
@@ -88,70 +152,224 @@ def __init__(
...
+
+        else:
+            key, init_key_w, midpoint_key = jrandom.split(key, 3)
+            w_1 = jrandom.normal(init_key_w, shape, dtype)
+            bhh = None
+            bkk = None
+
+        w = (w_0, w_1, w_1)
+
+        half, w_half, w_inc, bhh, bhh_inc, bkk, bkk_inc = self._brownian_arch(
```
patrick-kidger, line 360: Heads-up that it's good practice to try and avoid a separate call outside of a loop, if possible: this is because JAX will compile each call to `_brownian_arch` separately. So instead picking differential initial conditions and letting the loop run for one more iteration is preferred.

That said, in this case `_brownian_arch` is pretty small, so if that's awkward to do don't worry about that here, and just treat it as a learning point.

test/test_brownian.py
```diff
@@ -155,7 +155,7 @@ def test_conditional_statistics():
     out = sorted(out, key=lambda x: x[0])
 
     # Test their conditional statistics
-    for i in range(1, 98):
+    for i in range(1, len(ts) - 2):
```
patrick-kidger, line 158: Why does this file need changing?

test/test_stla.py
```diff
@@ -0,0 +1,231 @@
+import math
```
patrick-kidger, line 1: Nit: can we call this file `test_levy_area.py`?

patrick-kidger (commented):
Okay! I think this is basically done. Final round of nits and then let's merge this.
Great work getting this done, the result looks excellent.

diffrax/brownian/tree.py
andyElking: Oh just explicitly saying that in this branch of computation we have `levy_area = "space-time"` I now made the comment more clear.

diffrax/brownian/tree.py
andyElking: I don't really understand why this was an issue before, and is not now. Even before this, the distance between s and u would have eventually become smaller than `tol`, no? Anyway, yes I made the change :)

diffrax/brownian/tree.py
andyElking: I see, will do.

diffrax/brownian/tree.py
andyElking: Fair.

test/test_brownian.py
andyElking: The 98 was a bug, given that you had `assert len(ts) > 80` earlier (meaning you were expecting that the filtering process might throw some samples out if they were too close). Furthermore, you weren't adding the first t into the list. Not sure if that was intentional? The key needed changing, because there was this one case in which I'd get a pval of 0.0004... All the others are mostly around 0.5, but that one just happened to be very bad. With this key it passes, however.

test/test_stla.py
andyElking: Yes, I agree.

diffrax/brownian/tree.py
andyElking: Originally I just copied your design choice here. But now that you mention this, I realised it could be improved significantly, so I overhauled it. Now all calls to _brownian_arch are inside the lax.scan. This also makes _State much more concise, and the final interpolation is a bit nicer. I quickly did some tests, and it seems everything works fine, and also somewhat faster.

test/helpers.py
```diff
@@ -80,3 +93,107 @@ def shaped_allclose(x, y, **kwargs):
...
+                shape=shp_dtype,
+                tol=tol,
+                key=key,
+                levy_area=levy_area,
+            )
+        else:
+            return UnsafeBrownianPath(shape=shp_dtype, key=key, levy_area=levy_area)
+
+
+def batch_sde_solve(
```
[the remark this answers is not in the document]
andyElking: I removed this for now, but I should mention that I need these functions to test the order of SRKs. And for that, it is much better to do it properly. Not sure if the way I did it is the way to go, but I am sure something more principled than a for-loop will be needed then.

test/test_brownian.py
```diff
@@ -7,8 +7,11 @@
 import jax.tree_util as jtu
 import pytest
 import scipy.stats as stats
+from jax import config
 
 
+config.update("jax_enable_x64", True)
```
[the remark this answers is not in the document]
andyElking: Yes, just forgot to remove it, wasn't intending to include it really.

test/helpers.py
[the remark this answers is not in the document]
andyElking: Using a for-loop makes this painfully slow.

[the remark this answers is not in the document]
andyElking: I tried doing it with a for-loop with 5 iterations, and it was painfully slow, so I decided to keep it as is. Besides, these functions will be needed for checking the order of SRKs later, so if not now, they would be added in the next PR.

andyElking:
> Nice work fixing the SDE test. I think we can simplify it a little, but otherwise this looks good.
> 
> For next steps: so that we can merge this in, can you rebase on top of the current `dev` branch? This makes a few changes that mostly shouldn't affect you. (The changes are mostly twiddly things like renaming files + better type annotations + better formatting.)

As much as I am in favour of strong typing, this rebase was extremely fiddly. Yes, worth it, but I didn't expect so much resistance :)

patrick-kidger:
Aaaaaand merged! This is now on the dev branch and will be part of the upcoming v0.5.0 release.

I know how fiddly a task this was. Thank you for all your hard work getting this together. :D

andyElking:
Excellent!! Thank you so much for guiding me through this so well, and for all your time and effort!

PR #342, opened 2023-12-20
Added spline test for VBT, fixed a bug in levy_tree_transpose, and added references to VBT

andyElking:
@patrick-kidger This includes the updates you suggested. For the timeing tests, however, I made a branch identical to this one, but with an additional Jupyter notebook:
[Timing tests](https://github.com/andyElking/diffrax_STLA/blob/vbt_timing/VBT_timing_comparison.ipynb)

 I'm not really sure where you wanted me to put the timeing code, so I put it in a separate branch, but I can also add it here.

### DIFF
--- a/benchmarks/brownian_tree_times.py
+++ b/benchmarks/brownian_tree_times.py
@@ -1,0 +1,191 @@
+import timeit
+from typing import cast, Optional, Union
+
+import equinox as eqx
+import equinox.internal as eqxi
+import jax
+import jax.lax as lax
+import jax.numpy as jnp
+import jax.random as jr
+import jax.tree_util as jtu
+from diffrax import VirtualBrownianTree
+from diffrax._brownian.base import AbstractBrownianPath
+from diffrax._custom_types import RealScalarLike
+from diffrax._misc import default_floating_dtype, is_tuple_of_ints, split_by_tree
+from jaxtyping import Array, Float, PRNGKeyArray, PyTree
+
+
+key = jr.PRNGKey(0)
+t0, t1 = 0.3, 2.3
+ts = jnp.linspace(t0, t1, 100000)
+
+
+def vec_eval(_tree, _ts):
+    return jax.vmap(lambda _t: _tree.evaluate(_t, use_levy=True))(_ts)
+
+
+class _State(eqx.Module):
+    s: RealScalarLike
+    t: RealScalarLike
+    u: RealScalarLike
+    w_s: Float[Array, " *shape"]
+    w_t: Float[Array, " *shape"]
+    w_u: Float[Array, " *shape"]
+    key: PRNGKeyArray
+
+
+class OldVBT(AbstractBrownianPath):
+    t0: RealScalarLike
+    t1: RealScalarLike
+    tol: RealScalarLike
+    shape: PyTree[jax.ShapeDtypeStruct] = eqx.field(static=True)
+    key: PyTree[PRNGKeyArray]
+
+    def __init__(
+        self,
+        t0: RealScalarLike,
+        t1: RealScalarLike,
+        tol: RealScalarLike,
+        shape: Union[tuple[int, ...], PyTree[jax.ShapeDtypeStruct]],
+        key: PRNGKeyArray,
+        levy_area: str = "",
+    ):
+        del levy_area
+        self.t0 = t0
+        self.t1 = t1
+        self.tol = tol
+        self.shape = (
+            jax.ShapeDtypeStruct(shape, default_floating_dtype())
+            if is_tuple_of_ints(shape)
+            else shape
+        )
+        if any(
+            not jnp.issubdtype(x.dtype, jnp.inexact)
+            for x in jtu.tree_leaves(self.shape)
+        ):
+            raise ValueError(
+                "VirtualBrownianTree dtypes all have to be floating-point."
+            )
+        self.key = split_by_tree(key, self.shape)
+
+    @eqx.filter_jit
+    def evaluate(
+        self,
+        t0: RealScalarLike,
+        t1: Optional[RealScalarLike] = None,
+        left: bool = True,
+        use_levy: bool = False,
+    ) -> PyTree[Array]:
+        del left, use_levy
+        t0 = eqxi.nondifferentiable(t0, name="t0")
+        if t1 is None:
+            return self._evaluate(t0)
+        else:
+            t1 = cast(RealScalarLike, eqxi.nondifferentiable(t1, name="t1"))
+            return jtu.tree_map(
+                lambda x, y: x - y,
+                self._evaluate(t1),
+                self._evaluate(t0),
+            )
+
+    def _evaluate(self, τ: RealScalarLike) -> PyTree[Array]:
+        map_func = lambda key, struct: self._evaluate_leaf(key, τ, struct)
+        return jtu.tree_map(map_func, self.key, self.shape)
+
+    def _brownian_bridge(self, s, t, u, w_s, w_u, key, shape, dtype):
+        mean = w_s + (w_u - w_s) * ((t - s) / (u - s))
+        var = (u - t) * (t - s) / (u - s)
+        std = jnp.sqrt(var)
+        return mean + std * jr.normal(key, shape, dtype)
+
+    def _evaluate_leaf(
+        self,
+        key,
+        τ: RealScalarLike,
+        struct: jax.ShapeDtypeStruct,
+    ) -> Array:
+        shape, dtype = struct.shape, struct.dtype
+
+        cond = self.t0 < self.t1
+        t0 = jnp.where(cond, self.t0, self.t1).astype(dtype)
+        t1 = jnp.where(cond, self.t1, self.t0).astype(dtype)
+
+        t0 = eqxi.error_if(
+            t0,
+            τ < t0,
+            "Cannot evaluate VirtualBrownianTree outside of its range [t0, t1].",
+        )
+        t1 = eqxi.error_if(
+            t1,
+            τ > t1,
+            "Cannot evaluate VirtualBrownianTree outside of its range [t0, t1].",
+        )
+        τ = jnp.clip(τ, t0, t1).astype(dtype)
+
+        key, init_key = jr.split(key, 2)
+        thalf = t0 + 0.5 * (t1 - t0)
+        w_t1 = jr.normal(init_key, shape, dtype) * jnp.sqrt(t1 - t0)
+        w_thalf = self._brownian_bridge(t0, thalf, t1, 0, w_t1, key, shape, dtype)
+        init_state = _State(
+            s=t0,
+            t=thalf,
+            u=t1,
+            w_s=jnp.zeros_like(w_t1),
+            w_t=w_thalf,
+            w_u=w_t1,
+            key=key,
+        )
+
+        def _cond_fun(_state):
+            return (_state.u - _state.s) > self.tol
+
+        def _body_fun(_state):
+            _key1, _key2 = jr.split(_state.key, 2)
+            _cond = τ > _state.t
+            _s = jnp.where(_cond, _state.t, _state.s)
+            _u = jnp.where(_cond, _state.u, _state.t)
+            _w_s = jnp.where(_cond, _state.w_t, _state.w_s)
+            _w_u = jnp.where(_cond, _state.w_u, _state.w_t)
+            _key = jnp.where(_cond, _key1, _key2)
+            _t = _s + 0.5 * (_u - _s)
+            _w_t = self._brownian_bridge(_s, _t, _u, _w_s, _w_u, _key, shape, dtype)
+            return _State(s=_s, t=_t, u=_u, w_s=_w_s, w_t=_w_t, w_u=_w_u, key=_key)
+
+        final_state = lax.while_loop(_cond_fun, _body_fun, init_state)
+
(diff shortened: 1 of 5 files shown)


diffrax/_brownian/tree.py
```diff
@@ -378,6 +412,21 @@ def _brownian_arch(
...
+          school = {University of Oxford},
+          title = {Numerical approximations for stochastic differential equations},
+          author = {Foster, James M.},
+          year = {2020}
+        }
+
+        In particular see Theorem 6.1.6.
+        ```
+
+
```
patrick-kidger, line 429: This should be formatted as:
```
??? cite "References"

    ```bibtex
    ...
    ```
```
i.e. with a new line, with an indent, and with only a single blank line between the reference section and the arguments section.

diffrax/_custom_types.py
```diff
@@ -1,9 +1,9 @@
 import typing
-from typing import Any, Optional, TYPE_CHECKING, Union
+from typing import Any, Literal, Optional, TYPE_CHECKING, TypeAlias, Union
```
patrick-kidger, line 2: `TypeAlias` should be imported from `typing_extensions` for now, as we don't yet require Python 3.10.

diffrax/_custom_types.py
```diff
@@ -52,6 +52,8 @@ class Real(AbstractDtype):
 BufferDenseInfos = dict[str, PyTree[eqxi.MaybeBuffer[Shaped[Array, "times ..."]]]]
 sentinel: Any = eqxi.doc_repr(object(), "sentinel")
 
+_LA: TypeAlias = Literal["", "space-time"]
```
patrick-kidger, line 55: This shouldn't start with an underscore if it's imported into another module. I'd suggest calling it just `LevyArea`.

test/test_brownian.py
```diff
@@ -146,33 +149,35 @@ def _eval(key):
     assert pval > 0.1
 
 
-@pytest.mark.parametrize("levy_area", ["", "space-time"])
-@pytest.mark.parametrize("use_levy", (False, True))
-def test_conditional_statistics(levy_area, use_levy):
+def conditional_statistics(
+    levy_area: _LA, use_levy: bool, tol=2**-6, spacing=2**-6, spline: _Spline = "sqrt"
```
patrick-kidger, line 153: Let's remove the default arguments here as I don't think they're ever used.

This is a general principle -- prefer not to use default arguments where possible, in particular in internal APIs, as they're a common source of surprising behaviour.

test/test_brownian.py
```diff
@@ -146,33 +149,35 @@ def _eval(key):
...
+):
     key = jr.PRNGKey(5678)
     bm_key, sample_key, permute_key = jr.split(key, 3)
-
     # Get >80 randomly selected points; not too close to avoid discretisation error.
     t0 = 0.3
     t1 = 8.7
     ts = jr.uniform(sample_key, shape=(100,), minval=t0, maxval=t1)
+    # ts = jnp.array([1.0, 3.0, 6.0, 7.0])
```
patrick-kidger, line 161: ?

test/test_brownian.py
```diff
@@ -247,7 +256,54 @@ def test_conditional_statistics(levy_area, use_levy):
...
+@pytest.mark.parametrize("levy_area", ["", "space-time"])
+@pytest.mark.parametrize("use_levy", (False, True))
+def test_conditional_statistics(levy_area: _LA, use_levy):
+    pvals_w1, pvals_w2, pvals_h = conditional_statistics(
+        levy_area, use_levy, tol=2**-8, spacing=2**-10, spline="sqrt"
+    )
+    assert jnp.all(pvals_w1 > 0.1 / pvals_w1.shape[0])
+    if levy_area == "space-time" and use_levy:
+        assert jnp.all(pvals_w2 > 0.1 / pvals_w2.shape[0])
+        assert jnp.all(pvals_h > 0.1 / pvals_h.shape[0])
```
patrick-kidger, line 272:
```python
else:
    assert len(pvals_w2) == 0
    assert len(pvals_h) == 0
```
?

test/test_brownian.py
```diff
@@ -146,33 +149,35 @@ def _eval(key):
...
+    ts.append(prev_ti)
     for ti in sorted_ts[1:]:
-        if ti < prev_ti + 2**-10:
+        if ti < prev_ti + spacing:
             continue
         prev_ti = ti
         ts.append(ti)
     ts = jnp.stack(ts)
-    assert len(ts) > 80
+    assert len(ts) > min(0.2 * (8.0 / spacing), 75)  # for spacing = 2**-5, this is 51
```
patrick-kidger, line 172: maybe just provide a lower bound as an explicit argument to this function, rather than using a heuristic like this?

test/test_brownian.py
```diff
@@ -247,7 +256,54 @@ def test_conditional_statistics(levy_area, use_levy):
...
+    if levy_area == "":
+        splines = ("quad", "sqrt", "zero")
+    elif levy_area == "space-time":
+        splines = ("sqrt", "zero")
+    else:
+        assert False
+
+    for spline in splines:
+        pvals_w1, pvals_w2, pvals_h = conditional_statistics(
+            levy_area, use_levy=True, tol=2**-5, spacing=2**-6, spline=spline
```
patrick-kidger, line 286: Can we parameterise this test by `use_levy=False/True`?

test/test_brownian.py
```diff
@@ -247,7 +256,54 @@ def test_conditional_statistics(levy_area, use_levy):
...
+            # For the correct spline, make sure that all p-values are above
+            # 0.1 (subject to multiple-testing correction) and the average
+            # p-value is above 0.3.
+            def pred(pvals):
+                return jnp.min(pvals) > 0.1 / pvals.shape[0] and jnp.mean(pvals) > 0.3
+
+        else:
+            # make sure that for incorrect splines at least one p-value is
+            # below 0.01 (subject to multiple-testing correction) and the
+            # average p-value is below 0.2.
```
patrick-kidger, line 299: I think the average p-value should probably be a lot smaller than this?

patrick-kidger:
For the timing code, I'd suggest adding it as a standalone `.py` file inside `benchmarks/`.

diffrax/_brownian/tree.py
andyElking: Oops, not sure what happened with the indent there. Will fix

diffrax/_custom_types.py
andyElking: Aha, I see. Didn't know it was new in 3.10.

diffrax/_custom_types.py
andyElking: Good point, yes

test/test_brownian.py
andyElking: I know, thanks for the reminder. I guess old habits die hard.

test/test_brownian.py
andyElking: Yeah, might as well.

test/test_brownian.py
andyElking: This is a bit of a subtle issue. So normally yes, it is a lot smaller, exept in one case. When `levy_area="spacetime"` then `pvals_w1` actually are quite good even with `spline="zero"`, due to the randomness the w_r receives from h_su.

What I'm saying is, that the variance that `pval_w1` sees is the variance of the Brownian parabola, but without accounting for the conditioning on H, i.e.
$$
\E[ ( \E[W_{s,r} | W_{s,u}, H_{s,u} ] )^2 - ( \E[W_{s,r} | W_{s,u} ] )^2  ]
$$
Which is actually a very good approximation of the actual variance. In `pval_w2` the influence of H is accounted for in the mean, which is subtracted, and hence the variance can only come from x1 (which is zero when `spline="zero"`). I hope this makes sense.

So what I will do, is to just keep this more permissive upper bound to `pvals_w1` when `levy_area="spacetime"`, and add a stricter upper bound to all the other cases.

andyElking:
> For the timing code, I'd suggest adding it as a standalone `.py` file inside `benchmarks/`.

Honestly, I wasn't even aware the `benchmarks/` folder even existed until now. Thanks, will do.

test/test_brownian.py
andyElking: I think it might be a bit pointless, but sure, it's not that expensive.

test/test_brownian.py
andyElking: Good idea.

benchmarks/brownian_tree_times.py
```diff
@@ -0,0 +1,191 @@
...
+        A = jnp.array([[2, -4, 2], [-3, 4, -1], [1, 0, 0]])
+        coeffs = jnp.tensordot(A, jnp.stack([w_s, w_t, w_u]), axes=1)
+        return jnp.polyval(coeffs, rescaled_τ)
+
+
+def time_tree(tree_cls, tol, levy_area):
+    tree = tree_cls(t0=t0, t1=t1, tol=tol, shape=(100,), key=key, levy_area=levy_area)
+
+    def f():
+        return jax.block_until_ready(vec_eval(tree, ts))
```
patrick-kidger, line 172: I think there should be a `jax.jit` here as `vec_eval` involves a `jax.vmap`, which is a JAX operation happening outside of vmap.

Also, for timing benchmarks like this it is better to do many repeats and take a minimum. (Noise can only increase things from the best possible time, so you want a minimum, not a mean.)

Thus I think you want something like:
```python
@jax.jit
def run(ts):
    return jax.vmap(lambda _t: tree.evaluate(_t, use_levy=True))(_ts)

return min(timeit.repeat(lambda: jax.block_until_ready(run(ts)), number=1, repeat=100))
```

test/test_brownian.py
```diff
@@ -247,7 +256,74 @@ def test_conditional_statistics(levy_area, use_levy):
...
+        assert len(pvals_w2) == 0
+        assert len(pvals_h) == 0
+
+
+@pytest.mark.parametrize("levy_area", ["", "space-time"])
+@pytest.mark.parametrize("use_levy", (False, True))
+@pytest.mark.parametrize("spline", ("quad", "sqrt", "zero"))
+def test_spline(levy_area: LevyArea, use_levy, spline):
+    if levy_area == "space-time" and spline == "quad":
+        pytest.skip("Quad spline is not implemented for space-time Levy area")
```
patrick-kidger, line 283: FYI, a skipped test is usually used to refer to something that is currently broken, but for which we just don't to worry about right now. In particular it still shows up in the output test log each time.
This means that skips should usually only be used in the short-term, not in the long-term.

Anyway, here you probably want this:
```python
def _levy_area_spline():
    for levy_area in ("", "space-time"):
        for spline in ("quad", "sqrt", "zero"):
            if levy_area == "space-time" and spline == "quad":
                continue
            yield levy_area, spline

@pytest.mark.parametrize("levy_area,spline", _levy_area_spline())
```
which won't even generate the test to be skipped.

patrick-kidger (commented):
Minor comments on the new benchmark/test, but otherwise LGTM. :)

andyElking:
> Minor comments (really teaching points) on the new benchmark/test, but otherwise LGTM. :)

Thanks! These all make sense. I won't be able to make these edits today or tomorrow, however, so is it okay if you fix these in your new_pr_branch and merge if you're already at it? Otherwise I can do it on Monday 😊.

patrick-kidger:
Haha, no worries! I'll let you make the changes after Christmas.

Merry Christmas!

andyElking:
Merry Christmas to you too!

benchmarks/brownian_tree_times.py
```diff
@@ -0,0 +1,191 @@
...
+
+import equinox as eqx
+import equinox.internal as eqxi
+import jax
+import jax.lax as lax
+import jax.numpy as jnp
+import jax.random as jr
+import jax.tree_util as jtu
+from diffrax import VirtualBrownianTree
+from diffrax._brownian.base import AbstractBrownianPath
```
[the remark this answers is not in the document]
andyElking: I think that was done automatically by the IDE, and I haven't noticed.

benchmarks/brownian_tree_times.py
```diff
@@ -0,0 +1,191 @@
...
+from diffrax._custom_types import RealScalarLike
+from diffrax._misc import default_floating_dtype, is_tuple_of_ints, split_by_tree
```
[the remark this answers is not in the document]
andyElking: Yes, that makes sense

test/test_brownian.py
andyElking: Thanks, that's good to know.

benchmarks/brownian_tree_times.py
andyElking: Oh I see, so any JAX operation should be wrapped inside a jit?

andyElking:
I managed to do it today after all 😊

benchmarks/brownian_tree_times.py
patrick-kidger: Yup! At least, if you want them to fast.

So for typical neural network training, I often won't bother putting a JIT when `__init__`ialising a model (which only happens once), but will definitely put one around the entirety of the forward pass.

patrick-kidger:
Alright, LGTM! I've squashed these together and made a few tweaks, and you can see the result back in #337. Let's continue the discussion over there :)

PR #346, opened 2023-12-29
Fixed the tolerance in test_conditional_statistics

andyElking:
I accidentally swapped the value for tol and spacing in test_conditional_statistics. Now the test passes.


patrick-kidger:
Aha, marvellous! Thank you for the fix.

Looks like the failing CI is due to some updated infra -- looks like this also slipped through whilst we were waiting for the new Optimistix release. I've just submitted #347, let's see if that fixes it.

Issue #87, opened 2022-03-29
[Feature Request] Add optional progress bar / log


ma-sadeghi:
When integrating large systems, it might take a while before the solution is returned. It'd be nice if there was a `verbose=False` argument that would either generate a progress bar (maybe using [rich](https://github.com/Textualize/rich)) or just old-school print some stats to console after each time step is done.

patrick-kidger:
So don't forget that the stepping happens under JIT. That does limit the sorts of output we can perform like this.

What you could do already is to use `jax.experimental.host_callback.{id_print,id_tap,call}` inside your vector field, and for which (besides the occasional rejected step) the `t` input serves as a measure of progress.

PR #357, opened 2024-01-18
Add progress bar

abocquet:
Hello @patrick-kidger ,
I'm part of the [dynamiqs](https://github.com/dynamiqs/dynamiqs) team with @gautierronan and @pierreguilmin. I saw https://github.com/patrick-kidger/diffrax/issues/87 and since it was in our plans to implement a progress bar in dynamiqs, I thought I could try to implement it directly into diffrax. I'm very interested in your feedback in order to make this PR meet diffrax standards.

I had to put all tqdm instances in a global object because I think I cannot instanciate them inside a jitted function. I'm open to suggestions if you see a better way.


diffrax/_progress_bar.py
```diff
@@ -0,0 +1,186 @@
+import abc
+
+import equinox as eqx
+import jax
+import jax.numpy as jnp
+import numpy as np
+from jax import Array
+from jax.experimental import io_callback
+from tqdm import tqdm
```
patrick-kidger, line 9: As a stylistic convention, we tend to reserve `from` inputs for either type hints or within-package hints. I'd prefer:
```python
import jax.experimental.io_callback as io_callback
import tqdm.tqdm as tqdm
from jaxtyping import Array  # and by convention we import Array from here
```

Also, I note that this will fail if `tqdm` isn't available on the user's machine. Can we lazy-load it when `TqdmProgressBar` is instantiated for the first time?

diffrax/_progress_bar.py
```diff
@@ -0,0 +1,186 @@
...
+
+from diffrax._custom_types import FloatScalarLike, IntScalarLike
```
patrick-kidger, line 11: Stylistic convention here too: `from ._custom types import ...`

diffrax/_progress_bar.py
```diff
@@ -0,0 +1,186 @@
...
+            current_progress: current progress in percent
+            previous_progress: previous progress in percent
+            current_step: current step number
+        """
+
+        def update_aux(progress):
+            # currently
+            # jax.debug.print("{progress:.2f}%", progress=progress)
+            # does not work, so we have to round the progress ourselves
+            jax.debug.print("{}%", np.round(progress, 1), ordered=True)
```
patrick-kidger, line 83: Why does this work? `np.round` should be called on NumPy arrays, not JAX arrays?

diffrax/_progress_bar.py
```diff
@@ -0,0 +1,186 @@
...
+        previous_progress: FloatScalarLike,
+        current_step: IntScalarLike,
+    ) -> FloatScalarLike:
+        """
+        Args:
+            pb_id: progress bar id
+            current_progress: current progress in percent
+            previous_progress: previous progress in percent
+            current_step: current step number
+        """
```
patrick-kidger, line 77: By the way, the docstring convention we use is Markdown:
```md
**Arguments:**

- `pb_id`: ...
- `current_progress`: ...
- `previous_progress`: ...
- `current_step`: ...

**Returns:**

Foobar.
```

diffrax/_progress_bar.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def close(self, pb_id: IntScalarLike):
+        """
+        Args:
+            pb_id: progress bar id
+        """
+        io_callback(progress_bar_manager.close, None, pb_id, ordered=True)
+
+
+class TqdmProgressBarManager:
```
patrick-kidger, line 158: Nit: can you start this with an underscore to indicate that it's private to this module?

diffrax/_progress_bar.py
```diff
@@ -0,0 +1,186 @@
...
+    def close(self, idx):
+        idx = int(idx)
+        bar = self.bars[int(idx)]
+        bar.n = float(100.0)
+        bar.update()
+        bar.close()
+        del self.bars[idx]
+
+
+progress_bar_manager = TqdmProgressBarManager()
```
patrick-kidger, line 186: (this too)

patrick-kidger (commented):
Right, sorry for taking so long to get around to reviewing this; I've been on holiday. :) On with the show!

So right now I can see that we have a `pb_current_progress` on the state, and a `pb_id` that's instantiated for every run. I think we can simplify the design a little here by doing the same thing as we do for the solvers and stepsize controllers and such -- something like this:
```python
class AbstractProgressBar(eqx.Module, Generic[_State]):
    @abc.abstractmethod
    def init(self) -> _State:
        ...

    @abc.abstractmethod
    def step(self, step: IntScalarLike, state: _State) -> _State:
        ...

    @abc.abstractmethod
    def close(self, state: _State):
        ...
```
and then placing both `pb_current_progress` and `pb_id` in this state.
This means that `NoProgressBar` and `TextProgressBar` don't have to return a dummy `pb_id` -- they can just have an empty state containing no information -- and moreover that any future progress bar implementation can use whatever state it likes; it's not restricted to just a single progress counter.

(Also side note, you can see I've renamed `new -> init` and `update -> step` to match the names that the other classes tend to use.)

---

Anyway, other than these tweaks, design-wise I like what you've got here. This'll be a great addition to Diffrax! :)

test/test_progress_bar.py
```diff
@@ -0,0 +1,57 @@
...
+    term = diffrax.ODETerm(lambda t, y, args: -0.2 * y)
+    solver = diffrax.Dopri5()
+    t0 = 0
+    t1 = 5
+    dt0 = 0.01
+    y0 = 1.0
+    saveat = diffrax.SaveAt(steps=True)
+
+    f = io.StringIO()
+    with contextlib.redirect_stdout(f):
```
patrick-kidger, line 39: Oh by the way, pytest has a built-in for this: see [here](https://github.com/patrick-kidger/equinox/blob/34252989c251c315c6eee5895dd64e9645b350ce/tests/test_while_loop.py#L196-L198) for an example

diffrax/_progress_bar.py
abocquet: I agree with your comments, however if I try to replace `from jax.experimental import io_callback` by `import jax.experimental.io_callback as io_callback` I get a `ModuleNotFoundError`:
```
../diffrax/_progress_bar.py:8: in <module>
    import jax.experimental.io_callback as io_callback
E   ModuleNotFoundError: No module named 'jax.experimental.io_callback'
```
For some reason I cannot seem to be able to import functions direclty from `jax.experimental`. Maybe I'm missing something there.

diffrax/_progress_bar.py
abocquet: I beleive there is interoperability between numpy and jax.numpy functions, as long as you are not in a jitted piece of code
```python
import jax.numpy as jnp
import numpy as np

a = jnp.linspace(0, 10, 101)
print(jnp.sum(a)) # 505.0
print(np.sum(a)) # 505.0
```

abocquet:
Thank you for taking the time to review this PR and all your great comments. I really liked the generic state idea. 
I addressed all comments but the one concerning the import style (see answer)

diffrax/_integrate.py
```diff
@@ -54,7 +59,7 @@ class SaveState(eqx.Module):
     save_index: IntScalarLike
 
 
-class State(eqx.Module):
+class State(eqx.Module, Generic[ProgressBarGenericState]):
```
patrick-kidger, line 62: I'd probably remove the `Generic` here, and just annotate `progress_bar_state: PyTree[Array]`. This will
(a) avoid the `ProgressBarGenericState` complexity -- imagine how highly parameterised this class would be if we included all of them! -- and
(b) means that at test-time, when we do runtime type checking, we'll more carefully check that this attribute consists only of arrays.

(Also, note how I'm suggesting calling it `progress_bar_state` -- I think verbosity is better here, for something that isn't used very much.)

diffrax/_progress_bar.py
```diff
@@ -0,0 +1,186 @@
...
+        self, state: ProgressBarGenericState, progress: FloatScalarLike
+    ) -> ProgressBarGenericState:
+        """Updates the progress bar and returns the last printed progress value"""
+
+    @abc.abstractmethod
+    def close(self, state: ProgressBarGenericState):
+        """Closes the progress bar"""
+
+
+NoProgressBarState = NamedTuple("NoProgressBarState", [])
```
patrick-kidger, line 32: Just `None` will work for an empty PyTree here. E.g.
```python
def init(self) -> None:
    return None
```

diffrax/_progress_bar.py
```diff
@@ -0,0 +1,186 @@
...
+    ) -> NoProgressBarState:
+        return state
+
+    def close(self, state: NoProgressBarState):
+        pass
+
+
+_TextProgressBarState = NamedTuple(
+    "_TextProgressBarState", [("progress", FloatScalarLike)]
+)
```
patrick-kidger, line 52: Just as a convention, I try to use Equinox modules instead of NamedTuples for this kind of thing:
```python
class _TextProgressBarState(eqx.Module):
    progress: FloatScalarLike
```

diffrax/_progress_bar.py
```diff
@@ -0,0 +1,186 @@
...
+    def step(
+        self,
+        state: _TextProgressBarState,
+        progress: FloatScalarLike,
+    ) -> _TextProgressBarState:
+        """
+        **Arguments:**
+            - `state`: progress bar state
+            - `progress`: current progress in percent
+        """
```
patrick-kidger, line 77: I think this docstring can be removed as it's already covered in the abstract base class.

diffrax/_progress_bar.py
patrick-kidger: Hmm, but isn't all of this code called inside JIT?

diffrax/_progress_bar.py
```diff
@@ -0,0 +1,186 @@
...
+)
+
+
+class TqdmProgressBar(AbstractProgressBar):
+    refresh_rate: int
+
+    def __init__(self, refresh_rate: int = 20):
+        """
+        **Arguments:**
+            - `refresh_rate`: number of steps between refreshes
```
patrick-kidger, line 113: Nit: this won't format as Markdown correctly; it needs a newline between the "Arguments" and the list, and the list should be deindented. (Take a look at some of the existing code for an example.)

This matters as we'll put this in the documentation, so it has to parse correctly.

diffrax/_progress_bar.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def new(self):
+        import tqdm
+
+        self.idx += 1
+        bar = tqdm.tqdm(total=100)
+        self.bars[self.idx] = bar
+        return jnp.array(self.idx, dtype=jnp.int32)
+
+    def update(self, progress, idx):
```
patrick-kidger, line 172: Nit: can we call these `init` and `step` to match the progress bars themselves?

patrick-kidger (commented):
Okay, I really like this! I've left nitty comments only -- I'll aim to merge this after this round of comments.

diffrax/_progress_bar.py
abocquet: If I understand correclty it works because [internally](https://github.com/numpy/numpy/blob/d35cd07ea997f033b2d89d349734c61f5de54b0d/numpy/core/fromnumeric.py#L3360) `np.round(a)` calls `a.round()`. To remove all doubt and since is the only use of `np` in the file, I switched to `jnp`

diffrax/_integrate.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
abocquet, line 1: This addition was necessary to prevent typing errors on [line 80](https://github.com/patrick-kidger/diffrax/pull/357/files#diff-d4d01cfb3cf14986b1415437667bdfecc07730f408e4686a75836cfea459f874R80):
```
progress_bar_state: Optional[PyTree[Array]]
```

patrick-kidger: I think those typing errors are valid: you have e.g. `_TextProgressBarState(progress=0.0)` rather than `_TextProgressBarState(progress=jnp.array(0.0))`.

All adding this here does is to disable some of those runtime type checks, because it can't parse the string as a type! (For this reason I don't use this future import anywhere, as it means that jaxtyping can't do its job.)

diffrax/_progress_bar.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def close(self, state):
+        pass
+
+
+class _TextProgressBarState(eqx.Module):
+    progress: FloatScalarLike
+
+    def __init__(self, progress):
+        self.progress = progress
```
patrick-kidger, line 41: This is superfluous by the way: Equinox modules are dataclasses so a default `__init__` method will be synthesise automatically from the fields.

diffrax/_progress_bar.py
```diff
@@ -0,0 +1,174 @@
...
+        jax.debug.print("100%", ordered=True)
+
+
+class _TqdmProgressBarState(eqx.Module):
+    progress_bar_id: IntScalarLike
+    step: IntScalarLike
+
+    def __init__(self, progress_bar_id, step):
+        self.progress_bar_id = progress_bar_id
+        self.step = step
```
patrick-kidger, line 90: (likewise here)

diffrax/_progress_bar.py
```diff
@@ -0,0 +1,174 @@
...
+class TqdmProgressBar(AbstractProgressBar):
+    refresh_rate: int
+
+    def __init__(self, refresh_rate: int = 20):
+        """
+        **Arguments:**
+
+        - `refresh_rate`: number of steps between refreshes
+        """
+        self.refresh_rate = refresh_rate
```
patrick-kidger, line 102: Oh actually, and here! (For the docstring I tend to just do `Foo.__init__.__doc__ = "..."` after the class definition, see e.g. https://github.com/patrick-kidger/diffrax/blob/8aafefcd081d51240ac865662a1767ca999d2455/diffrax/_step_size_controller/adaptive.py#L711 .)

diffrax/_integrate.py
abocquet: Ok thank you for these details! I'

patrick-kidger:
Okay, I think this is now ready to merge! I've made some extra tweaks; the notable ones are:

- Renamed "bar"->"meter" to reflect the fact that e.g. the text output isn't a bar.
- Added docs.
- Fixed bug with `t0 != 0` producing the wrong results.
- Fixed vmap'd diffeqsolves crashing progress bars. (Making this work turned out to involve some fairly advanced JAX stuff, like `nonbatchable` and `unvmap_*`, if you're curious.)

If you get the chance then it'd be great to test that this works for your use-case! Other than that, I'm happy to merge this as-is. :)

abocquet:
I tested on my side and it works like a charm. Thank your for taking the time to polish all these important details!

patrick-kidger:
Awesome. I've just merged into the dev branch, so this will appear in the upcoming v0.5.1 release.

Thank you for contributing!

PR #344, opened 2023-12-25
SRK

andyElking:
This PR adds all the shiny new SRK methods, while leaving out the Langevin-only methods (ALIGN and SORT).
A few things to keep in mind:

- I wrote some comments in test/helpers.py, test_integrate.py, and _solvers/srk.py that are there for you to read and will be removed later. They are mostly questions about somewhat shaky parts of my code that I want to hear your opinions on.
- I included langevin.ipynb in the examples folder. Because we are not merging in the LangevinTerm in _terms.py, I just copied the code for that into the notebook. If you want, I can remove this whole notebook until the next PR, but I do think it is very informative as to how the new stochastic solvers work. Furthermore, it gives you a showcase of what is to come later, which might make it clearer why I made some design decisions, as they make adding the Langevin things more seamless later on.
- I added the new solvers to test_integrate.test_sde_strong_order. There I added a new "additive noise" mode in addition to the "commutative" and "non-commutative" noise options. Later on, once we add LangevinTerm, I also have a new set of tests called test_langevin.py prepared, which further test these methods (although they are primarily aimed and ALIGN and SORT, as those don't work with non-Langevin SDEs).
- Apart from SRA1, I currently don't have any of Rossler's other solvers. Adding these will amount to adding just a few tableaus, which can easily be done later down the line (i.e. after you review the existing code). To be fair these aren't really my priority right now, but I will try to add them before we merge this whole PR.
- I will probably be quite busy over the next two weeks, so I might not be able to make many edits at this point. I'm mostly making this PR now to give you ample time to read all the code 😊.

PR #364, opened 2024-01-30
Parametric control types

### LINKED ISSUE #359: Parametric type hinting for control paths
I would like to type hint that my custom term expects a control which has a certain type of path. E.G. only VirtualBrownianTree not UnsafeBrownianPath.

Ideally the code would look something like this:

```python
from diffrax import AbstractTerm, ODETerm, WeaklyDiagonalControlTerm

class CustomTerm(AbstractTerm)
    ode: ODETerm
    cde: WeaklyDiagonalControlTerm[VirtualBrownianPath]
```

I'm not sure if this is already possible, or would require a change to the code. My thought was to do something like the following:

```python
class _ControlTerm[T: Union[AbstractPath, Callable]](AbstractTerm):
    vector_field: ...
    control: T = ...

class WeaklyDiagonalControlTerm[T](_ControlTerm[T]):
    ...
```

I'm not very experienced with the more advanced aspects of type hinting so perhaps there is a better way, or this is already possible? Any thought are much appreciated

tttc3:
Pull request that closes #359.

Core Changes:
-  Change `_ControlTerm` to `AbstractControlTerm[_Control]`;
- Allow type hinting of the form `ControlTerm[AbstractBrownianPath]`. Provided demonstration in `_solvers/milstein.py`. Will all relevant solvers want updating?

Additional Changes:
- Bumped ruff and pyright pre-commit versions;
- Corrected typos in `_solvers/milstein.py`.

Known Issues:
- In `test/test_term.py:133`  the following type error is thrown (not sure why):
```bash
test/test_term.py:133:46 - error: Argument of type "(t0: Unknown, t1: Unknown) -> tuple[Array]" 
cannot be assigned to parameter "control" of type "_Control@ControlTerm" in function "__init__"
    Type "(t0: Unknown, t1: Unknown) -> tuple[Array]" cannot be assigned to type "AbstractPath"
      "function" is incompatible with "AbstractPath" (reportArgumentType)
```
- In `diffrax/_term.py:266` the following type error is being suppressed (as I can't find a nice way to avoid it):
```bash
diffrax/_term.py:266:14 - error: Cannot assign member "control" for type "AbstractControlTerm[_Control@AbstractControlTerm]*"
    Expression of type "AbstractPath | _CallableToPath" cannot be assigned to member "control" of class "AbstractControlTerm[_Control@AbstractControlTerm]"
      Member "__set__" is unknown
      Type "AbstractPath | _CallableToPath" cannot be assigned to type "_Control@AbstractControlTerm" (reportAttributeAccessIssue)
```

### DIFF
--- a/diffrax/_brownian/base.py
+++ b/diffrax/_brownian/base.py
@@ -4,17 +4,17 @@ from typing import Optional, TypeVar, Union
 from equinox.internal import AbstractVar
 from jaxtyping import Array, PyTree
 
-from .._custom_types import AbstractLevyReturn, RealScalarLike
+from .._custom_types import AbstractBrownianReturn, RealScalarLike
 from .._path import AbstractPath
 
 
-_Control = TypeVar("_Control", bound=Union[PyTree[Array], AbstractLevyReturn])
+_Control = TypeVar("_Control", bound=Union[PyTree[Array], AbstractBrownianReturn])
 
 
 class AbstractBrownianPath(AbstractPath[_Control]):
     """Abstract base class for all Brownian paths."""
 
-    levy_area: AbstractVar[type[AbstractLevyReturn]]
+    levy_area: AbstractVar[type[AbstractBrownianReturn]]
 
     @abc.abstractmethod
     def evaluate(
--- a/diffrax/_brownian/base.py
+++ b/diffrax/_brownian/base.py
@@ -1,5 +1,5 @@
 import abc
-from typing import Optional, Union
+from typing import Optional, TypeVar, Union
 
 from equinox.internal import AbstractVar
 from jaxtyping import Array, PyTree
@@ -8,7 +8,10 @@ from .._custom_types import LevyArea, LevyVal, RealScalarLike
 from .._path import AbstractPath
 
 
-class AbstractBrownianPath(AbstractPath[Union[PyTree[Array], LevyVal]]):
+_Control = TypeVar("_Control", bound=Union[PyTree[Array], LevyVal])
+
+
+class AbstractBrownianPath(AbstractPath[_Control]):
     """Abstract base class for all Brownian paths."""
 
     levy_area: AbstractVar[LevyArea]
@@ -20,7 +23,7 @@ class AbstractBrownianPath(AbstractPath[Union[PyTree[Array], LevyVal]]):
         t1: Optional[RealScalarLike] = None,
         left: bool = True,
         use_levy: bool = False,
-    ) -> Union[PyTree[Array], LevyVal]:
+    ) -> _Control:
         r"""Samples a Brownian increment $w(t_1) - w(t_0)$.
 
         Each increment has distribution $\mathcal{N}(0, t_1 - t_0)$.
--- a/diffrax/_brownian/base.py
+++ b/diffrax/_brownian/base.py
@@ -4,17 +4,17 @@ from typing import Optional, TypeVar, Union
 from equinox.internal import AbstractVar
 from jaxtyping import Array, PyTree
 
-from .._custom_types import LevyArea, LevyVal, RealScalarLike
+from .._custom_types import AbstractLevyArea, RealScalarLike
 from .._path import AbstractPath
 
 
-_Control = TypeVar("_Control", bound=Union[PyTree[Array], LevyVal])
+_Control = TypeVar("_Control", bound=Union[PyTree[Array], AbstractLevyArea])
 
 
 class AbstractBrownianPath(AbstractPath[_Control]):
     """Abstract base class for all Brownian paths."""
 
-    levy_area: AbstractVar[LevyArea]
+    levy_area: AbstractVar[type[AbstractLevyArea]]
 
     @abc.abstractmethod
     def evaluate(
--- a/diffrax/_brownian/base.py
+++ b/diffrax/_brownian/base.py
@@ -8,7 +8,7 @@ from .._custom_types import LevyArea, LevyVal, RealScalarLike
 from .._path import AbstractPath
 
 
-class AbstractBrownianPath(AbstractPath):
+class AbstractBrownianPath(AbstractPath[Union[PyTree[Array], LevyVal]]):
     """Abstract base class for all Brownian paths."""
 
     levy_area: AbstractVar[LevyArea]
--- a/diffrax/_brownian/base.py
+++ b/diffrax/_brownian/base.py
@@ -4,17 +4,17 @@ from typing import Optional, TypeVar, Union
 from equinox.internal import AbstractVar
 from jaxtyping import Array, PyTree
 
-from .._custom_types import AbstractLevyArea, RealScalarLike
+from .._custom_types import AbstractLevyReturn, RealScalarLike
 from .._path import AbstractPath
 
 
-_Control = TypeVar("_Control", bound=Union[PyTree[Array], AbstractLevyArea])
+_Control = TypeVar("_Control", bound=Union[PyTree[Array], AbstractLevyReturn])
 
 
 class AbstractBrownianPath(AbstractPath[_Control]):
     """Abstract base class for all Brownian paths."""
 
-    levy_area: AbstractVar[type[AbstractLevyArea]]
+    levy_area: AbstractVar[type[AbstractLevyReturn]]
 
     @abc.abstractmethod
     def evaluate(
--- a/diffrax/_brownian/path.py
+++ b/diffrax/_brownian/path.py
@@ -10,7 +10,12 @@ import jax.tree_util as jtu
 import lineax.internal as lxi
 from jaxtyping import Array, PRNGKeyArray, PyTree
 
-from .._custom_types import levy_tree_transpose, LevyArea, LevyVal, RealScalarLike
+from .._custom_types import (
+    levy_tree_transpose,
+    RealScalarLike,
+    SpaceTimeLevyArea,
+    TimeLevyArea,
+)
 from .._misc import (
     force_bitcast_convert_type,
     is_tuple_of_ints,
@@ -42,14 +47,14 @@ class UnsafeBrownianPath(AbstractBrownianPath):
     """
 
     shape: PyTree[jax.ShapeDtypeStruct] = eqx.field(static=True)
-    levy_area: LevyArea = eqx.field(static=True)
+    levy_area: type[Union[TimeLevyArea, SpaceTimeLevyArea]] = eqx.field(static=True)
     key: PRNGKeyArray
 
     def __init__(
         self,
         shape: Union[tuple[int, ...], PyTree[jax.ShapeDtypeStruct]],
         key: PRNGKeyArray,
-        levy_area: LevyArea = "",
+        levy_area: type[Union[TimeLevyArea, SpaceTimeLevyArea]],
     ):
         self.shape = (
             jax.ShapeDtypeStruct(shape, lxi.default_floating_dtype())
@@ -57,10 +62,6 @@ class UnsafeBrownianPath(AbstractBrownianPath):
             else shape
         )
         self.key = key
-        if levy_area not in ["", "space-time"]:
-            raise ValueError(
-                f"levy_area must be one of '', 'space-time', but got {levy_area}."
-            )
         self.levy_area = levy_area
 
         if any(
@@ -70,11 +71,11 @@ class UnsafeBrownianPath(AbstractBrownianPath):
             raise ValueError("UnsafeBrownianPath dtypes all have to be floating-point.")
 
     @property
-    def t0(self):
+    def t0(self):  # pyright: ignore[reportIncompatibleVariableOverride]
         return -jnp.inf
 
     @property
-    def t1(self):
+    def t1(self):  # pyright: ignore[reportIncompatibleVariableOverride]
         return jnp.inf
 
     @eqx.filter_jit
@@ -84,7 +85,7 @@ class UnsafeBrownianPath(AbstractBrownianPath):
         t1: Optional[RealScalarLike] = None,
         left: bool = True,
         use_levy: bool = False,
-    ) -> Union[PyTree[Array], LevyVal]:
+    ) -> Union[PyTree[Array], TimeLevyArea, SpaceTimeLevyArea]:
         del left
         if t1 is None:
             dtype = jnp.result_type(t0)
@@ -112,7 +113,7 @@ class UnsafeBrownianPath(AbstractBrownianPath):
         )
         if use_levy:
             out = levy_tree_transpose(self.shape, self.levy_area, out)
-            assert isinstance(out, LevyVal)
+            assert isinstance(out, Union[TimeLevyArea, SpaceTimeLevyArea])
         return out
 
     @staticmethod
@@ -121,25 +122,26 @@ class UnsafeBrownianPath(AbstractBrownianPath):
         t1: RealScalarLike,
         key,
         shape: jax.ShapeDtypeStruct,
-        levy_area: str,
+        levy_area: type[Union[TimeLevyArea, SpaceTimeLevyArea]],
         use_levy: bool,
     ):
         w_std = jnp.sqrt(t1 - t0).astype(shape.dtype)
+        w = jr.normal(key, shape.shape, shape.dtype) * w_std
+        dt = t1 - t0
 
-        if levy_area == "space-time":
+        if issubclass(levy_area, SpaceTimeLevyArea):
             key, key_hh = jr.split(key, 2)
             hh_std = w_std / math.sqrt(12)
             hh = jr.normal(key_hh, shape.shape, shape.dtype) * hh_std
-        elif levy_area == "":
-            hh = None
+            levy_val = levy_area(dt, w, hh)
+        elif issubclass(levy_area, TimeLevyArea):
+            levy_val = levy_area(dt, w)
         else:
             assert False
-        w = jr.normal(key, shape.shape, shape.dtype) * w_std
 
         if use_levy:
-            return LevyVal(dt=t1 - t0, W=w, H=hh, bar_H=None, K=None, bar_K=None)
-        else:
-            return w
+            return levy_val
+        return w
 
 
 UnsafeBrownianPath.__init__.__doc__ = """
--- a/diffrax/_brownian/path.py
+++ b/diffrax/_brownian/path.py
@@ -11,10 +11,10 @@ import lineax.internal as lxi
 from jaxtyping import Array, PRNGKeyArray, PyTree
 
 from .._custom_types import (
+    BrownianIncrement,
     levy_tree_transpose,
     RealScalarLike,
     SpaceTimeLevyArea,
-    TimeLevyArea,
 )
 from .._misc import (
     force_bitcast_convert_type,
@@ -47,14 +47,16 @@ class UnsafeBrownianPath(AbstractBrownianPath):
     """
 
     shape: PyTree[jax.ShapeDtypeStruct] = eqx.field(static=True)
-    levy_area: type[Union[TimeLevyArea, SpaceTimeLevyArea]] = eqx.field(static=True)
+    levy_area: type[Union[BrownianIncrement, SpaceTimeLevyArea]] = eqx.field(
+        static=True
+    )
     key: PRNGKeyArray
 
     def __init__(
         self,
         shape: Union[tuple[int, ...], PyTree[jax.ShapeDtypeStruct]],
         key: PRNGKeyArray,
-        levy_area: type[Union[TimeLevyArea, SpaceTimeLevyArea]],
+        levy_area: type[Union[BrownianIncrement, SpaceTimeLevyArea]],
     ):
         self.shape = (
             jax.ShapeDtypeStruct(shape, lxi.default_floating_dtype())
@@ -85,7 +87,7 @@ class UnsafeBrownianPath(AbstractBrownianPath):
         t1: Optional[RealScalarLike] = None,
         left: bool = True,
         use_levy: bool = False,
-    ) -> Union[PyTree[Array], TimeLevyArea, SpaceTimeLevyArea]:
+    ) -> Union[PyTree[Array], BrownianIncrement, SpaceTimeLevyArea]:
         del left
         if t1 is None:
             dtype = jnp.result_type(t0)
@@ -113,7 +115,7 @@ class UnsafeBrownianPath(AbstractBrownianPath):
         )
         if use_levy:
             out = levy_tree_transpose(self.shape, self.levy_area, out)
-            assert isinstance(out, (TimeLevyArea, SpaceTimeLevyArea))
+            assert isinstance(out, (BrownianIncrement, SpaceTimeLevyArea))
         return out
 
     @staticmethod
@@ -122,7 +124,7 @@ class UnsafeBrownianPath(AbstractBrownianPath):
         t1: RealScalarLike,
         key,
         shape: jax.ShapeDtypeStruct,
-        levy_area: type[Union[TimeLevyArea, SpaceTimeLevyArea]],
+        levy_area: type[Union[BrownianIncrement, SpaceTimeLevyArea]],
         use_levy: bool,
     ):
         w_std = jnp.sqrt(t1 - t0).astype(shape.dtype)
@@ -134,8 +136,8 @@ class UnsafeBrownianPath(AbstractBrownianPath):
             hh_std = w_std / math.sqrt(12)
             hh = jr.normal(key_hh, shape.shape, shape.dtype) * hh_std
             levy_val = SpaceTimeLevyArea(dt, w, hh)
-        elif levy_area is TimeLevyArea:
-            levy_val = TimeLevyArea(dt, w)
+        elif levy_area is BrownianIncrement:
+            levy_val = BrownianIncrement(dt, w)
         else:
             assert False
 
--- a/diffrax/_brownian/path.py
+++ b/diffrax/_brownian/path.py
@@ -114,7 +114,7 @@ class UnsafeBrownianPath(AbstractBrownianPath):
             self.shape,
         )
         if use_levy:
-            out = levy_tree_transpose(self.shape, self.levy_area, out)
+            out = levy_tree_transpose(self.shape, out)
             assert isinstance(out, (BrownianIncrement, SpaceTimeLevyArea))
         return out
 
--- a/diffrax/_brownian/path.py
+++ b/diffrax/_brownian/path.py
@@ -135,9 +135,9 @@ class UnsafeBrownianPath(AbstractBrownianPath):
             key, key_hh = jr.split(key, 2)
             hh_std = w_std / math.sqrt(12)
             hh = jr.normal(key_hh, shape.shape, shape.dtype) * hh_std
-            levy_val = SpaceTimeLevyArea(dt, w, hh)
+            levy_val = SpaceTimeLevyArea(dt=dt, W=w, H=hh, K=None)
         elif levy_area is BrownianIncrement:
-            levy_val = BrownianIncrement(dt, w)
+            levy_val = BrownianIncrement(dt=dt, W=w)
         else:
             assert False
 
--- a/diffrax/_brownian/path.py
+++ b/diffrax/_brownian/path.py
@@ -62,7 +62,7 @@ class UnsafeBrownianPath(AbstractBrownianPath):
             else shape
         )
         self.key = key
-        self.levy_area = levy_area
+        self.levy_area = levy_area  # pyright: ignore[reportIncompatibleVariableOverride]
 
         if any(
             not jnp.issubdtype(x.dtype, jnp.inexact)
@@ -113,7 +113,7 @@ class UnsafeBrownianPath(AbstractBrownianPath):
         )
         if use_levy:
             out = levy_tree_transpose(self.shape, self.levy_area, out)
-            assert isinstance(out, Union[TimeLevyArea, SpaceTimeLevyArea])
+            assert isinstance(out, (TimeLevyArea, SpaceTimeLevyArea))
         return out
 
     @staticmethod
@@ -129,13 +129,13 @@ class UnsafeBrownianPath(AbstractBrownianPath):
         w = jr.normal(key, shape.shape, shape.dtype) * w_std
         dt = t1 - t0
 
-        if issubclass(levy_area, SpaceTimeLevyArea):
+        if levy_area is SpaceTimeLevyArea:
             key, key_hh = jr.split(key, 2)
             hh_std = w_std / math.sqrt(12)
             hh = jr.normal(key_hh, shape.shape, shape.dtype) * hh_std
-            levy_val = levy_area(dt, w, hh)
-        elif issubclass(levy_area, TimeLevyArea):
-            levy_val = levy_area(dt, w)
+            levy_val = SpaceTimeLevyArea(dt, w, hh)
+        elif levy_area is TimeLevyArea:
+            levy_val = TimeLevyArea(dt, w)
         else:
             assert False
 
--- a/diffrax/_brownian/tree.py
+++ b/diffrax/_brownian/tree.py
@@ -1,5 +1,5 @@
 import math
-from typing import Literal, Optional, Union
+from typing import Literal, Optional, TypeVar, Union
 from typing_extensions import TypeAlias
 
 import equinox as eqx
@@ -16,9 +16,9 @@ from .._custom_types import (
     BoolScalarLike,
     IntScalarLike,
     levy_tree_transpose,
-    LevyArea,
-    LevyVal,
     RealScalarLike,
+    SpaceTimeLevyArea,
+    TimeLevyArea,
 )
 from .._misc import (
     is_tuple_of_ints,
@@ -69,7 +69,11 @@ class _State(eqx.Module):
     bkk_s_u_su: Optional[FloatTriple]  # \bar{K}_s, _u, _{s,u}
 
 
-def _levy_diff(x0: LevyVal, x1: LevyVal) -> LevyVal:
+_CompatibleLevyArea = Union[TimeLevyArea, SpaceTimeLevyArea]
+_LevyArea = TypeVar("_LevyArea", TimeLevyArea, SpaceTimeLevyArea)
+
+
+def _levy_diff(x0: _LevyArea, x1: _LevyArea) -> _LevyArea:
     r"""Computes $(W_{s,u}, H_{s,u})$ from $(W_s, \bar{H}_{s,u})$ and
     $(W_u, \bar{H}_u)$, where $\bar{H}_u = u * H_u$.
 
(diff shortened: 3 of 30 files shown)


patrick-kidger:
I think this looks excellent!

> Allow type hinting of the form ControlTerm[AbstractBrownianPath]. Provided demonstration in _solvers/milstein.py. Will all relevant solvers want updating?

So for this, I'm now thinking that what we really need is a way to indicate the type of the control directly in `AbstractTerm`, not just `AbstractControlTerm`. The fact that we had `_ControlTerm` was only ever just an implementation detail.
Put another way, from a user perspective, then ideally they should only need to know about `AbstractTerm` and its concrete subclasses; I don't want to introduce a whole abstract hierarchy!

It's a bit odd, but would something like this work do you think?
```python
class AbstractTerm(eqx.Module, Generic[_Control]):
    ...  # doesn't actually use _Control

class _AbstractControlTerm(AbstractTerm[_Control]):
    control: _Control
    ...
```
(possibly with a `pyright: ignore` to prevent any complaints about an unused generic parameter?)

I'm not completely sold on the design here, so happy to hear alternate suggestions.

> In test/test_term.py:133 the following type error is thrown (not sure why):

Ah, this is due to a quirk of `__init__` methods and inheritance. This takes a small amount of explanation.

Dataclasses will create a new `__init__` method on subclasses, even if an `__init__` method already exists in the parent class. (I.e. dataclasses do not inherit `__init__` methods.) So in this case pyright thinks that is what is happening, and that `ControlTerm.__init__` is being created just by looking at the fields... including `control: _Control`. (And not `control: Union[AbstractPath, Callable]`.

In practice we've found that this is a bit awkard ergonomically, so Equinox deliberately diverges from standard dataclass behaviour here, by just inheriting `__init__` methods as normal if one is provided. (If you're curious, Equinox has precisely two divergences from dataclasses: the one above, and the fact that you can write `self.foo = bar` inside `__init__` methods, rather than requiring the `object.__setattr__(self, "foo", bar)` required with regular frozen dataclasses.)

> In diffrax/_term.py:266 the following type error is being suppressed (as I can't find a nice way to avoid it):

I've bumped into versions of this one before. It's an annoying case with generics + fields + `__init__` methods that I've not found a nice way to resolve. I think adding a `pyright: ignore` annotation is indeed appropriate here.

tttc3:
Thank you for the information, and explanations. They will save me from many future headaches.

Regarding the design of the type hinting for `AbstractTerm`. An alternative approach would be to type hint the output type of `contr`. For example, with the new `UnsafeBrownianPath`, you could do something like:
```python
class AbstractTerm(eqx.Module, Generic[_Control]):
    def contr(self, t0:RealScalarLike, t1:RealScalarLike) -> _Control:
        ...

# Restrict to control which evaluate to provide a LevyArea.
term: AbstractTerm[LevyArea]
# This would be valid
term = ControlTerm(UnsafeBrownianPath(..., levy_area="space-time"))
# This would be invalid.
term = ControlTerm(UnsafeBrownianPath(...))
```
This approach is more permissive than hinting on the paths themselves. It is also closer to my original problem, where I don't really care what path the user provides, only that the control evaluates to a certain shape.

The downside is that you might need to define new types for distinguishing between different controls.

patrick-kidger:
Ah, that's a nice alternative solution. Indeed it does mean that there's no way to say something like 'this solver only accepts Brownian motions', since our existing Brownian paths just return normal JAX arrays. Another possible difficulty is that I'm not sure how easy this will be for either runtime or static type checkers to pick up on.

However on the plus side, it's something which really does make sense for `AbstractTerm` itself -- and arguably a point of Diffrax is that you *shouldn't* need to distinguish between different types of control, with the our term system deliberately erasing any difference in types there. So I think on balance I like your suggestion better - let's go with that!

I think one change that we would want to introduce is an `AbstractLevyVal`, so that we can distinguish between the different kinds of Levy area that we might want. The goal I have in mind here is that if we're going to do parameterisation by control, then we should be able to replace this (from #344):

https://github.com/patrick-kidger/diffrax/blob/59778bdc285aa6ddbc8068c9344175fb06bb71fe/diffrax/_solver/srk.py#L285-L301

with something more principled.

tttc3:
Implemented so far:
- Generic typing of `AbstractTerm.contr`
- Generic typing of `AbstractPath.evaluate` and `AbstractPath.derivative`.

Remaining:
- Abstraction of `LevyVal`.

Known Issues:
- test/test_term.py:135 from the original comment.
- Unfortunately it doesn't appear that any of mypy, pyright or beartype can spot the type violation in the example `test/test_term.py:67`.  I've tested with `reveal_type` and the Generic term remains as Unknown.

patrick-kidger:
> Unfortunately it doesn't appear that any of mypy, pyright or beartype can spot the type violation in the example test/test_term.py:67.

It's not obvious to me what the error is supposed to be here? `ControlTwo()` has its return type annotated as `LevyVal`, which is consistent with the `term_typed: diffrax.ControlTerm[diffrax.LevyVal]` annotation.

For what it's worth, when it comes to tackling the fact that our Brownian motions switch their return type based on an `__init__` argument, I think I've managed to figure out how to get pyright to check that (by adding some dummy annotations for `__new__`):
```python
from typing import cast, Generic, Literal, overload, TypeVar, TYPE_CHECKING, Union

T = TypeVar("T")

class Foo(Generic[T]):
    def meth(self) -> T:
        ...

class Bar(Foo[T]):
    if TYPE_CHECKING:
        @overload
        def __new__(cls, arg: Literal["use_int"]) -> "Bar[int]":
            ...

        @overload
        def __new__(cls, arg: Literal["use_str"]) -> "Bar[str]":
            ...

        def __new__(cls, arg: Literal["use_int", "use_str"]):
            if arg == "use_int":
                return cast(Bar[int], None)
            elif arg == "use_str":
                return cast(Bar[str], None)
            else:
                assert False

    def __init__(self, arg: Literal["use_int", "use_str"]):
        self.arg = arg

    def meth(self) -> T:
        if self.arg == "use_int":
            return 2  # pyright: ignore
        elif self.arg == "use_str":
            return "hi"  # pyright: ignore
        else:
            assert False

def f(x: Foo[int]):
    pass

f(Bar("use_str"))  # this produces an error with pyright!
```

tttc3:
So I've managed to avoid the error in `test/test_term.py:136` by utilising a property that calls `_callable_to_path` on the `_control` attribute. 

I've also simplified the type checking tests and added a placeholder for beartype runtime type checking, which can be enabled if https://github.com/beartype/beartype/issues/238 is implemented.  Without this feature there will be no way to perform checking of any control typed as "ControlTerm[PyTree[...]]", as static checking evaluates such hints to `Unknown`.
That said, the hints will still be useful as documentation.

I like your trick for the init argument dependant hints!

Is there any reason why `use_levy` and `levy_area` both need to exist If we had types like these:

```python
class BrownianIncrement(eqx.Module):
    dt: ...
    W: ...

class SpaceTimeLevyArea(BrownianIncrement):
    H: ...
    bar_H: ...
    K: ...
    bar_k: ...
```
Maybe not exactly like this as they don't obey concrete is final?

diffrax/_brownian/base.py
```diff
@@ -8,7 +8,7 @@
 from .._path import AbstractPath
 
 
-class AbstractBrownianPath(AbstractPath):
+class AbstractBrownianPath(AbstractPath[Union[PyTree[Array], LevyVal]]):
```
patrick-kidger, line 11: Note that this means that `AbstractBrownianPath` can no longer be subscripted. I think you probably mean something like
```python
_Control = TypeVar("_Control", bound=Union[PyTree[Array], LevyVal])

class AbstractBrownianPath(AbstractPath[_Control]):
```
to preserve the possibility of subscripting its subclasses.

diffrax/_custom_types.py
```diff
@@ -46,6 +46,7 @@
 Y = PyTree[Shaped[ArrayLike, "?*y"], "Y"]
 VF = PyTree[Shaped[ArrayLike, "?*vf"], "VF"]
 Control = PyTree[Shaped[ArrayLike, "?*control"], "C"]
+_Control = TypeVar("_Control", bound=Control)
```
patrick-kidger, line 49: On which note, I'd suggest not putting `_Control` here, but just copying it into every file it's needed. This (a) means that the `bound` can be refined, and (b) I generally try not to import `TypeVar`s, but instead treat them as part of the implementation of whatever it is they're subcripting.

(Also (c) the convention in Diffrax is that having an underscore at the start of a name indicates that it is private to the *file* it is in, not to the overall package.)

diffrax/_solver/milstein.py
```diff
@@ -42,13 +43,15 @@ class StratonovichMilstein(AbstractStratonovichSolver):
         Note that this commutativity condition is not checked.
     """  # noqa: E501
 
-    term_structure: ClassVar = MultiTerm[tuple[ODETerm, AbstractTerm]]
+    term_structure: ClassVar = MultiTerm[
+        tuple[ODETerm, AbstractTerm[AbstractBrownianPath]]
```
patrick-kidger, line 47: I think this needs updating to use the return type, not the type of the control path itself.

(Which in this case I think means we can just keep the original unparameterised
```python
term_structure: ClassVar = MultiTerm[tuple[ODETerm, AbstractTerm]]
```
)

diffrax/_term.py
```diff
@@ -219,45 +219,55 @@ def _mul(v):
...
+    @staticmethod
+    def _callable_to_path(
+        x: Union[
+            AbstractPath[_Control], Callable[[RealScalarLike, RealScalarLike], _Control]
+        ],
+    ) -> AbstractPath[_Control]:
+        if isinstance(x, AbstractPath):
+            return x
+        else:
+            return _CallableToPath(x)
```
patrick-kidger, line 265: To be honest, I'd probably sacrifice accuracy in the name of brevity here, and just do
```python
control: Union[AbstractPath[_Control], Callable[[RealScalarLike, RealScalarLike], _Control]] = eqx.field(converter=_callable_to_path)
```

This is a slightly-too-broad annotation for the field itself, of course, but I think it's more readable overall.

test/test_term.py
```diff
@@ -42,6 +43,27 @@ def derivative(self, t, left=True):
     assert vf_prod.shape == (3,)
     assert tree_allclose(vf_prod, term.prod(vf, dx))
 
+    # `# type: ignore` is used for contrapositive static type checking as per:
+    # https://github.com/microsoft/pyright/discussions/2411#discussioncomment-2028001
+    term_static: diffrax.ControlTerm[Array] = term  # noqa: F841
+    term_static_contra: diffrax.ControlTerm[diffrax.LevyVal] = term  # type: ignore # noqa: F841
```
patrick-kidger, line 49: Can we drop the `noqa` if the variable is named `_`?

patrick-kidger:
Oh for completeness -- regarding the `BrownianIncrement` and `SpaceTimeLevyArea` types, this is how you'd write them in a concrete-means-final way:
```python
class AbstractBrownianReturn(eqx.Module):
    dt: AbstractVar[...]
    W: AbstractVar[...]

class BrownianIncrement(AbstractBrownianReturn):
    dt: ...
    W: ...

class SpaceTimeLevyArea(AbstractBrownianReturn):
    dt: ...
    W: ...
    H: ...
    bar_H: ...
    K: ...
    bar_k: ...
```
if need be you could also introduce an `AbstractLevyArea` that sits between `SpaceTimeLevyArea` and `AbstractBrownianReturn` in the type hierarchy, if you wanted to additionally have another kind of `FooLevyArea` type.

(Or you could just keep just `AbstractBrownianReturn`, and then just use an ad-hoc supertype like `Union[SpaceTimeLevyArea, FooLevyArea]` rather than a named supertype like `AbstractLevyArea` -- this is particularly useful in cases where you don't control the type hierarchy because it's in someone else's library.)

patrick-kidger:
*Also also*, would it be possible to add checking for this type into the existing runtime `term_structure` check here:

https://github.com/patrick-kidger/diffrax/blob/8aafefcd081d51240ac865662a1767ca999d2455/diffrax/_integrate.py#L80

?

Probably by doing a `jax.eval_shape(term.contr, ...)` and checking the return type.

That way we can have `diffeqsolve` enforce the validity of `term_structure` (and we don't depend on beartype to do so).

patrick-kidger:
(let me know when you next want my input on this PR by the way)

tttc3:
I think almost everything should be there now. The primary additions since the last review are:
- Corrections to directly address the review comments;
- The addition of a control check in `_integrate.py`, along with a dedicated test in `test_integrate.py`;
- The addition of `AbstractLevyArea` and the subclasses `TimeLevyArea` and `SpaceTimeLevyArea`. 
- Updated Brownian paths to expect one of the new `AbstractLevyArea` types for the `levy_area`.

`TimeLevyArea` might be better noted as `BrownianIncrement`. However, my reservation is that even if the path evaluates to a `PyTree[Array]`, semantically, we still expect this array to represent a Brownian increment of the path?

diffrax/_brownian/path.py
```diff
@@ -121,25 +122,26 @@ def _evaluate_leaf(
...
-        levy_area: str,
+        levy_area: type[Union[TimeLevyArea, SpaceTimeLevyArea]],
         use_levy: bool,
     ):
         w_std = jnp.sqrt(t1 - t0).astype(shape.dtype)
+        w = jr.normal(key, shape.shape, shape.dtype) * w_std
+        dt = t1 - t0
 
-        if levy_area == "space-time":
+        if issubclass(levy_area, SpaceTimeLevyArea):
```
patrick-kidger, line 132: Nit: for clarity, I think I'd probably go for the stricter `if levy_area is SpaceTimeLevyArea` here. We shouldn't be allowing strict subclasses of `SpaceTimeLevyArea` anyway.

diffrax/_brownian/tree.py
```diff
@@ -222,31 +223,31 @@ def evaluate(
         t1: Optional[RealScalarLike] = None,
         left: bool = True,
         use_levy: bool = False,
-    ) -> Union[PyTree[Array], LevyVal]:
-        def _is_levy_val(obj):
-            return isinstance(obj, LevyVal)
+    ) -> Union[PyTree[Array], Union[TimeLevyArea, SpaceTimeLevyArea]]:
+        def _is_levy_area(obj):
+            return isinstance(obj, Union[TimeLevyArea, SpaceTimeLevyArea])
```
patrick-kidger, line 228: Note that `isinstance(..., Union[...])` are buggy across different Python versions. (If you're curious, I go on [a bit of a rant about this here](https://github.com/patrick-kidger/jaxtyping/issues/73#issuecomment-1493431459).) Better to use tuples instead:
```python
isinstance(obj, (TimeLevyArea, SpaceTimeLevyArea))
```

diffrax/_brownian/tree.py
```diff
@@ -222,31 +223,31 @@ def evaluate(
...
 
         t0 = eqxi.nondifferentiable(t0, name="t0")
         # map the interval [self.t0, self.t1] onto [0,1]
         t0 = linear_rescale(self.t0, t0, self.t1)
         levy_0 = self._evaluate(t0)
+        # Should this really be returning a levy area straight away?
```
patrick-kidger, line 234: What do you mean?

diffrax/_custom_types.py
```diff
@@ -53,34 +52,41 @@
...
 
+class AbstractLevyArea(eqx.Module):
+    dt: eqx.AbstractVar[PyTree]
+    W: eqx.AbstractVar[PyTree]
 
-class LevyVal(eqx.Module):
+
+class TimeLevyArea(AbstractLevyArea):
+    dt: PyTree
+    W: PyTree
```
patrick-kidger, line 63: I think I'd probably call this something like `BrownianIncrement`. (And if you object to that subclassing something called `AbstractLevyArea` then we could change the name to something like `AbstractBrownianReturn`.)

There really isn't any Levy area here so I don't think we should use that in the name.

diffrax/_custom_types.py
```diff
@@ -53,34 +52,41 @@
...
     dt: PyTree
     W: PyTree
-    H: Optional[PyTree]
-    bar_H: Optional[PyTree]
-    K: Optional[PyTree]
-    bar_K: Optional[PyTree]
+    H: PyTree
+    bar_H: Optional[PyTree] = None
+    K: Optional[PyTree] = None
+    bar_K: Optional[PyTree] = None
```
patrick-kidger, line 72: @andyElking can you remind me, when are each of these values set?

Now that we explicitly distinguish at the type level whether or not the space-time Levy area is available, I'm thinking we should be able to remove the `Optional`, or move them to other subclasses of `AbstractLevyArea`?

diffrax/_brownian/tree.py
```diff
@@ -206,14 +211,10 @@ def sqrt_mult(z):
...
-            W=jtu.tree_map(sqrt_mult, x.W),
-            H=jtu.tree_map(sqrt_mult, x.H),
-            bar_H=None,
-            K=jtu.tree_map(sqrt_mult, x.K),
-            bar_K=None,
-        )
+        leaves, treedef = jtu.tree_flatten(x)
+        dt_normalized = [mult(leaves[0])]
+        other_normalized = jtu.tree_map(sqrt_mult, leaves[1:])
+        return jtu.tree_unflatten(treedef, dt_normalized + other_normalized)
```
patrick-kidger, line 217: So I think this is a bit dangerous, quietly assuming that `dt` is going to be the first return value. It's a little more effort, but you should be able to extract it like so:
```python
dt_normalised = mult(x.dt)
x = eqx.tree_at(lambda y: y.dt, x, None)
x = jtu.tree_map(sqrt_mult, x)
return eqx.tree_at(lambda y: y.dt, x, dt_normalised, is_leaf=lambda z: z is None)
```
which replaces the `dt` field with a `None` whilst applying the `sqrt_mult`, and then reinserts the normalised version back.

diffrax/_brownian/tree.py
```diff
@@ -261,27 +262,23 @@ def _evaluate_leaf(
...
-            bhh_0 = jnp.zeros_like(bhh_1)
-            bhh = (bhh_0, bhh_1, bhh_1)
-            bkk = None
-
-        else:
-            state_key, init_key_w = jr.split(key, 2)
-            bhh = None
-            bkk = None
-
+        state_key, init_key_w, init_key_la = jr.split(key, 3)
```
patrick-kidger, line 271: This was originally inside the `if self.levy_area == "space time"` branch so that we wouldn't generate an unused `init_key_la` in the common case that the output doesn't require Levy area. I think we should keep that, for efficiency.

diffrax/_brownian/tree.py
```diff
@@ -474,22 +465,17 @@ def _brownian_arch(
...
             w_t = w_s + w_st
-            w_stu = (w_s, w_t, w_u)
 
             bhh_t = bhh_s + bhh_st + 0.5 * (t * w_s - s * w_t)
             bhh_stu = (bhh_s, bhh_t, bhh_u)
-            bkk_stu = None
-            bkk_st_tu = None
 
         else:
-            assert _state.bhh_s_u_su is None
```
patrick-kidger, line 485: I think we lost this assert statement in the refactor.

I'm a little antsy about the refactor here -- easiser to be sure this is doing the desired thing (e.g. like the above assert statement) if it's just the `if` statement that changes to accomodate hew new syntax.

By the way, note that the original version also used single-assignment, e.g. `bhh_stu` was assigned to precisely once, regardless of which branch was taken. In this version, it's instead assigned to as `None` at the start, and then overriden inside one of the `if` branches. Generally speaking single-assignment tends to be (I believe!) a more reliable programming style, as you don't rely on default values. (This has been a staple of FP languages for a while; more recently single-assignment is something that has been popularised by Rust.)

diffrax/_integrate.py
```diff
@@ -73,23 +73,30 @@ class State(eqx.Module):
...
-                    raise ValueError
-            else:
-                raise ValueError
+                if _term_compatible(term.terms, get_args(_tmp)):
+                    return
         else:
-            if not isinstance(term, term_cls):
-                raise ValueError
+            term_args_expected = get_args(term_cls)
+            control_compatible = True
```
patrick-kidger, line 90: Another example of single-assignment here: better to have an `else: control_compatible = True` branch, I think.

diffrax/_integrate.py
```diff
@@ -73,23 +73,30 @@ class State(eqx.Module):
...
+            if term_args_expected:
```
patrick-kidger, line 91: Here's another safe-programming-style hobby-horse of mine :)
Always explicitly use booleans in `if` statements. IIUC, I think we're (currently) expecting this to be either a `None` or a length-1 tuple of the control type? In that case I would write this as:
```python
if term_args_expected is None:
    control_compatible = True
else:
    (term_arg_expected,) = term_args_expected
    ...
```

diffrax/_integrate.py
```diff
@@ -73,23 +73,30 @@ class State(eqx.Module):
...
+                term_cls = get_origin(term_cls)
+                t0 = getattr(term, "control.t0", -jnp.inf)
+                t1 = getattr(term, "control.t1", jnp.inf)
+                control = jax.eval_shape(term.contr, t0, t1)
```
patrick-kidger, line 95: `jax.eval_shape` will just look at the shape and dtype and ignore the rest. I think Just `jax.eval_shape(term.contr, 0.0, 0.0)` should work.

test/test_term.py
```diff
@@ -42,6 +43,27 @@ def derivative(self, t, left=True):
     assert vf_prod.shape == (3,)
     assert tree_allclose(vf_prod, term.prod(vf, dx))
 
+    # `# type: ignore` is used for contrapositive static type checking as per:
+    # https://github.com/microsoft/pyright/discussions/2411#discussioncomment-2028001
+    _term_static: diffrax.ControlTerm[Array] = term
+    _term_static_contra: diffrax.ControlTerm[diffrax.LevyVal] = term  # type: ignore
```
patrick-kidger, line 49: I think the variables can be just a `_` as they're unused:
```python
_: diffrax.ControlTerm[Array] = term
_: diffrax.ControlTerm[diffrax.LevyVal] = term
```

patrick-kidger (commented):
You can see my main comments are really just about a few points of either programming style (single-assignment, explicit bools in if statements) or PR style (try to minimise diffs so that they're easy for me to review!)

So basically, they're nits and I like how this looks overall :)

One question -- how do you want to tackle this alongside #370? I'm thinking we probably want the VF type to appear in the signature before the control type, e.g. `AbstractTerm[_VF, _Control]`.

diffrax/_brownian/path.py
tttc3: I agree, the only minor issue is that when doing `levy_area is SpaceTimeLevyArea` pyright can't seem to narrow `levy_area` down to a `SpaceTimeLevyArea`. 
```python
if levy_area is SpaceTimeLevyArea:
    reveal_type(levy_area) # type[TimeLevyArea] | type[SpaceTimeLevyArea]
    levy_area(dt, w, h) # PyRight complains that `h` is in the call signature.
    SpaceTimeLevyArea(dt, w, h) # This explicit call to the initialiser is fine.
elif levy_area is TimeLevyArea:
    reveal_type(levy_area) # type[TimeLevyArea] | type[SpaceTimeLevyArea]
    levy_area(dt, w) # PyRight complains that `h` is missing from the call signature.
```
I presume this is a bug in PyRight, but I'm not sure. In any case, I can replace the `levy_area` with explicit calls to the appropriate initialiser?

diffrax/_brownian/tree.py
tttc3: Apologies for the rogue comment, I have no idea what I was meaning either.

diffrax/_brownian/tree.py
tttc3: I was not aware of this. Your post was a good read thank you!

diffrax/_brownian/tree.py
tttc3: Understood. I have reintroduced the assertion and ensured single assignment. If you would like me to revert all but the changes to the if statement then no worries. I couldn't find one a quick Look, but it would be nice if ruff/pyright had an option to flag multi-assignment?

diffrax/_brownian/tree.py
tttc3: I had originally attempted to use `eqx.partition` and `eqx.combine` with `is_leaf=lambda x: isinstance(x, AbstractLevyArea)`, and `filter_spec=lambda x: x.dt` but received TracerBooleanConversionErrors. Would you have expected this to fail? Also, would the tree_at approach actually occur in place, unlike the partition combine case?

diffrax/_custom_types.py
tttc3: How about renaming `AbstractLevyArea` to `AbstractLevyVal` and then having `LevyIncrement`?

tttc3:
Regarding #370, I'm happy to include vector_field typing in this pull, and agree with your suggested format for`AbstractTerm[_VF, _Control]`.

andyElking:
Hey @tttc3 if you want a somewhat clearer overview of what the VBT is doing with Levy Areas, I have a paper in the works here:
https://www.overleaf.com/read/zqvmccrdwzsy#ccc24b

I will try to give this thread a bit more attention once I figure out how `eqxi.scan` works 😅

andyElking:
> `TimeLevyArea` might be better noted as `BrownianIncrement`. However, my reservation is that even if the path evaluates to a `PyTree[Array]`, semantically, we still expect this array to represent a Brownian increment of the path?

In fact the Brownian increment is a space Levy Area if anything. Time Levy area would literally just be t.

diffrax/_brownian/path.py
patrick-kidger: Ah, I'd forgotten about this. This is an intentional choice in pyright. (Although not one I agree with.)
In this case let's stick with `issubclass` then!

diffrax/_custom_types.py
patrick-kidger: I think it's really the "Levy" I'm objecting to here, which is usually taken to refer to some kind of quantity that is second-or-higher-order.

diffrax/_custom_types.py
patrick-kidger: Ah, thanks!
In that case I think an internal type might make the most sense. It'd certainly help me wrap my head around the interface here :D

diffrax/_brownian/tree.py
patrick-kidger: I think what's happening htere is that the `x.dt` in `filter_spec=lambda x: x.dt` is trying to be interpreted as a boolean. The intention of a filter spec is that it returns bool values to specify which half of the partition an element ends up in.
I think that could be made to work as well!

`tree_at` operates out of place. (Basically everything in JAX+Equinox does.)

diffrax/_custom_types.py
tttc3: I see, I was going to say, based on @andyElking's comment https://github.com/patrick-kidger/diffrax/pull/364#issuecomment-1938722145 `SpaceLevyArea` would be my preference, but you're far more informed on these matters then myself so I'm more than happy to do the `BrownianIncrement` if you think that is the best option.

andyElking: > I see, I was going to say, based on @andyElking's comment [#364 (comment)](https://github.com/patrick-kidger/diffrax/pull/364#issuecomment-1938722145) `SpaceLevyArea` would be my preference, but you're far more informed on these matters then myself so I'm more than happy to do the `BrownianIncrement` if you think that is the best option.

I suppose calling this "Levy" might not be the most relevant anyway. It's just different terms in the signature of the path (t, W). The terms space-time and space-time-time Levy areas were introduced by James and the signature by Terry. So frankly we could easily name this after someone else...

But anyway, any control could have a signature attached to it. So maybe instead of doing this specifically for a BM we could just allow a general control to either just return its value (as with `use_levy=False`), or to also be endowed with some terms of its signature. And then we can just have a SignatureVal instead of a LevyVal?

tttc3:
I think, with the exception of the discussion on the LevyArea types, my recent commits should address the review comments. The only remaining unimplemented feature is the vector field type hinting for `AbstractTerm`?

test/test_term.py
tttc3: I've just noticed that if you do something like this pyright will raise a `reportAssignmentType` on line one, as it infers the type of `_` based on the hint on line two.
```python
_: diffrax.ControlTerm[Array] = term
reveal_type(_) # diffrax.ControlTerm[diffrax.LevyVal]
_: diffrax.ControlTerm[diffrax.LevyVal] = term

# Simple fix
_: diffrax.ControlTerm[Array] = term
__: diffrax.ControlTerm[diffrax.LevyVal] = term
```

tttc3:
I've now added draft vector field typing support.

diffrax/_custom_types.py
patrick-kidger: Sometimes we also find ourselves considering functions-of-the-signature-of-the-path. E.g. "full" Levy area itself is a difference of two terms.

Indeed, I think one could think about generalising what we're doing here to handle arbitrary terms in the signature. I don't think we need to worry about that here though: we can always just pursue the same strategy as we already have here, namely to have the control return some kind of special type indicating that.

Anyway, to bring this full-circle to this particular comment: I think let's call this thing `BrownianIncrement`.

diffrax/_brownian/tree.py
patrick-kidger: I think changing just the if statement is probably what I'd go for here. I'm not worried about small amounts of code duplication between the branches of the if statement (if anything I think it's easier to read when you don't have to worry about keeping the in-common code correct between both branches), and I think this will be easier for to review!

diffrax/_custom_types.py
```diff
@@ -53,34 +52,41 @@
...
+def levy_tree_transpose(tree_shape, levy_area: type[AbstractLevyArea], tree: PyTree):
     """Helper that takes a PyTree of LevyVals and transposes
     into a LevyVal of PyTrees.
 
     **Arguments:**
 
     - `tree_shape`: Corresponds to `outer_treedef` in `jax.tree_transpose`.
-    - `levy_area`: can be `""` or `"space-time"`, which indicates
-    which fields of the LevyVal will have values.
+    - `levy_area`: the type of `AbstractLevyReturn`
```
patrick-kidger, line 82: "`AbstractLevyReturn`" isn't used as a name anywhere else.

diffrax/_integrate.py
```diff
@@ -73,23 +73,57 @@ class State(eqx.Module):
...
+    def __new__(cls, term_type: Any):
+        if not issubclass(term_type, AbstractArray):
+            return term_type
+
+        cls.array_type = jax.ShapeDtypeStruct
+        cls.dtypes = term_type.dtypes
+        cls.dims = term_type.dims
+        cls.index_variadic = term_type.index_variadic
+        cls.dim_str = term_type.dim_str
+        return cls
```
patrick-kidger, line 86: Haha, I see what you're doing here! FWIW the attributes on `AbstractArray` aren't public API (maybe we should add some underscores in front) so this is a bit dangerous.

To check the shape and dtype of a struct against a jaxtyping array annotation e.g. `Float[Array, "foo"]`, then I'd probably use `jax.eval_shape` to cast the struct to an array:
```python
def _assert(x):
    assert isinstance(x, Float[Array, "foo"])
jax.eval_shape(_assert, struct)
```

diffrax/_integrate.py
```diff
@@ -73,23 +73,57 @@ class State(eqx.Module):
...
+                if n_term_args == 2 and vector_field_compatible:
+                    control = jax.eval_shape(term.contr, 0.0, 0.0)
+                    compatible_control_type = _ShapeDtypeTypeCoercer(term_args[1])
+                    control_compatible = isinstance(control, compatible_control_type)
+                else:
+                    control_compatible = False
+                args_compatible = vector_field_compatible and control_compatible
+            else:
+                _term_cls = term_cls
+                args_compatible = True
```
patrick-kidger, line 123: This is pretty complicated :)
Can you add some comments describing how the annotations are unpacked and tackled here?

diffrax/_integrate.py
```diff
@@ -73,23 +73,57 @@ class State(eqx.Module):
...
+                if _term_compatible(y, args, term.terms, get_args(_tmp)):
+                    return
         else:
-            if not isinstance(term, term_cls):
-                raise ValueError
+            term_args = get_args(term_cls)
+            n_term_args = len(term_args)
+            if n_term_args >= 1:
+                _term_cls = get_origin(term_cls)
+                vector_field = jax.eval_shape(term.vf, 0.0, y, args)
```
patrick-kidger, line 111: This will need to be `eqx.filter_eval_shape` as `args` might include non-JAX-types.

diffrax/_integrate.py
```diff
@@ -73,23 +73,57 @@ class State(eqx.Module):
...
+                compatible_vf_type = _ShapeDtypeTypeCoercer(term_args[0])
+                vector_field_compatible = isinstance(vector_field, compatible_vf_type)
+                if n_term_args == 2 and vector_field_compatible:
+                    control = jax.eval_shape(term.contr, 0.0, 0.0)
+                    compatible_control_type = _ShapeDtypeTypeCoercer(term_args[1])
+                    control_compatible = isinstance(control, compatible_control_type)
```
patrick-kidger, line 117: If the annotation is a more general type that doesn ot support `isinstance`, e.g. `tuple[int, Float[Array, "foo"]]`, then this will fail.

This calls for a runtime type checker!

As it happens, via jaxtyping we already have a transitive dependency on `typeguard==2.13.3` (this exact version), so we can assume it is present. I think add it to `pyproject.toml` to make things explicit, and then you should be able to do:
```python
def better_isinstance(x, annotation):
    @typeguard.typechecked
    def f(y: annotation):
        pass
    try:
        f(x)
    except TypeError:
        return False
    else:
        return True
```

patrick-kidger (commented):
(See also my responses to the two outstanding comments above.)

diffrax/_custom_types.py
tttc3: Given the decision to use `BrownianIncrement`, do you want to use the `AbstractLevyReturn` or `AbstractBrownianReturn` for this type?

diffrax/_custom_types.py
tttc3: Shall I remove the bar attributes from the `SpaceTimeLevyArea` in this pull then?

diffrax/_custom_types.py
patrick-kidger: Let's go for `AbstractBrownianReturn`!

diffrax/_custom_types.py
andyElking: Yes, probably best, but they need to be kept internally in VBT (for the purposes of levy_diff). If you need, I can make an edit to VBT tomorrow so that it doesn't use LevyVal internally, and you can freely reform it elsewhere.

tttc3: Yeah that would be great thanks.

andyElking: You can check PR #378

patrick-kidger:
Heads-up that I've just (a) merged #378 into `dev`, and (b) switched the target branch on this PR to `dev`. The conflicts now being reported are just the changes from that PR -- I think just rebase on top of `dev` and things should work.

diffrax/_integrate.py
tttc3: Really like your solution, much better than my hack haha!

tttc3:
I've added comments to, and attempted to simplify the logic in, the `term_compatibility` check. I've also minimised the diffs where possible and extended the tests to cover the types for which `better_isinstance` is needed. Providing the new tests are sufficient, I think this is now (hopefully) feature complete.

patrick-kidger:
Okay, take a look at the `parametric-tweaks` branch I've just pushed. It's not 100% ready because (a) it's completely untested -- I gotta dash for now -- and (b) I want to add some tests for the tweaks I've made to the term compatbility. To explain what's going on here:
- I don't think the single-argument return from `get_args` is ever possible; Python demands that all type annotations be filled.
- However, they *can* be filled with `TypeVar`s, so I've added a check for that.
- The `y` should be tree-mapped alongside the terms. This is admittedly an entirely undocumented point that I need to make clearer somewhere at some point. (But take a look at `SemiImplicitEuler` for an example of this.)

patrick-kidger:
Aaaaand opened https://github.com/tttc3/diffrax/pull/1 with these changes.

patrick-kidger:
And merged! Gosh, that ended up being much harder than we all thought it would be, huh?

Thank you for the contribution, and great work putting this all together. I'm really happy to have this in, it's a huge improvement to the codebase. :)

tttc3:
I just noticed that at some point we lost the `reportUnnecessaryTypeIgnoreComment = true` from the pyproject.toml, so the contrapositive static type checking test in `test_terms` aren't actually doing anything.

patrick-kidger:
Ah, agreed throughout. Can you send a PR?

pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Mathematics",
 ]
 urls = {repository = "https://github.com/patrick-kidger/diffrax" }
-dependencies = ["jax>=0.4.23", "jaxtyping>=0.2.24", "typing_extensions>=4.5.0", "equinox>=0.11.2", "lineax>=0.0.4", "optimistix>=0.0.6"]
+dependencies = ["jax>=0.4.23", "jaxtyping>=0.2.24", "typing_extensions>=4.5.0", "typeguard==2.13.3", "equinox>=0.11.2", "lineax>=0.0.4", "optimistix>=0.0.6"]
```
dhirschfeld, line 26: Why is `typeguard` pinned to an exact version, especially one that is so old?

It's generally considered very bad packaging practice to pin to exact versions as it's ~never required and prevents getting even security patches without publishing a new version.

Also, it's ancient and won't be getting any updates - current HEAD is 4.2.1:


image

image

Issue #371, opened 2024-02-13
PID controller tolerances sharp bits


ParticularlyPythonicBS:
Hi,
I recently noticed that the default tolerances suggested for the PID controller can give erroneous results even for very simple systems.
Here is a minimum working example:
Lets start with a simple pendulum with a small angle kick:
```python
class SimplePendulum(eqx.Module):
    length: float
    def __call__(self, t, state, args=None):
        theta, omega = state
        dtheta = omega
        domega = - self.length * jnp.sin(theta)
        return jnp.array([dtheta, domega])
dynamics = SimplePendulum(length=1)
init_angle = 0.1
init_omega = 0
state0 = jnp.array([init_angle, init_omega])
```
Then we can specify an integration function:
```python
def integrator(dynamics, state0, stepsize_controller):
    term = diffrax.ODETerm(dynamics)
    solver = diffrax.Dopri8()
    t0 = 0.0
    dt0 = 0.1
    t1 = 1e4 * dt0
    ts = jnp.arange(t0, t1, dt0)
    stepsize_controller = stepsize_controller
    solution = diffrax.diffeqsolve(
        term,
        solver,
        t0,
        t1,
        dt0,
        state0,
        saveat=diffrax.SaveAt(ts=ts),
        stepsize_controller=stepsize_controller,
        max_steps=2**20,
    )
    return solution
```
Using this function we can compare a constant step size controller to PID with different tolerances:
```python
step_size_controller = diffrax.PIDController(rtol=1e-3, atol=1e-6)
sol_incorrect_pid = integrator(dynamics, state0, step_size_controller)
step_size_controller = diffrax.PIDController(rtol=1e-7, atol=1e-9)
sol_correct_pid = integrator(dynamics, state0, step_size_controller)
step_size_controller = diffrax.ConstantStepSize()
sol_const = integrator(dynamics, state0, step_size_controller)
```
And then we can plot the phase portraits by:
```python
fig, axs = plt.subplots(1, 3, figsize=(15, 5), sharey=True)
axs[0].plot(np.sin(sol_const.ys[:,0]), sol_const.ys[:,1])
axs[0].set_title('Constant step size')
axs[1].plot(np.sin(sol_incorrect_pid.ys[:,0]), sol_incorrect_pid.ys[:,1])
axs[1].set_title('PID (incorrect) rtol=1e-3, atol=1e-6')
[...]

PR #372, opened 2024-02-13
Example for adaptive step size choice

ParticularlyPythonicBS:
PR to complete [https://github.com/patrick-kidger/diffrax/issues/371](https://github.com/patrick-kidger/diffrax/issues/371
)
Added example to PIDController with a simple pendulum example illustrating the effect of tolerance of adaptive step choice controller.


diffrax/_step_size_controller/adaptive.py
```diff
@@ -179,6 +179,52 @@ class PIDController(
         common to refer to solving an equation to specific tolerances, without
         necessarily stating which solver was used.)
 
+        !!! Example
```
patrick-kidger, line 182: This is quite a long example, so let's maybe make it expandable with `???`

diffrax/_step_size_controller/adaptive.py
```diff
@@ -179,6 +179,52 @@ class PIDController(
...
+                    domega = - jnp.sin(theta)
+                    return jnp.array([dtheta, domega])
+            dynamics = SimplePendulum()
+            init_angle = 0.1
+            init_omega = 0
+            state0 = jnp.array([init_angle, init_omega])
+            ```
+            We can integrate this using:
+            ```python
+            def integrator(dynamics, state0, stepsize_controller):
```
patrick-kidger, line 201: I think it'd be good to make this example a little more concise. Perhaps something like this:
```python
import diffrax as dfx

def dynamics(t, y, args):
    dtheta = y["omega"]
    domega = - jnp.sin(y["theta"])
    return dict(theta=dtheta, omega=domega)

y0 = dict(theta=0.1, omega=0)
term = dfx.ODETerm(dynamics)
sol = dfx.diffeqsolve(
    term, solver, t0=0, t1=1000, dt0=0.1, y0,
    saveat=dfx.SaveAts(ts=jnp.linspace(0, 1000, 10000),
    max_steps=2**20,
    stepsize_controller=...
)
```
WDYT?

(WIth a little `...`, as we discuss different choices of that below.)

diffrax/_step_size_controller/adaptive.py
```diff
@@ -179,6 +179,52 @@ class PIDController(
...
+            to compare the effect of different tolerances:
+            ```python
+            PID_controller_incorrect = diffrax.PIDController(rtol=1e-3, atol=1e-6)
+            PID_controller_correct = diffrax.PIDController(rtol=1e-7, atol=1e-9)
+            Constant_controller = diffrax.ConstantStepSize()
+            ```
+            The phase portraits of the pendulum from the different tolerances clearly
+            illustrate the impact of the choice of `rtol` and `atol` on the accuracy of
+            the solution.
+            ![Phase portrait of pendulum](../imgs/pendulum_adaptive_steps.png)
```
patrick-kidger, line 226: I probably wouldn't use "incorrect" in the title here, as it's suggestive of a bug. This is the numerical method doing exactly what's expected, namely a cheap low-accuracy solution!
Maybe something like "less accurate" and "more accurate" instead?

patrick-kidger (commented):
Thanks for the PR! I think this'd be a great addition to the docs.

diffrax/_step_size_controller/adaptive.py
ParticularlyPythonicBS: That sounds great!

diffrax/_step_size_controller/adaptive.py
ParticularlyPythonicBS: That works well, 
yeah the class was a vestige of the much larger system I simplified for this, but this is much better as an example

diffrax/_step_size_controller/adaptive.py
ParticularlyPythonicBS: pendulum_adaptive_steps

Sounds good, here is the corrected plot.

ParticularlyPythonicBS:
Just following up to see if anything else is needed

patrick-kidger:
Can you push your changes from the last round of comments?
After that, let's merge this! :)

patrick-kidger:
Alright, merged -- thank you for the contribution!

PR #377, opened 2024-02-16
Avoid jax.config import

jakevdp:
The `jax.config` submodule is empty and will soon be removed. The `jax.config` object, which is how configuration is done in this file, only requires `import jax`.


patrick-kidger:
Cheers Jake! Merged. :)

PR #378, opened 2024-02-16
got Rid of internally using LevyVal in VBT

andyElking:
Just a quick proposal how to get rid of LevyVal used internally in VBT, meaning that bar_H and bar_K can be removed.
@patrick-kidger @tttc3

### DIFF
--- a/diffrax/_brownian/tree.py
+++ b/diffrax/_brownian/tree.py
@@ -69,12 +69,13 @@ class _State(eqx.Module):
     bkk_s_u_su: Optional[FloatTriple]  # \bar{K}_s, _u, _{s,u}
 
 
-def _levy_diff(x0: LevyVal, x1: LevyVal) -> LevyVal:
+def _levy_diff(_, x0: tuple, x1: tuple) -> LevyVal:
     r"""Computes $(W_{s,u}, H_{s,u})$ from $(W_s, \bar{H}_{s,u})$ and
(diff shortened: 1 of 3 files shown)


diffrax/_brownian/tree.py
```diff
@@ -83,19 +84,39 @@ def _levy_diff(x0: LevyVal, x1: LevyVal) -> LevyVal:
...
-    hh_su = inverse_su * bhh_su
-    return LevyVal(dt=su, W=w_su, H=hh_su, bar_H=None, K=None, bar_K=None)
+    if len(x0) == 2:  # BM only case
+        assert len(x1) == 2
+        dt0, w0 = x0
+        dt1, w1 = x1
+        su = jnp.asarray(dt1 - dt0, dtype=w0.dtype)
+        return LevyVal(dt=su, W=w1 - w0, H=None, K=None)
+
+    elif len(x0) == 4:  # space-time levy area case
```
patrick-kidger, line 94: I don't feel so strongly that I think this needs to be changed now, but if I was to do this again then I would recommend creating an internal-only type that's used across this interface, so that we can use safer `if isinstance(x0, Foo): x0.bar` checks and lookups.

(Maybe let's do that whenever space-time-time gets added here, to help avoid any bugs then.)

patrick-kidger:
LGTM, no changes needed -- merged!

PR #381, opened 2024-02-25
Fixup test_term static type checking.

tttc3:
- set reportUnnecessaryTypeIgnoreComment=true, as required by the static type "tests" in `test_term`;

- remove now redundant, commented out, beatype based tests in `test_term`;

- remove all unnecessary type ignore comments from across the codebase.

### DIFF
--- a/benchmarks/small_neural_ode.py
+++ b/benchmarks/small_neural_ode.py
@@ -20,7 +20,7 @@ from jaxtyping import Array
 class FuncTorch(torch.nn.Module):
     def __init__(self):
         super().__init__()
-        self.func = torch.jit.script(  # pyright: ignore
+        self.func = torch.jit.script(
             torch.nn.Sequential(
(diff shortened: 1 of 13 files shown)


tttc3:
The pre-commit complaints are from bumping the pyright version to `1.1.351`. I've had a quick look in `eqx.Enumeration` to see if this can be handled upstream, but I've drawn a blank; any thoughts?

patrick-kidger:
Looks like pyright getting stricter. I've fixed this upsteam on the Equinox dev branch. I'm aiming to do a new release of that fairly shortly, and then we can re-run things here.

tttc3:
I've removed the pyright bump as this wasn't really related to the purpose of this pull.

patrick-kidger:
LGTM! I finally did the new release of Equinox so I'm happy to have this merged.
Thank you for the contribution :D

PR #385, opened 2024-03-12
Update _misc.py

FrisoC13:
This part of the code raised the following error for me: AttributeError: module 'jax.numpy' has no attribute 'astype'. Did you mean: 'dtype'?. Therefore, this is my proposed solution.

### DIFF
--- a/diffrax/_misc.py
+++ b/diffrax/_misc.py
@@ -188,4 +188,4 @@ def upcast_or_raise(
             )
     elif config_value != "standard":
         assert False, f"Unrecognised `JAX_NUMPY_DTYPE_PROMOTION={config_value}`"
-    return jnp.astype(x, promote_dtype)
+    return jnp.array(x, dtype=promote_dtype)


patrick-kidger:
Hmm, I think you might just be on a really old version of JAX? At least on v0.4.25i then `jnp.astype` exists.

FrisoC13:
I was running it on the latest jax-metal version since I have an apple with M1 chip. That might be the root cause of the error. I switched to jax["cpu"] because with the jax libraries I was using I ran in to many errors. I will train on an Nvidia GPU anyway, so for now this is a sufficient solution. But I have noticed that many jax libraries don't integrate well with jax-metal or apple M1. 

Thank you for your rapid response, I have been enjoying reading your research on Neural ODEs and am using them (and Diffrax) now for my master thesis.

patrick-kidger:
Ah, perhaps that is a metal thing. Indeed the metal plugin is still known to be very incomplete.
On that basis I'm going to close this PR for now!

PR #387, opened 2024-03-15
Events

cholberg:
Updates to how events are handled in diffrax. The main changes are:

- There is now only one event class, `Event`.
- Multiple `cond_fn` are supported. An event is triggered whenever one of them changes sign.
- Supports differentiable event times if there are real-valued `cond_fn` and a root_finder is provided.
- New tests for the new event implementation.

Some things that still might require a little thinking about:

- Update documentation (and maybe examples, too).
- What to do with outdated tests.
- How to handle the case where the control is not smooth, e.g., Brownian motion.


diffrax/_integrate.py
```diff
@@ -84,12 +87,20 @@ class State(eqx.Module):
     dense_infos: Optional[BufferDenseInfos]
     dense_save_index: Optional[IntScalarLike]
     progress_meter_state: PyTree[Array]
+    event_result: Optional[PyTree[Union[BoolScalarLike, RealScalarLike]]] = None
+    event_mask: Optional[PyTree[BoolScalarLike]] = None
+    dense_info_for_event: Optional[DenseInfo] = None
+    tprevprev: Optional[FloatScalarLike] = None
```
patrick-kidger, line 93: I don't think default arguments are needed here.

(This is also just a general good-practice-rule-of-thumb: try to avoid default arguments where possible, as they're a common source of unexpected behaviour. So in practice this usually means putting them on public interfaces.)

diffrax/_integrate.py
```diff
@@ -227,14 +238,45 @@ def _maybe_static(static_x: Optional[ArrayLike], x: ArrayLike) -> ArrayLike:
         return x
 
 
+def _is_cond_fn(x: Any) -> bool:
+    return isinstance(x, Callable)
```
patrick-kidger, line 242: I think you can replace this whole function with just the built-in `callable`.

diffrax/_integrate.py
```diff
@@ -227,14 +238,45 @@ def _maybe_static(static_x: Optional[ArrayLike], x: ArrayLike) -> ArrayLike:
...
+
+
+def _compare_events(
+    old_event_result: Union[BoolScalarLike, RealScalarLike],
+    new_event_result: Union[BoolScalarLike, RealScalarLike],
+) -> BoolScalarLike:
+    return jnp.sign(jnp.array(old_event_result, float)) != jnp.sign(
+        jnp.array(new_event_result, float)
```
patrick-kidger, line 250: Don't use the Python builtin `float` (or `int` etc.) as a dtype. This will actually give different behaviour on different platforms (MacOS vs Windows etc.) for some strange reason.
I think in this case you're probably looking for `dtype=jnp.result_type(old_event_result.dtype, jnp.float32)`?

diffrax/_integrate.py
```diff
@@ -309,6 +351,14 @@ def body_fun_aux(state):
         # everything breaks.) See #143.
         y_error = jtu.tree_map(lambda x: jnp.where(jnp.isnan(x), jnp.inf, x), y_error)
 
+        # Save info for event handling
+        if event is not None:
```
patrick-kidger, line 355: Nit: remove the `not` and flip teh branches.

diffrax/_integrate.py
```diff
@@ -957,6 +1036,21 @@ def _allocate_output(subsaveat: SubSaveAt) -> SaveState:
...
+        tprevprev = tprev
+        _, _, dense_info_for_event, _, _ = solver.step(
+            terms,
+            tprev,
+            tnext,
+            y0,
+            args,
+            solver_state,
+            made_jump,
+        )
```
patrick-kidger, line 1049: Note that this will actually evaluate the function, I think essentially pointlessly (as it will be re-evaluated in the loop).
I think using `jax.eval_shape` or `eqx.filter_eval_shape` will allow you to get the structure/dtype/shape of `dense_info_for_event` without having to actually do any work at runtime. (And without having to compile `solver.step` an additional time during compilation time either, which is a nontrivial concern.)

diffrax/_integrate.py
```diff
@@ -977,8 +1071,36 @@ def _allocate_output(subsaveat: SubSaveAt) -> SaveState:
...
+            )
+
+        event_result = jtu.tree_map(_call_event, event.cond_fn, is_leaf=_is_cond_fn)
+        event_mask = jtu.tree_map(lambda x: False, event.cond_fn, is_leaf=_is_cond_fn)
+        init_state = eqx.tree_at(
+            lambda s: s.event_result, init_state, event_result, is_leaf=_is_none
+        )
+        init_state = eqx.tree_at(
+            lambda s: s.event_mask, init_state, event_mask, is_leaf=_is_none
+        )
```
patrick-kidger, line 1102: (a) I think if you put this before `init_state = ...` then you can just pass in this information during initialisation, without `tree_at`.
(b) I think again you can use `eval_shape` to initialise arrays of the appropriate sort without having to actually evaluate each `cond_fn`.

diffrax/_integrate.py
```diff
@@ -1012,6 +1134,67 @@ def _allocate_output(subsaveat: SubSaveAt) -> SaveState:
         lambda s: s.ts * direction, final_state.save_state, is_leaf=is_save_state
     )
     ys = jtu.tree_map(lambda s: s.ys, final_state.save_state, is_leaf=is_save_state)
+
+    # Do root find for exact event times
+    if event is not None:
+        event_mask = final_state.event_mask
+        event_happened = _event_happened(event_mask)
+        tevent = final_state.tprev
```
patrick-kidger, line 1142: Nit: prefer single-assignment form:
```python
if event.root_finder is None:
    tevent = final_state.tprev
else:
    # all the rest of the branch below
    tevent = ...
```
one of these software best-practices for writing easy-to-read code.

patrick-kidger (commented):
This looks very clean! I've gone through and left some comments.

cholberg:
Thank you for the comments, Patrick. I have gone through and made changes accordingly. The only thing I would disagree with is your comment on not having to evaluate `cond_fn`. Correct me if I am wrong, but would we not need to know the sign of `cond_fn` evaluated at the intial condition. I.e., apriori you cannot know if an event happens when the condition function goes from negative to positive or vice versa. Furthermore, if we want `cond_fn` to also possibly take the state as an argument, `event_result` would need to be initialized after `init_state` is defined.

patrick-kidger:
Just letting you know that I've not forgotten about this! I'm trying to focus on getting #344 in first, and then I'm hoping to return to this. They're both quite large changes so I don't want them to step on each other's toes.

patrick-kidger:
Okay, #344 is in! I'd love to get this in next.

I appeciate that's rather a lot of merge conflicts. If you're able to rebase on to the latest `dev` branch then we can start work on getting this in :)

cholberg:
Perfect, I rebased and squashed all the commits into a single big one. Quite a few tests are failing when I run it locally now, but I just wanted to update it so you could have a look. 

Two thoughts since we last touched base:
1. We should probably narrow down exaclty what arguments should be passed to the functions in `cond_fn`.
2. We might want to make it so that the user has a way of passing arguments to the `optimistix.root_find` call that finds the exact event time.

LuggiStruggi:
Hi! Maybe i am using it wrong but at the moment I can't get the root_finder to do smth. I get the same event times when i use the Newton Method as a root finder or just `root_finder = None`. Also the tests still pass when setting the root finder to None:
 ``` python
 45 def test_continuous_event_time():
 46     term = diffrax.ODETerm(lambda t, y, args: 1.0)
 47     solver = diffrax.Tsit5()
 48     t0 = 0
 49     t1 = jnp.inf
 50     dt0 = 1.0
 51     y0 = -10.0
 52 
 53     def cond_fn(state, y, **kwargs):
 54         assert isinstance(state.y, jax.Array)
 55         return y
 56 
 57     #root_finder = optx.Newton(1e-5, 1e-5, optx.rms_norm)
 58     root_finder = None
 59     event = diffrax.Event(cond_fn, root_finder)
 60     sol = diffrax.diffeqsolve(term, solver, t0, t1, dt0, y0, event=event)
 61     assert jnp.all(jnp.isclose(cast(Array, sol.ts), 10.0, 1e-5))
```
Or in my code the resulting event times do not change if having a root finder or not:
``` python
  7 def bouncing_ball():
  8     g = 9.81
  9     damping = 0.8
 10     max_bounces = 10
 11     vx_0 = 5.0
 12 
 13     def dynamics(t, y, args):
 14         x, y, vx, vy = y
 15         dxdt = vx
 16         dydt = vy
 17         dvxdt = 0
 18         dvydt = -g
 19         return jnp.array([dxdt, dydt, dvxdt, dvydt])
 20 
 21     def cond_fn(state, y, **kwargs):
 22         return y[1] < 0
 23 
 24     y0 = jnp.array([0.0, 10.0, vx_0, 0.0])
 25     t0, t1 = 0, float('inf')
 26 
 27     times = []
 28     states = []
 29 
 30     for _ in range(max_bounces):
 31         root_finder = optx.Newton(1e-5, 1e-5)
 32         #root_finder = None
 33         event = Event(cond_fn, root_finder=root_finder)
 34         solver = Tsit5()
 35 
 36         sol = diffeqsolve(ODETerm(dynamics), solver, t0, t1, 0.01, y0, event=event)
 37 
 38         t0 = sol.ts[-1]
 39         last_y = sol.ys[-1]
 40         y0 = last_y * jnp.array([1, 0, 1, -damping])
 41         times.append(sol.ts)
 42         states.append(y0)
 43 
 44 
 45     return jnp.array(times), jnp.array(states)
```
Thanks for the help!

cholberg:
> Hi! Maybe i am using it wrong but at the moment I can't get the root_finder to do smth. I get the same event times when i use the Newton Method as a root finder or just `root_finder = None`. Also the tests still pass when setting the root finder to None:

Ah, thanks for mentioning this. This is essentially due to the fact that the solution to the ODE is linear and the fact that `dt0` divides `10.0` which is exactly the time at which the solution crosses 0. In other words, it just so happens that the root is exactly at the end point of the last step of the solver which is also the event time that is returned when no root finder is provided. I have tweaked the test so this is no longer the case.

> Or in my code the resulting event times do not change if having a root finder or not:

In your example the cond_fn returns a boolean. In this case the returned event time is exactly the first step of the solver for which `cond_fn` switches sign, i.e., your event time will be `n * dt0` where `n` is the number of steps taken until and including the point at which `y[1] >= 0`. Note that this is the same behaviour as when `root_finder=None`. 

If you want continuous event times, you should specify a real-valued condition function. In your bouncing ball example, this would simply correspond to setting:

```python
def cond_fn(state, y, **kwargs):
	return y[1]
```

LuggiStruggi:
Ah i see! Was just a bit confusing that both boolean and comparison with 0 is possible. I tried a real valued cond_fn initially but then my it was every time directly triggered at t=0 bc i reset the state to 0 at every jump. I guess setting it to a small value should work.

LuggiStruggi:
Regarding my confusion: Maybe this is a bad idea but maybe it would make sense to have a warning if both a boolean `cond_fn` and a rooot finder are passed to the event since this should never make sense then?

LuggiStruggi:
If I pass multiple `cond_fn`, what would you suggest to determine which of them caused the Event?

cholberg:
> Regarding my confusion: Maybe this is a bad idea but maybe it would make sense to have a warning if both a boolean `cond_fn` and a root finder are passed to the event since this should never make sense then?

Hmm, I'm not too sure about this. There might be cases where you have one real-valued event function and one boolean. E.g., in the bouncing ball example we might want to add an extra function to `cond_fn` stopping the solve when the velocity is low enough (i.e., the ball has reached a steady state and stopped bouncing). But happy to hear your thoughts as well @patrick-kidger.

> If I pass multiple cond_fn, what would you suggest to determine which of them caused the Event? :)
Thanks for the help!

The solution returned by `diffeqsolve` has an attribute called `event_mask` which is a PyTree of the same structure as your `cond_fn` where each leaf is `False` if the corresponding condition function did not trigger an event and `True` otherwise. (Note: for now, only one leaf can be `True`). If you're interested in a more involved example of how this all works with multiple event handling you might want to check out [this repo](https://github.com/cholberg/event_rde/tree/main) (specifically [snn.py](https://github.com/cholberg/event_rde/blob/main/event_rde/snn.py)). This is a very much a work in progress, but it should serve as an example of how it all works. Hope that helps!

patrick-kidger:
Agreed on the first point. Happy to add a warning if all events are Boolean, though -- no strong feelings.

dkweiss31:
Hi folks! Very excited about this PR, as I'm thinking about quantum jump applications for [dynamiqs](https://github.com/dynamiqs/dynamiqs). I'm unfortunately running into an error If I try to pass an option different from `saveat = SaveAt(t1=True)` :


```python
import diffrax as dx
import optimistix as optx
import jax
import jax.numpy as jnp

term = dx.ODETerm(lambda t, y, args: y)
solver = dx.Tsit5()
t0 = 0
t1 = 100.0
dt0 = 1.0
y0 = 1.0
ts = jnp.arange(t0, t1, dt0)

def cond_fn(state, y, **kwargs):
    assert isinstance(state.y, jax.Array)
    return y - jnp.exp(1.0)

fn = lambda t, y, args: y

subsaveat_a = dx.SubSaveAt(ts=ts, fn=fn)  # save solution regularly
subsaveat_b = dx.SubSaveAt(t1=True)  # save last state
saveat = dx.SaveAt(subs=[subsaveat_a, subsaveat_b])

root_finder = optx.Newton(1e-5, 1e-5, optx.rms_norm)
event = dx.Event(cond_fn, root_finder)
sol = dx.diffeqsolve(
    term, solver, t0, t1, dt0, y0, saveat=saveat, event=event
)
``` 

This runs into an error on line 1204 of `_integrate.py`
```python
    ys = jtu.tree_map(lambda _y, _yevent: _y.at[-1].set(_yevent), ys, yevent)
ValueError: Expected list, got Traced<ShapedArray(float32[1])>with<DynamicJaxprTrace(level=1/0)>.
```
Thanks!

cholberg:
Ah, yes I forgot to handle the case where multiple `SubSaveAt` are passed. Should be fixed now with the latest commit. Let me know if that works for you.

dkweiss31:
Thanks for the quick response! So that fixed the example I posted, however I am still running into issues on slightly more complicated examples more in line with how dynamiqs actually calls `diffeqsolve` and more specifically how it saves data as the simulation progresses. Here is a MWE where `y` saves the state and `y2` saves "expectation values". 

```python
import diffrax as dx
import optimistix as optx
import jax
import jax.numpy as jnp
import equinox as eqx
from jax import Array

term = dx.ODETerm(lambda t, y, args: y + t)
solver = dx.Tsit5()
t0 = 0
t1 = 100.0
dt0 = 1.0
y0 = jnp.array([[1.0], [0.0]])
ts = jnp.arange(t0, t1, dt0)


def cond_fn(state, y, **kwargs):
    assert isinstance(state.y, jax.Array)
    norm = jnp.einsum("ij,ij->", y, y)
    return norm - jnp.exp(1.0)


class Saved(eqx.Module):
    y: Array
    y2: Array

def save_fn(t, y, args):
    ynorm = jnp.einsum("ij,ij->", y, y)
    return Saved(y, jnp.array([ynorm, 3 * ynorm]))

subsaveat_a = dx.SubSaveAt(ts=ts, fn=save_fn)  # save solution regularly
subsaveat_b = dx.SubSaveAt(t1=True)  # save last state
saveat = dx.SaveAt(subs=[subsaveat_a, subsaveat_b])

root_finder = optx.Newton(1e-5, 1e-5, optx.rms_norm)
event = dx.Event(cond_fn, root_finder)
sol = dx.diffeqsolve(
    term, solver, t0, t1, dt0, y0, saveat=saveat, event=event
)

``` 
This runs into 
```python
ValueError: Cannot broadcast to shape with fewer dimensions: arr_shape=(2, 2) shape=(2,)
``` 
Interestingly, the code runs without errors if I replace `return Saved(y, jnp.array([ynorm, 3 * ynorm]))
` with `return Saved(y, jnp.array([y, 3 * y]))
`

cholberg:
You're right I did not account for the fact that `SubSaveAt.fn` could return a PyTree. Should be fixed now. At least your MWE works with the latest commit.

dkweiss31:
Indeed that fixed my MWE! I hate to be such a pain but I am now running into another issue, here is an example that is now much closer to the actual code I am interested in running. 

```python
import diffrax as dx
import optimistix as optx
import jax.numpy as jnp

L_op = 0.1 * jnp.array([[0.0, 1.0],
                        [0.0, 0.0]], dtype=complex)
H = 0.0 * L_op
t0 = 0
t1 = 100.0
dt0 = 1.0
y0 = jnp.array([[0.0], [1.0]], dtype=complex)


def vector_field(t, state, _args):
    L_d_L = jnp.transpose(L_op) @ L_op
    new_state = -1j * (H - 1j * 0.5 * L_d_L) @ state
    return new_state


def cond_fn(state, **kwargs):
    psi = state.y
    prob = jnp.abs(jnp.einsum("id,id->", jnp.conj(psi), psi))
    return prob - 0.95


term = dx.ODETerm(vector_field)
root_finder = optx.Newton(1e-5, 1e-5, optx.rms_norm)
event = dx.Event(cond_fn, root_finder)

sol = dx.diffeqsolve(
    term, dx.Tsit5(), t0, t1, dt0, y0, event=event
)
``` 

This runs into 

```python
equinox._errors.EqxRuntimeError: The linear solver returned non-finite (NaN or inf) output. This usually means that the
operator was not well-posed, and that the solver does not support this.
``` 
It's possible this could be due to my use of complex numbers, which as I understand are only partly supported in diffrax? However with the previous `DiscreteTerminatingEvent` I did not run into such errors. Note that if I change 0.95 to 0.0 I don't see this error (likely because the event is not triggered)

patrick-kidger:
Complex number support is definitely still iffy. Can you try reproducing this without using them? You can still solve the same equation, mathematically speaking, just by splitting things into separate real and imaginary components.

dkweiss31:
Right, here is the same example using the complex->real isomorphism described e.g. [here](https://arxiv.org/abs/1612.04929) (see Eq. (9)). I am getting the same error as before, so it seems then this is not a complex number issue

```python
import diffrax as dx
import optimistix as optx
import jax.numpy as jnp


def mat_cmp_to_real(matrix):
    re_matrix = jnp.real(matrix)
    im_matrix = jnp.imag(matrix)
    top_row = jnp.hstack((re_matrix, -im_matrix))
    bottom_row = jnp.hstack((im_matrix, re_matrix))
    return jnp.vstack((top_row, bottom_row))


def vec_cmp_to_real(vector):
    re_vec = jnp.real(vector)
    im_vec = jnp.imag(vector)
    return jnp.vstack((re_vec, im_vec))


L_op = 0.1 * jnp.array([[0.0, 1.0],
                        [0.0, 0.0]], dtype=complex)
L_d_L = jnp.transpose(L_op) @ L_op
H = 0.0 * L_op
_prop = -1j * (H - 1j * 0.5 * L_d_L)
_y0 = jnp.array([[0.0], [1.0]], dtype=complex)
prop = mat_cmp_to_real(_prop)
y0 = vec_cmp_to_real(_y0)
t0 = 0
t1 = 100.0
dt0 = 1.0


def vector_field(t, state, _args):
    new_state = prop @ state
    return new_state


def cond_fn(state, **kwargs):
    psi = state.y
    prob = jnp.abs(jnp.einsum("id,id->", jnp.conj(psi), psi))
    return prob - 0.95


term = dx.ODETerm(vector_field)
root_finder = optx.Newton(1e-5, 1e-5, optx.rms_norm)
event = dx.Event(cond_fn, root_finder)

sol = dx.diffeqsolve(
    term, dx.Tsit5(), t0, t1, dt0, y0, event=event
)
```

patrick-kidger:
Thankyou @dkweiss31!
@cholberg, are you able to have a look at this example?

Anyway, as promised! Getting this in is my next priority. As such I've gone through and submitted a PR against this branch [here](https://github.com/cholberg/diffrax/pull/1). I don't claim that everything I've done is necessarily correct, so @cholberg I'd appreciate a review! :D

cholberg:
@dkweiss31, sorry for being a little unresponsive these last couple of days! I will be quite busy the next 2-3 days, but will try to look at it asap.

@patrick-kidger, oh, that's great! But yea, deadline coming up so can't promise that I'll have time to look at it before Thursday. Will make it a priority afterwards, though :)

dkweiss31:
@patrick-kidger FWIW both of my examples run without errors using your branch [387-tweaks](https://github.com/cholberg/diffrax/pull/1)! Only change I had to make was changing the signature of my conditional function to 
```python
def cond_fn(t, y, *args, **kwargs):
``` 

 Moreover, the behavior is as expected and the solve terminates at e.g. 0.95. (Something interesting is that if I ask it to save intermediate values, the saved output includes values for times after the termination time. I assume this is as expected, as the code then backtracks to the point where `cond_fn` changes sign?) Thanks to you and to @cholberg for your assistance!

patrick-kidger:
I'm glad it now works! I have no idea what I changed to make that happen, although I did have a variety of small bugfixes as well.

I think saving anything after the termination time should be considered a bug though :) I knew I must have missed something! Assuming the above MWE demonstrates this, I'll leave this to @cholberg to consider post-NeurIPS-deadline :)

patrick-kidger:
For the second one that's probably on us.

For the first one, you are running afoul of https://docs.astral.sh/ruff/rules/function-uses-loop-variable/

LuggiStruggi:
Cool thanks! Wow I didnt know about this :D thanks for the help!

cholberg:
@dkweiss31, @patrick-kidger: I ran your MWE above with `steps=True` (I am assuming that is what you mean, but correct me if I am wrong). I actually did not notice when walking through [387-tweaks](https://github.com/cholberg/diffrax/pull/1), but it turns out that the values returned by the root finding are not saved anywhere in this case which is also what caused the confusion I am expecting. This should be fixed with the latest commit, but, as suggested by @patrick-kidger, I really need to go through and write tests for all the different saving configurations in combination with events to ensure that everything works as expected. Do let me know if you encounter anymore unexpected behaviours.

cholberg:
@LuggiStruggi: With the latest commits, your second example seems to work for me :)

LuggiStruggi:
cool thanks :) sorry maybe i didnt pull recently enough

dkweiss31:
Hey friends, pulling from the most recent version of cholberg:dev the following example fails:

```python
import diffrax as dx
import optimistix as optx
import jax.numpy as jnp
import equinox as eqx
from jax import Array


L_op = 0.1 * jnp.array([[0.0, 1.0],
                        [0.0, 0.0]], dtype=complex)
t0 = 0
t1 = 100.0
dt0 = 1.0
y0 = jnp.array([[0.0], [1.0]], dtype=complex)
ts = jnp.arange(t0, t1, dt0)


class Saved(eqx.Module):
    y: Array

def save_fn(t, y, args):
    ynorm = jnp.einsum("ij,ij->", y, y)
    return Saved(jnp.array([ynorm,]))

subsaveat_a = dx.SubSaveAt(ts=ts, fn=save_fn)
subsaveat_b = dx.SubSaveAt(t1=True)
saveat = dx.SaveAt(subs=[subsaveat_a, subsaveat_b])


def vector_field(t, state, _args):
    L_d_L = jnp.transpose(L_op) @ L_op
    new_state = -1j * (- 1j * 0.5 * L_d_L) @ state
    return new_state


def cond_fn(t, y, *args, **kwargs):
    psi = y
    prob = jnp.abs(jnp.einsum("id,id->", jnp.conj(psi), psi))
    return prob - 0.95


term = dx.ODETerm(vector_field)
root_finder = optx.Newton(1e-5, 1e-5, optx.rms_norm)
event = dx.Event(cond_fn, root_finder)

sol = dx.diffeqsolve(
    term, dx.Tsit5(), t0, t1, dt0, y0, event=event, saveat=saveat
)
``` 

on line 715 of `_integrate.py` with 
```python
    save_index = final_state.save_state.save_index - 1
AttributeError: 'list' object has no attribute 'save_index'
```

patrick-kidger:
@dkweiss31 -- thank you for the report! @cholberg -- looks like it should be a tree-map'd `save_index = jtu.tree_map(lambda x: x.save_index - 1, final_state.save_state)`. I guess let's add "a PyTree of `SubSaveAt`s" the list of thing things we should add a test for.

cholberg:
Should be fixed with the latest commit. I also added more tests for events. Will get around to adding tests for different `SaveAt` structures next which should hopefully prevent such errors in the future. 

@patrick-kidger: One quick question on the way non-scalar `cond_fn` is handled right now. In principle, the user could pass a single function to `cond_fn` that returns a PyTree. In this case the following error would not be triggered, since we call `jtu.tree_flatten` on the output of that function:

```python
event_values_leaves, event_structure = jtu.tree_flatten(event_values)
event_mask_leaves = []
for event_value_i in event_values_leaves:
    if jnp.shape(event_value_i) != ():
        raise ValueError(
            "Event functions must return a scalar, got shape "
            f"{jnp.shape(event_value_i)}."
        )
```

I guess we should just add another check to ensure that each function in `cond_fn` does not return a PyTree?

patrick-kidger:
Good catch!
Completely agreed. Probably the simplest thing to do would be to promote this lambda into a normal function:

https://github.com/patrick-kidger/diffrax/blob/05f636f79fc7bda1c32d5b797a1bfbfab038fdd0/diffrax/_integrate.py#L1203

and add the shape+dtype checks there?

dkweiss31:
> Should be fixed with the latest commit. I also added more tests for events. Will get around to adding tests for different `SaveAt` structures next which should hopefully prevent such errors in the future.
> 
Unfortunately it seems the same example is still failing, with the same error this time on line 722
```python
AttributeError: 'list' object has no attribute 'save_index'
```

dkweiss31:
Also: I'm not sure if this is the appropriate place to post this issue but this code seems to run fine on the master branch of diffrax but fails with this strange error on the cholberg:dev branch: on line 1210, in equinox/internal/_loop/checkpointed.py `_resolve_perturb_val_impl` (note that events are not involved at all for this simulation)

```python
    assert new_perturb_val is not sentinel
AssertionError
``` 

```python
import jax
import jax.numpy as jnp
import dynamiqs as dq

H0 = dq.sigmaz()
H1s = [dq.sigmax(), dq.sigmay()]
tsave = jnp.linspace(0, 1.0, 21)
psi0 = dq.basis(2, 1)

def cost_function(amps):
    Hc = jnp.einsum("i,ijk->jk", amps, H1s)

    def _H(t):
        return H0 + Hc

    H = dq.timecallable(_H)
    result = dq.sesolve(H, psi0, tsave, options=dq.Options(save_states=False))
    return dq.expect(dq.sigmaz(), result.states).real

amps = jnp.array([0.01, 0.02])
print(jax.grad(cost_function)(amps))
``` 
All good if this is a strictly dynamiqs issue! But it seems strange to me that it breaks on one version of diffrax but not the other

cholberg:
Okay, finally done adding more comprehensive testing. It took a little while because of some unforseen difficulties:

1. As we discussed above, we need to error out when the user provides a single function to `cond_fn` that returns a non-trivial PyTree. What I ended up doing, was the following:
https://github.com/cholberg/diffrax/blob/1202f9828cea701046aa33e02f72debad9fe2d0b/diffrax/_integrate.py#L1277-L1332

2. Figuring out the right way to save the values of the solver is a little involved and I am not sure my current solution is optimal. Mainly the issue is that when `SaveAt(ts=ts)` is passed for some array `ts`, it might happen that values are saved after the time an event has happened (because these are saved as part of the step and we only do the root find outside the loop). As a result, we need to go back and delete these values once we know the exact event time. This is complicated further if also e.g. `steps=True`. Current solution is:
https://github.com/cholberg/diffrax/blob/1202f9828cea701046aa33e02f72debad9fe2d0b/diffrax/_integrate.py#L723-L797

@patrick-kidger, I would really appreciate a review on these two changes (and the addtional tests if time permits). I will add documentation and update the examples tomorrow :) 

@dkweiss31, really that was my bad for not checking your example after making the changes. For what it's worth, it worked for me with the latest commit.

patrick-kidger:
Great stuff!

I'll do a more thorough review in a bit, but two quick comments on the bits you've highlighted:

- I think this line:
https://github.com/cholberg/diffrax/blob/1202f9828cea701046aa33e02f72debad9fe2d0b/diffrax/_integrate.py#L1307
should be a `ValueError` (not an assertion), as it can occur if the user returns the wrong dtype.

- For the "unsaving", can we simply switch the event-detection and the saving inside the body of the loop? And then condition the saving on whether an event occurred or not. (To be precise: whether the time to save is less than the time of the event! *Not* just whether an event occurred. For `SaveAt(ts=...)` then we can save multiple time on each step.) For context we already do this for steps that are rejected by the adaptive step sizing, this is the `pred` option here:
https://github.com/cholberg/diffrax/blob/1202f9828cea701046aa33e02f72debad9fe2d0b/diffrax/_integrate.py#L458

cholberg:
> * I think this line:
>   https://github.com/cholberg/diffrax/blob/1202f9828cea701046aa33e02f72debad9fe2d0b/diffrax/_integrate.py#L1307
>   should be a `ValueError` (not an assertion), as it can occur if the user returns the wrong dtype.

Ah yes, I was wondering about that as well, but forgot to ask you about it. Agreed!

> * For the "unsaving", can we simply switch the event-detection and the saving inside the body of the loop? And then condition the saving on whether an event occurred or not. (To be precise: whether the time to save is less than the time of the event! _Not_ just whether an event occurred. For `SaveAt(ts=...)` then we can save multiple time on each step.) For context we already do this for steps that are rejected by the adaptive step sizing, this is the `pred` option here:
>   https://github.com/cholberg/diffrax/blob/1202f9828cea701046aa33e02f72debad9fe2d0b/diffrax/_integrate.py#L458

So you mean doing the root finding inside the loop as well (ofc only when an event is triggered)? Because you would need to know the exact event time before you know which values to save.

cholberg:
Added the last few changes now. Note a few additional changes:
- I updated the docstring for `steady_state_event` since this now returns a function to be passed to `cond_fn` instead of returning an instance of `Event`.
- The docstring for `ImplicitAdjoint` was referencing the deprecated `SteadyStateEvent` so I changed it to be in line with the new event syntax.

patrick-kidger:
> So you mean doing the root finding inside the loop as well (ofc only when an event is triggered)? Because you would need to know the exact event time before you know which values to save.

Oh bother, I see what you're saying.

No, we should definitely do the root find outside the loop.

I think you're right, "unsaving" may be the best we can do. Ach, that's frustrating.

patrick-kidger:
(By the way @cholberg heads-up on the test failures.)

cholberg:
Yea, for some reason the type checker was complaining when I passed `None` as the second argument to `jtu.tree_transpose`. Weirdly, the pre-commit hooks did not throw an error when I comitted, though. Anyway, I now pass an explicit PyTree structure to avoid this issue. Hopefully this fixes it.

dkweiss31:
Hey @cholberg and @patrick-kidger I'm wondering if you saw my above MWE that runs fine on the master branch but fails on this one?

cholberg:
Hmm, I haven't had time to look at it in detail, but at first glance I am not sure if this is related to the changes made here since ideally we would never be on the "event branch" in your example. Maybe @patrick-kidger has some intuition about what might be going wrong here.

patrick-kidger:
@dkweiss31 -- thanks for the MWE! This should now be fixed in https://github.com/patrick-kidger/equinox/pull/752.

@cholberg -- I'm sure this branch still needs a review! I'll try to find the time... 😅

cholberg:
@patrick-kidger: Not sure why `test_implicit` failed since it worked locally. But I noticed that it used the old `SteadyStateEvent` syntax so I updated it to use the new one. Everything works locally so hopefully this should fix it.

And yes, a review would be much appreciated! :)

dkweiss31:
@patrick-kidger thanks for the fix! Indeed that example now works. (Based on your fix it seems strange that the code worked everywhere but on the cholberg:dev branch? No doubt I'm missing something simple.)  I'm unfortunately still running into errors on this branch, though I am struggling to get a MWE that only uses diffrax/equinox/optimistix code, I'll update when I can. In case its helpful, the error is

```python
NotImplementedError: Transpose rule (for reverse-mode differentiation) for 'stop_gradient' not implemented
```

patrick-kidger:
Whoops, GitHub auto-closed this from a comment over in Equinox, it seems...

patrick-kidger:
Okay, I've just done a review! I've opened https://github.com/cholberg/diffrax/pull/2 with some minor doc tweaks.

Besides that I have only one comment, and that's on the unsaving: I think what we have here may misbehave if you have a combination e.g. `SaveAt(ts=..., steps=...)` or `SaveAt(steps=..., t1=...)` etc.

Fortunately, I think the fix is really simply! Can we simply make the unsaving just be something like `ys = ys.at[ts > tfinal].set(0)`? I think this should handle all possible combinations.

Modulo the above -- let me know what you think -- I'm now happy to merge this!

(The test failures are due to an upstream JAX change, and already fixed on main. Rebase and these should go away.)

cholberg:
@patrick-kidger: Perfect, thanks! I have merged your tweaks and rebased. Regarding the unsaving, I am not completely sure your approach would work since we need to update the `save_index` as well for when we save the final values after the while loop: https://github.com/cholberg/diffrax/blob/84bd8cb65e2a19f3ca7544fb9070b55e017c8b06/diffrax/_integrate.py#L815-L817. 
Unless you mean to do the unsaving at the very end (after saving the final value) in which case the result would in some cases look something like `[t1, t2, ..., tn, jnp.inf, jnp.inf, ..., jnp.inf, tfinal, jnp.inf, ...]` (and similarly for `ys`) which I think you want to avoid?

For what it is worth, `test_event.py::test_event_ts()` actually checks whether anything breaks when passing `SaveAt(ts=..., steps=...)`. For all the configurations I checked, everything seems to work out, but I am happy to make changes.

cholberg:
> @patrick-kidger thanks for the fix! Indeed that example now works. (Based on your fix it seems strange that the code worked everywhere but on the cholberg:dev branch? No doubt I'm missing something simple.) I'm unfortunately still running into errors on this branch, though I am struggling to get a MWE that only uses diffrax/equinox/optimistix code, I'll update when I can. In case its helpful, the error is
> 
> ```python
> NotImplementedError: Transpose rule (for reverse-mode differentiation) for 'stop_gradient' not implemented
> ```

I wonder whether this is because your example involves a brownian control? As it is right now, events are not differentiable when you add a diffusion term (essentially since you are trying to compute a time derivative of Brownian motion which in theory does not exist and which in practice `diffrax` will prevent you from doing: https://github.com/patrick-kidger/diffrax/blob/f19b66f66beb2458fdeba01adee9108251df18fd/diffrax/_brownian/tree.py#L258

patrick-kidger:
Good point about also needing to update the `save_index` as well. There may yet be a more elegant way to do this, but if nothing else I think updating it by something like `-jnp.sum((ts > tfinal) & jnp.invert(jnp.isposinf(ts)))` would probably accompilsh this?

(My hope here is that unsaving can be made really simple.)

> which I think you want to avoid?

Agreed, let's definitely not do this!

dkweiss31:
> 
> I wonder whether this is because your example involves a brownian control? As it is right now, events are not differentiable when you add a diffusion term (essentially since you are trying to compute a time derivative of Brownian motion which in theory does not exist and which in practice `diffrax` will prevent you from doing:
> 
> https://github.com/patrick-kidger/diffrax/blob/f19b66f66beb2458fdeba01adee9108251df18fd/diffrax/_brownian/tree.py#L258

@cholberg Not sure I follow, what brownian control are you referring to? I'm running code that worked based on the old events functionality and was differentiable in all the ways I needed, but is now throwing errors. I haven't had time yet to nail down a MWE based only on diffrax code, but will try to do so this week.

dkweiss31:
Here is an example showcasing the error I mentioned before. Sorry that it is a bit long, I've tried to shorten it as much as I can. I suspect something weird is going on with `_cartesian_vectorize` and `flat_vectorize` which are trying to allow us to batch over H, jump_ops and psi0. I can post the full error trace if needed (same error that I mentioned above).

```python
import jax
import jax.numpy as jnp
import diffrax as dx
import equinox as eqx
import optimistix as optx
from jax import Array
from jaxtyping import PyTree, ScalarLike
import jax.tree_util as jtu


class Shape(tuple):
    """Helper class to help with Pytree handling."""


def is_shape(x: object) -> bool:
    return isinstance(x, tuple())


class CallableTimeArray(eqx.Module):
    f: PyTree  # (n, n)

    @property
    def in_axes(self) -> PyTree[int]:
        return CallableTimeArray(Shape())

    def __call__(self, t: ScalarLike) -> Array:
        return self.f(t)


class Result(eqx.Module):
    ylast: Array | None


class MCResult(eqx.Module):
    tsave: Array
    _no_jump_res: Result

    @property
    def final_no_jump_state(self) -> Array:
        return self._no_jump_res.ylast


def _cartesian_vectorize(
    f: callable, n_batch: PyTree[int], out_axes: PyTree[int | None]
) -> callable:
    leaves, treedef = jtu.tree_flatten((None,) + n_batch[1:], is_leaf=is_shape)
    for i, leaf in reversed(list(enumerate(leaves))):
        leaf_len = len(leaf)
        if leaf_len > 0:
            in_axes = jtu.tree_map(lambda _: None, leaves)
            in_axes[i] = 0
            in_axes = jtu.tree_unflatten(treedef, in_axes)
            for _ in range(leaf_len):
                f = jax.vmap(f, in_axes=in_axes, out_axes=out_axes)
    return _flat_vectorize(f, n_batch[:1] + (None,) * len(n_batch[1:]), out_axes)


def _flat_vectorize(
    f: callable, n_batch: PyTree[int], out_axes: PyTree[int | None]
) -> callable:
    broadcast_shape = jtu.tree_leaves(n_batch, is_shape)
    broadcast_shape = jnp.broadcast_shapes(*broadcast_shape)
    in_axes = jtu.tree_map(
        lambda x: 0 if len(x) > 0 else None, n_batch, is_leaf=is_shape
    )
    for _ in range(len(broadcast_shape)):
        f = jax.vmap(f, in_axes=in_axes, out_axes=out_axes)
    return f


class AbstractSolver(eqx.Module):
    def run_traj(self, H, jump_ops, psi0):
        def vector_field(t, state, _args):
            Ls = jnp.stack([L(t) for L in jump_ops])
            Lsd = Ls.mT.conj()
            LdL = (Lsd @ Ls).sum(axis=0)
            new_state = -1j * (H(t) - 1j * 0.5 * LdL) @ state
            return new_state

        def cond_fn(t, y, *args, **kwargs):
            psi = y
            prob = jnp.abs(jnp.einsum("id,id->", jnp.conj(psi), psi))
            return prob - no_jump_prob

        term = dx.ODETerm(vector_field)
        root_finder = optx.Newton(1e-5, 1e-5, optx.rms_norm)
        event = dx.Event(cond_fn, root_finder)
        solution = dx.diffeqsolve(
            term, dx.Tsit5(), t0=t0, t1=ts[-1], dt0=dt0, y0=psi0, event=event,
        )
        return MCResult(ts, Result(solution.ys))

    def run(self, H, jump_ops, psi0) -> PyTree:
        out_axes = MCResult(None, 0)
        n_batch = (
            H.in_axes,
            [jump_op.in_axes for jump_op in jump_ops],
            Shape(psi0.shape[:-2]),
        )
        f = _cartesian_vectorize(self.run_traj, n_batch, out_axes)
        return f(H, jump_ops, psi0)


t0 = 0.0
t1 = 100.0
dt0 = 1.0
y0 = jnp.array([[0.0], [1.0]], dtype=complex)
jump_ops = [0.1 * jnp.array([[0.0, 1.0],
                        [0.0, 0.0]], dtype=complex),]
H = jnp.array([[1.0, 0.0],
               [0.0, -1.0]], dtype=complex)
ts = jnp.arange(t0, t1, dt0)
no_jump_prob = 0.0


def cost_function(amp):
    abs_solver = AbstractSolver()
    mcres = abs_solver.run(
        CallableTimeArray(jtu.Partial(lambda t: amp * H)),
        [CallableTimeArray(jtu.Partial(lambda t: L)) for L in jump_ops],
        jnp.asarray([y0, ]),
    )
    return jnp.log(jnp.sum(jnp.abs(mcres.final_no_jump_state)))


amps = 0.01
print(jax.grad(cost_function)(amps))

```

cholberg:
@patrick-kidger: Okay, I have now simplified the unsaving somewhat in that there is now only one `unsave` function which behaves more or less as your suggestion. One thing to notice is that I have had to replace the `.at[ts > tfinal]` parts with `jnp.where` to avoid indexing with a tracer. The expression for `ys` is not *super* readable:
https://github.com/cholberg/diffrax/blob/05bafb218f5396a397f4e1790b3ef519a8e8b7ee/diffrax/_integrate.py#L737
I just had to make sure that the arrays could be broadcasted and that the returned `ys` are of the right dimension. 

Anyway, all tests (related to events) still pass with this new modifcations so hopefully everything is good.

diffrax/_integrate.py
```diff
@@ -458,16 +620,154 @@ def body_fun(state):
...
+        tfinal, yfinal, result = lax.cond(
+            eqxi.unvmap_any(event_happened),
+            _root_find,
+            lambda: (final_state.tprev, final_state.y, result),
+        )
+
+        # We delete all the saved values after the event time.
+        def unsave(subsaveat: SubSaveAt, save_state: SaveState) -> SaveState:
+            ts = save_state.ts
+            mask = ts >= tfinal
```
patrick-kidger, line 729: IIUC this is `>=` and not `>` because we only want to save this final time on the `_save_t1` branch below? Is that correct?

diffrax/_integrate.py
```diff
@@ -458,16 +620,154 @@ def body_fun(state):
...
+            _save_index = save_state.save_index - jnp.sum(mask & (ts < jnp.inf))
+            _saveat_ts_index = save_state.saveat_ts_index - jnp.sum(
+                mask & (ts < jnp.inf)
+            )
+            _ts = jnp.where(mask, jnp.inf, ts)
+            _ys = jtu.tree_map(
+                lambda _, __ys: jnp.where(
+                    mask[(...,) + (jnp.newaxis,) * (__ys.ndim - 1)], jnp.inf, __ys
+                ),
+                subsaveat.fn(tfinal, yfinal, args),
```
patrick-kidger, line 739: Why this additional argument? I think `save_state.ys` should already have the correct PyTree structure?

(Perhaps this has been copied forward from something inside the while-loop-over-steps? Because at that point `ys` doesn't necessarily have the correct structure.)

patrick-kidger (commented):
I think the unsaving looks good to me!
Do you want to replay your changes on top of the latest `dev`? Then we'll get this merged!

diffrax/_integrate.py
patrick-kidger: (FWIW if you ever really do need a dummy value just for this kind of tree-manipulation then using `jax.eval_shape`, or `eqx.filter_eval_shape`, will save some compile time.)

diffrax/_integrate.py
cholberg: Yup, exactly!

diffrax/_integrate.py
cholberg: Ah yes, you're absolutely right! Fixed it in the latest commit :)

cholberg:
Perfect! Rebased on the latest `dev` now :)

johannahaffner:
Hi All! 

I just opened #448, I get the same error on `diffrax:main` and `cholberg/diffrax:dev`. 
When solver tolerances are `atol=1e-06, rtol=1e-03` on `cholberg/diffrax:dev`, the solver terminates without a single step (?), reported `t1` is `0.0`.

johannahaffner:
Alright, I was merely being obtuse and didn't think about the most obvious thing: how the absolute and relative tolerances actually affect the terminating condition, and what that is based on. I assumed that it would compare rates of change $dy$ for consecutive steps. If you think that others might make the same mistake, I suggest adding a sentence to the documentation, something like 

```python
class SteadyStateEvent(AbstractDiscreteTerminatingEvent):
    """Terminates the solve once it reaches a steady state, defined as 
    norm(dy) < atol + rtol * norm(y).
    """
```

diffrax/_integrate.py
patrick-kidger: Actually, I think this is wrong! If we have `SaveAt(ts=[t1] t1=True)` then Diffrax will -- when not using events -- actually save two copies of the output! (One could argue whether that's desirable, but anyway it is what we do.)

So I think we actually do want `>` here, so that we continue to follow the same behaviour.

cholberg: Sounds fine to me! (See latest commit)

patrick-kidger:
Aaaaand... merged! 🎉

Thank you for all your hard work @cholberg! This was a massive PR, so getting this in is no small feat. Especially in an large existing codebase like Diffrax.

This is now on the `dev` branch, which I'll be merging into `main` and doing a new release for shortly. :)

patrick-kidger:
By the way, @dkweiss31 -- since I've not seen any movement on https://github.com/google/jax/issues/22011, my guess is that we should probably work around this by registering a transpose rule for `lax.stop_gradient` ourselves. This transpose rule should just be the identity function.

I'd be happy to take a PR on this!

dkweiss31:
Nice I'd love to take a swing at that! This would probably want to be in equinox rather than diffrax?

patrick-kidger:
Hmm, I think probably Optimistix actually, thinking about it.
This is what's doing the manual linearisation and transposition, so it's there that we fall afoul of this bug.

dkweiss31:
Hey @patrick-kidger, sorry for the delay I was on vacation last week :).

Please forgive and apologies in advance for my likely elementary jax mistakes: as a test I've tried doing the simplest possible thing, and added

```python
from jax.interpreters import ad
from jax._src.ad_util import stop_gradient_p


def stop_gradient_transpose(ct, x):
    return ct

ad.primitive_transposes[stop_gradient_p] = stop_gradient_transpose
``` 

to the top of your example in https://github.com/google/jax/issues/22011 as well as in my above example. Interestingly, your example now runs, while my example fails on line 720 of `_integrate.py`

```python
line 720, in loop
    tfinal, yfinal, result = lax.cond(
jax._src.source_info_util.JaxStackTraceBeforeTransformation: AssertionError
``` 
Any idea what might be happening? (Also happy to move this conversation elsewhere, but just thought I'd keep it on this thread for continuities sake :))

patrick-kidger:
Happy to keep the discussion here!
The fix is just three characters: you want `return (ct,)`. :)

With this, your example works for me!

dkweiss31:
Ha! I knew it was something silly. Thank you! Just opened a PR in optimistix :)

dkweiss31:
I am, unfortunately, still encountering errors after the transpose fix. See my latest MWE, which fails with 
```python
EqxRuntimeError: Terminating differential equation solve because an event occurred.
``` 
after a few times through the for loop of the optimizer. Such a bizarre error, since I expect the differential equation to terminate!

```python
from functools import partial
import optax
import jax
import jax.numpy as jnp
import diffrax as dx
import optimistix as optx


def run_traj(H, jump_ops, psi0):
    def vector_field(t, state, _args):
        Ls = jnp.stack([L for L in jump_ops])
        LdL = (Ls.mT.conj() @ Ls).sum(axis=0)
        new_state = -1j * (H - 1j * 0.5 * LdL) @ state
        return new_state

    def cond_fn(t, y, *args, **kwargs):
        return jnp.abs(jnp.conj(y) @ y) - no_jump_prob

    event = dx.Event(cond_fn, optx.Newton(1e-5, 1e-5, optx.rms_norm))
    solution = dx.diffeqsolve(
        dx.ODETerm(vector_field), dx.Tsit5(), t0=ts[0], t1=ts[-1], dt0=dt0, y0=psi0, event=event,
    )
    return solution.ys


def run(H, jump_ops, psi0):
    f = jax.vmap(run_traj, in_axes=(None, None, 0))
    return f(H, jump_ops, psi0)


def optimize(params_to_optimize):
    opt_state = optimizer.init(params_to_optimize)
    for epoch in range(10):
        params_to_optimize, opt_state = step(params_to_optimize, opt_state)
    return params_to_optimize


@partial(jax.jit)
def step(params_to_optimize, opt_state,):
    grads = jax.grad(loss)(params_to_optimize,)
    updates, opt_state = optimizer.update(grads, opt_state)
    params_to_optimize = optax.apply_updates(params_to_optimize, updates)
    return params_to_optimize, opt_state


def loss(params_to_optimize):
    res = run(params_to_optimize * H1, [a, ], initial_states)
    return jnp.log(jnp.sum(jnp.abs(res)))


dim = 4
a = jnp.diag(jnp.sqrt(jnp.arange(1, stop=dim, dtype=complex)), k=1)
H1 = a + a.mT.conj()
ket0 = jnp.array([1.0, 0.0, 0.0, 0.0], dtype=complex)
ket1 = jnp.array([0.0, 1.0, 0.0, 0.0], dtype=complex)
initial_states = jnp.asarray([ket0, ket1])
dt0 = 1.0
ts = jnp.arange(0.0, 100.0, dt0)
no_jump_prob = 0.9
optimizer = optax.adam(learning_rate=0.9)
init_drive_params = 0.1
opt_params = optimize(
    params_to_optimize=init_drive_params,
)
```

dkweiss31:
Sorry, the previous example was not exactly minimal. Here is a more minimal example that doesn't need the optimization layer. The error is now

```python
equinox._errors.EqxRuntimeError: The maximum number of steps was reached in the nonlinear solver. The problem may not be solveable (e.g., a root-find on a function that has no roots), or you may need to increase `max_steps`.
``` 
I've tried increasing the number of max steps to no avail. Interestingly, choosing `init_drive_params = -1.1` or `1.2` runs without an error
```python
import jax.numpy as jnp
import diffrax as dx
import optimistix as optx


def run_traj(H, jump_op, psi0):
    def vector_field(t, state, _args):
        LdL = jump_op.mT.conj() @ jump_op
        new_state = -1j * (H - 1j * 0.5 * LdL) @ state
        return new_state

    def cond_fn(t, y, *args, **kwargs):
        return jnp.abs(jnp.conj(y) @ y) - no_jump_prob

    event = dx.Event(cond_fn, optx.Newton(1e-5, 1e-5, optx.rms_norm))
    solution = dx.diffeqsolve(
        dx.ODETerm(vector_field), dx.Tsit5(), t0=ts[0], t1=ts[-1], dt0=dt0, y0=psi0, event=event
    )
    return solution.ys


a = jnp.diag(jnp.sqrt(jnp.arange(1, stop=4, dtype=complex)), k=1)
ket1 = jnp.array([0.0, 1.0, 0.0, 0.0], dtype=complex)
dt0 = 1.0
ts = jnp.arange(0.0, 100.0, dt0)
no_jump_prob = 0.9
init_drive_params = -1.19
res = run_traj(init_drive_params * (a + a.mT.conj()), a, ket1)
```

patrick-kidger:
Hmm! I think a big part of what this is speaking to is the need for better debug tooling.

There is one big thing I've been wanting to do here, which is to stop using `throw=False` on all of our calls to Optimistix. Instead, we should respect the value of `diffeqsolve(..., throw=...)`. Then if this is `True`, the error will be localized to exactly where it occured. If this is `False`, we still get the desired behaviour of not throwing. I'd be happy to take a PR on this. (I think it might need a bit of threading through to all the places that we use it.)

Tweaking this locally and setting `EQX_ON_ERROR=breakpoint`, `JAX_DISABLE_JIT=1`, then I can see that this error is thrown from a root find to locate the event, seemingly on the very first step. Which seems accurate:
```python
> event.cond_fn(0, _interpolator.evaluate(0))
Array(0.10000002, dtype=float32)
> event.cond_fn(1, _interpolator.evaluate(1))
Array(-0.47552538, dtype=float32)
```

Can you add some `jax.debug.breakpoint` (or some of the tools from `eqx.debug.*` inside the iteration to try debugging this?

patrick-kidger:
FWIW you've spurred me to try and improve things from the Equinox side, at least.

https://github.com/patrick-kidger/equinox/pull/785

This should now give much more clearly informative error messages, at least once the `error_if` is actually triggered.

johannahaffner:
> By the way, @dkweiss31 -- since I've not seen any movement on [jax-ml/jax#22011](https://github.com/jax-ml/jax/issues/22011), my guess is that we should probably work around this by registering a transpose rule for `lax.stop_gradient` ourselves. This transpose rule should just be the identity function.

The bug reported in https://github.com/jax-ml/jax/issues/22011 no longer crashes. Just perused this PR discussion looking for something related, and thought I'd note this :)

Issue #389, opened 2024-03-15
Complex input in diffeqsolve with PIDController


Ricky5389:
Hi, I encountered an issue while using the PIDController. When using complex input in diffeqsolve, I cannot change the coefficient of the PIDController. The default coefficients works fine but when I try to change I get an error leading me to believe that the time variable is converted to complex type somewhere. 
Here is a MWE to reproduce the error
```python
# %% ==========================================================================
# Imports
# =============================================================================
import jax.numpy as jnp
import diffrax as dx
from jaxtyping import  Scalar

# %% ==========================================================================
# Smallest working example, solving a complex ODE with diffrax
# ODE to solve is dy/dt = iy with y a complex number
# Setting up the ODE
# =============================================================================
def vector_field(t: Scalar, y: Scalar, *args):
    dotY = 1j * y
    return dotY

tsave = jnp.linspace(0.0, 10.0, 1000)
sim_time = tsave[-1]

solver = dx.Tsit5()
saveat = dx.SaveAt(ts=tsave)
y0 = 1.0j
# %% ==========================================================================
# Solve using the default PIDController coefficients
# =============================================================================
stepsize_controller = dx.PIDController(rtol=1e-6, atol=1e-6, pcoeff=0, icoeff=1, dcoeff=0.0)

term = dx.ODETerm(vector_field = vector_field)
res_dx = dx.diffeqsolve(term, solver, t0=0.0, t1=sim_time, dt0=0.01, y0=y0, saveat=saveat,
        stepsize_controller=stepsize_controller)

# %% ==========================================================================
# Solve using the non-zero P coefficient
# =============================================================================
stepsize_controller = dx.PIDController(rtol=1e-6, atol=1e-6, pcoeff=0.3, icoeff=0.3, dcoeff=0.0)

term = dx.ODETerm(vector_field = vector_field)
res_dx = dx.diffeqsolve(term, solver, t0=0.0, t1=sim_time, dt0=0.01, y0=y0, saveat=saveat,
        stepsize_controller=stepsize_controller)


## Changing the coefficient in the PID controller seems to raise the error
```
```
ValueError: `body_fun` must have the same input and output structure. Difference is:
  State(
    y=c128[],
    tprev=f64[],
-   tnext=f64[],
+   tnext=c128[],
    made_jump=bool[],
    solver_state=(bool[], c128[]),
-   controller_state=(bool[], bool[], f64[], c128[], c128[]),
[...]

PR #391, opened 2024-03-22
PID for complex dtype fixes

Randl:
Depends on https://github.com/patrick-kidger/lineax/pull/91
Should fix https://github.com/patrick-kidger/diffrax/issues/389

More generally, it appears that `caveat` is not tested at all for solvers. From what I can see, the problem in `Tsit5` is not caught by any test (though maybe the relevant test just doesn't run with complex dtypes?). Maybe having a large grid of tests for solver/parameter setups with simple DE can be helpful for these?

### DIFF
--- a/diffrax/_solver/tsit5.py
+++ b/diffrax/_solver/tsit5.py
@@ -1,6 +1,7 @@
 from collections.abc import Callable
 from typing import ClassVar, Optional
 
+import jax
 import jax.numpy as jnp
 import numpy as np
 from equinox.internal import ω
@@ -147,10 +148,11 @@ class _Tsit5Interpolation(AbstractLocalInterpolation):
         )
         b6 = -34.87065786149660974 * (t - 1.2) * (t - 0.666666666666666667) * t**2
         b7 = 2.5 * (t - 1) * (t - 0.6) * t**2
-        return (
-            self.y0**ω
-            + vector_tree_dot(jnp.stack([b1, b2, b3, b4, b5, b6, b7]), self.k) ** ω
-        ).ω
+        with jax.numpy_dtype_promotion("standard"):
+            return (
+                self.y0**ω
+                + vector_tree_dot(jnp.stack([b1, b2, b3, b4, b5, b6, b7]), self.k) ** ω
+            ).ω
 
 
 class Tsit5(AbstractERK):
--- a/test/test_integrate.py
+++ b/test/test_integrate.py
@@ -45,27 +45,25 @@ def _all_pairs(*args):
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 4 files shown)


Randl:
Pre-commit fail seems to be unrelated?

diffrax/_solver/tsit5.py
```diff
@@ -149,7 +149,10 @@ def evaluate(
         b7 = 2.5 * (t - 1) * (t - 0.6) * t**2
         return (
             self.y0**ω
-            + vector_tree_dot(jnp.stack([b1, b2, b3, b4, b5, b6, b7]), self.k) ** ω
+            + vector_tree_dot(
+                jnp.stack([b1, b2, b3, b4, b5, b6, b7]).astype(self.k.dtype), self.k
```
patrick-kidger, line 153: Do I understand correctly that this is just to satisfy no-dtype-promotion in tests?

patrick-kidger (commented):
This LGTM! I'll leave this unmerged until we do the next Lineax release (itself blocked on the next JAX release).

I think the pre-commit fail is due to an upstream JAX change: they've changed how `config` should be imported, I think.

As for why `Tsit5` wasn't already caught, I think the answer might just be that we have close-to-zero coverage for complex dtypes in Diffrax.

diffrax/_solver/tsit5.py
Randl: Yep

test/test_integrate.py
```diff
@@ -45,27 +45,25 @@ def _all_pairs(*args):
...
+)
+@pytest.mark.parametrize("t_dtype", (jnp.float32, int, float, jnp.int32))
+@pytest.mark.parametrize("y_dtype", (jnp.float32, jnp.complex64))
+@pytest.mark.parametrize("treedef", treedefs)
+@pytest.mark.parametrize(
+    "stepsize_controller",
+    (
+        diffrax.ConstantStepSize(),
+        diffrax.PIDController(rtol=1e-5, atol=1e-8),
+        diffrax.PIDController(rtol=1e-5, atol=1e-8, pcoeff=0.3, icoeff=0.3, dcoeff=0.0),
```
patrick-kidger, line 66: I'm a little concerned by this change. IIRC the `_all_pairs` thing was to try and reduce test times, which I think end up being very substantial for this many possible combinations. Do you know how much this changes things?

Btw, can you have this PR target `dev` instead please?

Other than that this LGTM!

Randl: It takes 6 minute instead of 4 in CI, out of total 24 before/26 after. So it's like 10% runtime, but otherwise complex types are almost untested (since one non-default parameter is dtype)

patrick-kidger:
Alright, LGTM! Thank you for the fixes as always.

PR #392, opened 2024-03-23
Enable more complex tests, fix related errors

Randl:
Similarly to https://github.com/patrick-kidger/lineax/pull/89 many of fixes are just type casts due to strict typing. Should consider which of these need to be casts and which aren't

Also I included changes from https://github.com/patrick-kidger/diffrax/pull/391 for simplicity

### DIFF
--- a/diffrax/_brownian/tree.py
+++ b/diffrax/_brownian/tree.py
@@ -91,7 +91,7 @@ def _levy_diff(_, x0: tuple, x1: tuple) -> AbstractBrownianIncrement:
         assert len(x1) == 2
         dt0, w0 = x0
         dt1, w1 = x1
-        su = jnp.asarray(dt1 - dt0, dtype=w0.dtype)
+        su = jnp.asarray(dt1 - dt0, dtype=complex_to_real_dtype(w0.dtype))
         return BrownianIncrement(dt=su, W=w1 - w0)
 
     elif len(x0) == 4:  # space-time levy area case
@@ -100,12 +100,13 @@ def _levy_diff(_, x0: tuple, x1: tuple) -> AbstractBrownianIncrement:
         dt1, w1, hh1, bhh1 = x1
 
         w_su = w1 - w0
-        su = jnp.asarray(dt1 - dt0, dtype=w0.dtype)
+        su = jnp.asarray(dt1 - dt0, dtype=complex_to_real_dtype(w0.dtype))
         _su = jnp.where(jnp.abs(su) < jnp.finfo(su).eps, jnp.inf, su)
         inverse_su = 1 / _su
-        u_bb_s = dt1 * w0 - dt0 * w1
-        bhh_su = bhh1 - bhh0 - 0.5 * u_bb_s  # bhh_su = H_{s,u} * (u-s)
-        hh_su = inverse_su * bhh_su
+        with jax.numpy_dtype_promotion("standard"):
+            u_bb_s = dt1 * w0 - dt0 * w1
+            bhh_su = bhh1 - bhh0 - 0.5 * u_bb_s  # bhh_su = H_{s,u} * (u-s)
+            hh_su = inverse_su * bhh_su
         return SpaceTimeLevyArea(dt=su, W=w_su, H=hh_su)
     else:
         assert False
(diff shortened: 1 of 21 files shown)


diffrax/_brownian/tree.py
```diff
@@ -52,11 +52,9 @@
 # For more about space-time Levy area see Definition 4.2.1.
 # For the midpoint rule for generating space-time Levy area see Theorem 6.1.6.
 # For the general interpolation rule for space-time Levy area see Theorem 6.1.4.
-
-FloatDouble: TypeAlias = tuple[Float[Array, " *shape"], Float[Array, " *shape"]]
-FloatTriple: TypeAlias = tuple[
-    Float[Array, " *shape"], Float[Array, " *shape"], Float[Array, " *shape"]
-]
+FloatOrComplex = Union[Float[Array, " *shape"], Complex[Array, " *shape"]]
```
patrick-kidger, line 55: I think you can use `jaxtyping.Inexact` here.

diffrax/_brownian/tree.py
```diff
@@ -229,6 +227,7 @@ def _is_levy_val(obj):
         t0 = eqxi.nondifferentiable(t0, name="t0")
         # map the interval [self.t0, self.t1] onto [0,1]
         t0 = linear_rescale(self.t0, t0, self.t1)
+        print(t0)
```
patrick-kidger, line 230: .

diffrax/_global_interpolation.py
```diff
@@ -134,13 +134,22 @@ def evaluate(
         def _index(_ys):
             return _ys[index]
 
+        y_leaves = jtu.tree_leaves(self.ys)
+        if len(y_leaves) == 0:
+            ys_dtype = jnp.float64
```
patrick-kidger, line 139: I think we have a `default_floating_point` function floating around somewhere.

test/test_integrate.py
```diff
@@ -146,13 +144,16 @@ def f(t, y, args):
...
+def test_ode_order(solver, dtype):
     solver = implicit_tol(solver)
     key = jr.PRNGKey(5678)
     akey, ykey = jr.split(key, 2)
 
-    A = jr.normal(akey, (10, 10), dtype=jnp.float64) * 0.5
+    A = jr.normal(akey, (10, 10), dtype=dtype) * 0.5
 
+    if jnp.iscomplexobj(A) and isinstance(solver, diffrax.AbstractImplicitSolver):
+        return
```
patrick-kidger, line 156: What's the reason for this exclusion? I think ideally we do expect to be able to do this (once all the relevant upstream things have landed).

test/test_integrate.py
```diff
@@ -212,16 +213,24 @@ def _solvers():
...
+        return 0.25 * diffusion_mlp(y).reshape(3, noise_dim)
 
 
 @pytest.mark.parametrize("solver_ctr,commutative,theoretical_order", _solvers())
-def test_sde_strong_order(solver_ctr, commutative, theoretical_order):
+@pytest.mark.parametrize(
+    "dtype",
+    [
+        jnp.float64,
+    ],
```
patrick-kidger, line 231: Remove the comma at the end of both these lines and it should autoformat better.

test/test_integrate.py
Randl: Currently,
https://github.com/patrick-kidger/diffrax/blob/main/diffrax/_integrate.py#L638-L642
I agree that we will want to remove it in the future

Randl: It used to work (somehow) when we merged the basic support, but now I think strict typing broke it in multiple places.

Randl: I've added some related fixes (also see https://github.com/patrick-kidger/optimistix/pull/53), but I think enabling it should come in a separate PR

diffrax/_global_interpolation.py
```diff
@@ -8,6 +8,7 @@
 import jax.lax as lax
 import jax.numpy as jnp
 import jax.tree_util as jtu
+from lineax._norm import default_floating_dtype
```
patrick-kidger, line 11: Prefer `lineax.internal` for this.

diffrax/_global_interpolation.py
```diff
@@ -134,13 +135,22 @@ def evaluate(
         def _index(_ys):
             return _ys[index]
 
+        y_leaves = jtu.tree_leaves(self.ys)
+        if len(y_leaves) == 0:
+            ys_dtype = default_floating_dtype()
+        else:
+            ys_dtype = jnp.result_type(*y_leaves)
```
patrick-kidger, line 142: I think this is wrong: this will result in the entire operation below getting cast to the same largest dtype.
Rather, the casting should happen on a per-leaf basis, so that we do the minimal amount of upcasting necessary.

(If this is just to avoid the strict-dtype-casting rules, then I'm okay with locally disabling that using `with jax....` as discussed elsewhere.)

patrick-kidger: (Actually, I'm realising that means we probably don't need `default_floating_dtype` at all, as no-leaves means that we have a no-op.)

diffrax/_root_finder/_verychord.py
```diff
@@ -94,11 +94,12 @@ def init(
                 y_dtype = lxi.default_floating_dtype()  # pyright: ignore
             else:
                 y_dtype = jnp.result_type(*y_leaves)
+            diff_dtype = jnp.finfo(y_dtype).dtype
```
patrick-kidger, line 97: I think if this trick is going to be appearing in a few places then it should probably be pulled out into a `complex_to_real_dtype` function in `_misc.py`. This will improve readability by giving a name to this operation.

test/test_integrate.py
patrick-kidger: Okay, sounds reasonable!

diffrax/_solver/implicit_euler.py
```diff
@@ -81,14 +82,15 @@ def step(
...
-        nonlinear_sol = optx.root_find(
-            _implicit_relation,
-            self.root_finder,
-            k0,
-            args,
-            throw=False,
-            max_steps=self.root_find_max_steps,
-        )
+        with jax.numpy_dtype_promotion("standard"):
+            nonlinear_sol = optx.root_find(
```
Randl, line 86: I think this one and similar won't be needed after https://github.com/patrick-kidger/optimistix/pull/53
I'll check it later

patrick-kidger: Okay, lmk when you want a review of this!

Can you rebase on top of the latest `dev`? (We just merged a large change introducing stochastic Runge--Kutta methods. As this PR is small I'll try to get it in before the next large change... :D )

test/test_integrate.py
```diff
@@ -223,16 +226,19 @@ def _squareplus(x):
...
 def _diffusion(t, y, args):
     _, diffusion_mlp, noise_dim = args
-    return 0.25 * diffusion_mlp(y).reshape(3, noise_dim)
+    with jax.numpy_dtype_promotion("standard"):
+        return 0.25 * diffusion_mlp(y).reshape(3, noise_dim)
 
 
 @pytest.mark.parametrize("solver_ctr,noise,theoretical_order", _solvers_and_orders())
-def test_sde_strong_order(solver_ctr, noise, theoretical_order):
+@pytest.mark.parametrize("dtype", [jnp.float64])
```
patrick-kidger, line 240: Is this intentionally leaving off complex support?

test/test_sde.py
```diff
@@ -43,8 +43,12 @@ def _solvers_and_orders():
 # converges to its own limit (i.e. using itself as reference), and then in a
 # different test check whether that limit is the same as the Euler/Heun limit.
 @pytest.mark.parametrize("solver_ctr,noise,theoretical_order", _solvers_and_orders())
+@pytest.mark.parametrize(
+    "dtype",
+    (jnp.float64,),
+)
```
patrick-kidger, line 49: Likewise here.

test/test_solver.py
```diff
@@ -274,7 +274,8 @@ def order(self, terms):
 
 
 # Essentially used as a check that our general IMEX implementation is correct.
-def test_sil3():
+@pytest.mark.parametrize("dtype", (jnp.float64,))
```
patrick-kidger, line 277: And here!

patrick-kidger (commented):
Okay, this LGTM! Let me know whether the points I've flagged up are intentional or not. If they are then I'm happy to merge this as-is.

Randl:
Yep all three are intentional since there are still failing tests and i prefer to fix them separately, keeping this pr only to type errors and test modifications. I'll open a separate PR for these

patrick-kidger:
Okay then! Great stuff, I've just merged this.

Issue #399, opened 2024-04-14
Wrong annotations for DenseInfos


Randl:
`DenseInfos` are defined of shape `times`
https://github.com/patrick-kidger/diffrax/blob/38a40e7fba6cffcb7e8f65a1ab3a6a9ababacaac/diffrax/_custom_types.py#L52

While in fact it probably should be `times+1`:
https://github.com/patrick-kidger/diffrax/blob/38a40e7fba6cffcb7e8f65a1ab3a6a9ababacaac/diffrax/_global_interpolation.py#L322
https://github.com/patrick-kidger/diffrax/blob/38a40e7fba6cffcb7e8f65a1ab3a6a9ababacaac/diffrax/_integrate.py#L883-L887

As such global interpolation tests fail, e.g., see https://github.com/patrick-kidger/diffrax/pull/391 but also locally on master

It is possibly related to some update, maybe of jaxtyping (this PR is on the new Lineax version).

PR #400, opened 2024-04-16
Fix DenseInfo type

Randl:
Fix https://github.com/patrick-kidger/diffrax/issues/399


patrick-kidger:
Hmm, this is mysterious. How did this ever pass?

Thanks for the fix!

Issue #401, opened 2024-04-17
Best way to approach KL divergence


lockwo:
I was wondering what the best way of recreating the KL divergence from `torchsde` would be in diffrax. I have a hacked together version of https://github.com/google-research/torchsde/blob/master/examples/latent_sde.py, but it seems like there might be a more general approach. I didn't see any issues, but I see an older stale PR (https://github.com/patrick-kidger/diffrax/pull/104), might be able to revive that if this would be of interest to the package

patrick-kidger:
I think it should still be possible in a similar way to #104! I'm undecided on whether this makes sense to include as a function in Diffrax itself, but if you put some functionality together then I'd be interested to see it.

I think probably the only thing I'd remark is that we now have Lineax, and might be able to use a cleaner + more structured approach to solving the linear system -- using Lineax to exploiti the diagonal structure if present etc.

lockwo:
Definitely should be possible to clean it up a lot, but any preliminary thoughts on this approach would be much appreciated: https://github.com/patrick-kidger/diffrax/pull/402

Issue #370, opened 2024-02-10
Support Lineax operators for the vector field of ControlTerm


tttc3:
By providing support for `lineax.AbstractLinearOperators` in the vector field of a ControlTerm it may be possible to reduce the need for  `WeaklyDiagonalControlTerm` and/or any other such specialised control terms.

The following gives a MWE:
```python
class ControlTerm(_ControlTerm):
    def prod(self, vf, control):
        if isinstance(vf, lx.AbstractLinearOperator):
            return jtu.tree_map(lambda _vf, _control: _vf.mv(_control), vf, control)
        return jtu.tree_map(_prod, vf, control)

# These two are now equivalent.
ControlTerm(lx.DiagonalLinearOperator(jnp.array([1,2,3])), ...)
WeaklyDiagonalControlTerm(jnp.array([1,2,3]), ...)
```
Not sure if this is something you want to support, but it occurred to me that the operator tags might also be useful for some of the diffrax solvers?

patrick-kidger:
Yup, with the release of Lineax this is something I've been considering! This also dovetails nicely with #364, so that we may wish to also introduce a type parameter for the return type of the vector field.

(One very nitty concern I do have is that mathematically speaking, we tend to interpret `f(y) dx` as actually being a linear function `dx -> f( . ) dx` (returning a nonlinear function), rather than a nonlinear function `y -> f(y)` (returning a linear operator). But that's probably not a super important distinction, to be honest.)

I'd want to be sure that this works correctly with:
- solvers that evaluate and store the result of `.vf(...)` directly.
- `BacksolveAdjoint`
- returning "complicated" linear operators like `lx.JacobianLinearOperator`.

In principle all of those things should be solvable -- Diffrax allows the result of the vector field to be arbitrary -- I just think we'd want to explicitly test them.

I'd be happy to take a pull request on this!

lockwo:
I have a branch started for this, but I want to know the scope of the change you are looking for here. Having spent (a little) time thinking about it I see a couple options, in increasing impact on the package
- I can add support to ControlTerms to allow linear operators, which would really reduce the complexity of https://github.com/patrick-kidger/diffrax/pull/402, without changing much or adding many LoC (and introduce no breaking changes). This could be a weird edge case, like oh you can do weaklydiagonal or also just lx control (maybe not a big deal)
[...]

PR #402, opened 2024-04-17
Add KL divergence terms for Latent SDEs

lockwo:
Addresses https://github.com/patrick-kidger/diffrax/issues/401. Revives https://github.com/patrick-kidger/diffrax/pull/104. Based on that PR, I made the minimal requirements to get it up to current version (e.g. taking callables instead of ODE terms since we can't make these .vf becuase _broadcast_and_upcast requires that aug_y and drift(aug_y) are the same shape, but they aren't).


lockwo:
Before going further (there is a lot I am going to improve/polish) I wanted to check with your thoughts on the general approach of KL being terms and exposing the user to a function that converts their problem. An alternative could be something like in torchsde where it's part of the intregration method, i.e. the user flags it at integration time.

lockwo:
The other big question is if I could trivially expand this to pytree's rather than just arrays

diffrax/_kl_term.py
```diff
@@ -0,0 +1,175 @@
...
+    """General case"""
+    scale = jnp.linalg.pinv(diffusion) @ drift
+    return 0.5 * jnp.sum(scale**2)
+
+def _handle(drift: Array, diffusion: Array):
+    """According to the shape of drift and diffusion,
+    select the right way to compute KL divergence
+    """
+    eqx.error_if(drift, not eqx.is_array(drift), "Only array drifts are supported")
+    eqx.error_if(diffusion, not eqx.is_array(diffusion), "Only array diffusion are supported")
```
patrick-kidger, line 41: Note that you need to write `x = eqx.error_if(x, ...)`; otherwise these error checks get DCE'd.

patrick-kidger (commented):
On make these terms: I don't have super strong feelings, but compared to the original PR we have now more clearly defined what a term is in Diffrax, and I think there are other points on the design space.

To be precise: given a diffeq of the form
```
dy = f(y, z) da + g(y, z) db
dz = h(y, z) da + k(y, z) db
```
then this would be represented in Diffrax as
```
terms = (
    MultiTerm(f, g),
    MultiTerm(h, k),
)
```

In general: everything inside a `MultiTerm(...)` is all applied to the same `dfoo`. For example the SDE-specific solvers consume a `MultiTerm[ODETerm, AbstractTerm]`, for the drift and diffusion.
Meanwhile the PyTree structure of terms themselves corresponds to different `dfoo` and `dbar`. For example semi-implicit Euler takes a pair of `(AbstractTerm, AbstractTerm)`, corresponding to the two components that are being evolved.

In this case there is an argument that the extra KL-divergence term should really correspond to a new `dfoo`, and that as such the correct thing to do is to instead replace `terms` with `(terms, kl_term)`, and then provide a wrapper solver which understands this alternative term structure.

patrick-kidger:
On the topic of Lineax: indeed, this should definitely make handling PyTrees much easier.

lockwo:
I think your idea makes a lot of sense, and I made a fair amount of progress on the solver wrapper approach. I think it will require a few more changes outside the new functions (had to make some changes to subsave to log the KLs at each step) and it requires 2 redundant VF calls (unless there is a way to get the VF calls back from an arbitrary solver step?).

However, the problem I encountered is fundamentally that the problem is not really a well formed diffrax ODE. What I mean by that is the drift function in Latent SDEs takes concat(y, context) and maps it to a shape of y. However, an ODETerm requires input to vf and output of vf are the same shape (or at least broadcast able), per https://github.com/patrick-kidger/diffrax/blob/main/diffrax/_term.py#L186. However, we might map (68,) -> (4,) as a simple example of a drift term. This was able to be sidestepped with the term approach (because we directly access the function rather than the `.vf`) but for the solver it doesn't work. My method to get this to work with the solver wrapper was to either change ODETerm to remove this requirement (which seems bad) or modify the args so the solver.set gets passed args=(args, context), then the user has to define their vector field function which does the concatenation of y and args[1] which works but seems oddly specific and bespoke. Wdyt?

lockwo:
Ok, I added a solver wrapper (the term version is still up). With the solver wrapper comes a needed way of logging the KLs, the best way I thought of was just to add it to the SaveAt (not supporting ts currently). It's a bit specific, but maybe there are plans in the future to allow arbitrary saving of solver states? Regardless, I still have to test it, but it should (mostly) work. Not just if hijacking to solver state is ideal, but it makes sense to me as a wrapper solver. Let me know if this is what you had in mind when describing a wrapper and I will formalize this into a real PR.

lockwo:
Ok, I polished things up. I went with a sort of hybrid approach where the users specifies the SDEs as you described, then just wraps a solver and everything works smoothly. However, I did create internal terms, in order to get an arbitrary solver to integrate through the KL computation, that was the best way I could think of to do so, but they are completely hidden from the user. I also added the example (can be modified to add more text, or remove pmap although I do like having an example with distribution especially since its painfully slow without it) and a test and updated the docs. Taking it off draft now since its a real PR.

lockwo:
Looking at the tests while they are running, I see a fail at 

```
diffrax/_integrate.py:776: in diffeqsolve
    if not _term_compatible(y0, args, terms, solver.term_structure):
```

which isn't aline of code I can even find in diffrax main or in my PR?

All the tests pass locally, so if they fail on GitHub I probably just have to modify some minor thing to appease the action runner

frankschae:
This is a very cool feature/example! It looks like one needs to specify
```python
levy_area=diffrax.BrownianIncrement
```
in `diffrax.UnsafeBrownianPath`

lockwo:
Thanks @frankschae , good catch!

lockwo:
The test failures are all just the safe map 0.4.27 stuff

diffrax/_integrate.py
```diff
@@ -803,6 +804,9 @@ def _promote(yi):
                 "`UnsafeBrownianPath` cannot be used with adaptive step sizes."
             )
 
+    if isinstance(solver, KLSolver):
+        y0 = (y0, 0.0)
+        y0 = jtu.tree_map(_promote, y0)
     # Normalises time: if t0 > t1 then flip things around.
```
patrick-kidger, line 810: I don't think this is a thing we should put in `_integrate.py`. Rather, we should have some kind of
```python
def kl_foo_bar(...):
    ...
```
function, which can be used as
```python
solver, y0 = kl_foo_bar(...)
diffeqsolve(solver=solver, y0,=y0 ...)
```

diffrax/_solver/kl.py
```diff
@@ -0,0 +1,299 @@
+import operator
+from typing import Optional, Tuple, Union
```
patrick-kidger, line 2: As mentioned elsewhere, use `tuple` instead of `Tuple`!

diffrax/_solver/kl.py
```diff
@@ -0,0 +1,299 @@
...
+    WeaklyDiagonalControlTerm,
+)
+from .base import (
+    _SolverState,
+    AbstractSolver,
+    AbstractWrappedSolver,
+)
+
+
+_DiffusionTerm = Union[ControlTerm, WeaklyDiagonalControlTerm]
```
patrick-kidger, line 36: I don't think we want to specialise on specifically a weakly-diagonal kind of control.

Rather, I think any kind of control should be admissible! Then we use Lineax to do the linear solve against whatever kind of operator is returned.

You know, maybe what we should really be doing is implementing #370! That is to say, allow the `vector_field` of `diffrax.ControlTerm(vector_field, ...)` to instead return a `lx.AbstractLinearOperator`. That would then make things exceptionally easy over here.

WDYT?

(I'll hold off on doing the rest of the review yet as I can see this may change things.)

patrick-kidger (commented):
Gosh, it's taken me a long time to get around to this 😅

Thank you for your patience!

As you can see, I think there's actually a way we can make this much much simpler. :)

diffrax/_solver/kl.py
lockwo: I think that makes sense, that could reduce the parsing/if elsing complexity I have by a fair amount. I will make a PR for #370 then incorporate those changes into this and we can revisit this PR at that point.

diffrax/_integrate.py
lockwo: Would you want y0 out of `kl_foo_bar` to be specific `KL y0` class or would a tuple be fine? Also, would we want to modify the y0 and the solver at the same time or have a KL solver and a KL y0 transformer?

PR #405, opened 2024-04-20
Create py.typed

lockwo:
Without it, an error appears every time typing is checked

```
Skipping analyzing "diffrax": module is installed, but missing library stubs or py.typed marker  [import]
```

Same as https://github.com/patrick-kidger/equinox/pull/547


patrick-kidger:
Ah, makes sense! We should probably add this to Lineax and Optimistix too.
It looks like your branch has some spurious commits on it, but I'd be happy to click accept if you can remove those?

lockwo:
Oops wrong branch, here it is https://github.com/patrick-kidger/diffrax/pull/408

PR #409, opened 2024-04-27
Doc tweaks for SRK.

patrick-kidger:
This PR is just fixing up some docs issues I spotted whilst reviewing #344.

Whilst I'm here I also flipped some branches that looked to have a superfluous `not` attached to them.

@andyElking  WDYT?

### DIFF
--- a/diffrax/_solver/srk.py
+++ b/diffrax/_solver/srk.py
@@ -165,7 +165,7 @@ class StochasticButcherTableau(Generic[_Coeffs]):
 StochasticButcherTableau.__init__.__doc__ = """The coefficients of a
 [`diffrax.AbstractSRK`][] method.
 
-See the documentation for [`diffrax.AbstractSRK`][] for additional details on the
+See also the documentation for [`diffrax.AbstractSRK`][] for additional details on the
 mathematical meaning of each of these arguments.
 
 **Arguments:**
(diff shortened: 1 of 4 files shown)


diffrax/_solver/srk.py
```diff
@@ -198,8 +202,9 @@ class AbstractSRK(AbstractSolver[_SolverState]):
     r"""A general Stochastic Runge-Kutta method.
 
     This accepts `terms` of the form
-    `MultiTerm(ODETerm(drift), ControlTerm(diffusion, brownian_motion))`.
-     Depending on the solver, the Brownian motion might need to generate
+    `MultiTerm(ODETerm(drift), ControlTerm(diffusion, brownian_motion))` or
+    `MultiTerm(ODETerm(drift), WeaklyDiagonalControlTerm(diffusion, brownian_motion))`.
```
andyElking, line 206: I think it is in fact more general and accepts anything of the form `MultiTerm[tuple[ODETerm, AbstractTerm[Any, self.minimal_levy_area]]]`. This will be relevant for underdamped Langevin where a custom diffusion term is used, because the BM term only appears in dv, not in dx (and a WeaklyDiagonalTerm would be inefficient). But I agree if you want to keep it as you wrote it here, to avoid confusing users.

-- Nevermind, I see you adressed this later in `sde_solvers.md`.

diffrax/_solver/srk.py
```diff
@@ -514,11 +520,10 @@ def compute_and_insert_kf_j(_h_kfs_in):
             if ignore_stage_f is None:
                 _h_kfs = compute_and_insert_kf_j(_h_kfs)
             else:
-                drift_pred = jnp.logical_not(ignore_stage_f[j])
                 _h_kfs = lax.cond(
-                    eqxi.nonbatchable(drift_pred),
-                    compute_and_insert_kf_j,
+                    eqxi.nonbatchable(ignore_stage_f[j]),
```
andyElking, line 524: I think I intentionally had a superfluous not here, because I thought it made things clearer. But I agree with the change, this is better overall.

docs/api/solvers/sde_solvers.md
```diff
@@ -14,9 +16,9 @@ See also [How to choose a solver](../../usage/how-to-choose-a-solver.md#stochast
     diffeqsolve(terms, solver=Euler(), ...)
     ```
 
-    Some solvers are SDE-specific. For these, such as for example [`diffrax.StratonovichMilstein`][], then `terms` must specifically be of the form `MultiTerm(ODETerm(...), SomeOtherTerm(...))` (Typically `SomeOTherTerm` will be a `ControlTerm` or `WeaklyDiagonalControlTerm`) representing the drift and diffusion specifically.
+    For any individual solver then this is documented below, and is also available programatically under `<solver>.term_structure`.
 
-    For those SDE-specific solvers then this is documented below, and the term structure is available programmatically under `<solver>.term_structure`.
+    For advanced users, note that we typically accept any `AbstractTerm` for the diffusion, so it could be a custom one that implements more-efficient behaviour for the structure of your diffusion matrix. (Much like how [`diffrax.WeaklyDiagonalControlTerm`][] is more efficient than [`diffrax.ControlTerm`][] for diagonal diffusions.)
```
andyElking, line 21: Oh, I see, you adressed this here. You can then ignore my comment above.

andyElking (commented):
Looks good, I agree with all the changes, thanks! I left a few comments, but none of them require further changes.

patrick-kidger:
Wonderful stuff. Ty for the review, merged!

PR #410, opened 2024-04-28
fixed brownian_tree_times.py

andyElking:
We forgot to update the VBT timing benchmark when we changed Brownian return types, so here's a quick fix.


patrick-kidger:
LGTM! Thank you for the fix.

PR #414, opened 2024-05-08
STTLA PR

andyElking:
This PR adds space-time-time Levy area to `VirtualBrownianTree` and `UnsafeBrownianPath`. The changes to the overall logic are minor, I mostly just added the math for the space-time-time case.

The additions to `test_brownian.py` are more extensive:

- In `test_conditional_statistics` we test the pvals of the marginals for W, H, and K separately.  This confirms that they are all Gaussian, so their joint distribution is exactly determined by their means and covariances, which we also test for. All of these are conditioned on the sample just before and the one just after the current sample (as it was before). I moved the code for computing the statistics into the `conditional_statistics` function.
- Since the code of `test_conditional_statistics` itself is based on thm 3.7 from the Single-seed paper, I chose to test its correctness from a different angle in `test_whk_interpolation`. This test tries three things: a) just final interpolation with no halving (brownian bridge) steps, b) only one halving step followed by a final interpolation, c) 15 halving steps (i.e. `tol = 2**-15`) and no final interpolation (aka. "zero" spline). Since they all have the same distribution as predicted by thm 3.7, I conclude that indeed everything is as it should be. This still relies on the assumption that the halving steps are correct (i.e. thm 3.5), but that was computed by James in his thesis, so I suppose that's a fair assumption :).

In the third commit I added shape annotations to arrays in `AbstractBrownianIncrement` and its subclasses. This is optional, so if you prefer we can remove this commit.

In the fourth commit I just removed two unnecessary `# pyright: ignore` comments, so pyright doesn't complain. These have nothing to do with the rest of this PR.

I hope this PR is less challenging to review. No hurry though :)


patrick-kidger:
Okay, happy to start thinking about reviewing this PR now! If you can rebase on top of `main` then I'll take a look :)

andyElking:
That's great news! I rebased it on top of main and changed the base branch :)

diffrax/_autocitation.py
```diff
@@ -359,17 +360,34 @@ def _virtual_brownian_tree(terms):
...
+}
+
+% and Theorem 6.1.6 of
+@phdthesis{foster2020a,
+  publisher = {University of Oxford},
+  school = {University of Oxford},
+  title = {Numerical approximations for stochastic differential equations},
+  author = {Foster, James M.},
+  year = {2020}
+}
```
patrick-kidger, line 389: I think these citations deserve to also go `VirtualBrownianTree.__doc__`! :) You did a lot of work making this happen!

test/test_brownian.py
```diff
@@ -13,6 +13,11 @@
 import scipy.stats as stats
 
 
+levy_areas = (
```
patrick-kidger, line 16: Nit: should start with an underscore, it's private to this file.

test/test_brownian.py
```diff
@@ -31,56 +36,37 @@ def _make_struct(shape, dtype):
...
-        {"a": jnp.float64, "b": jnp.complex128},
-        (jnp.float16, (jnp.float32, jnp.float64)),
-    )
+    if (
+        ctr is diffrax.VirtualBrownianTree
+        and levy_area is diffrax.SpaceTimeTimeLevyArea
+    ):
+        # VBT with STTLA does not support float16 or complex dtypes
+        # because it uses jax.random.multivariate_normal
+        shapes_dtypes = shapes_dtypes[:6]
```
patrick-kidger, line 64: This seems a little error prone. Can this be switched into two groups `shapes_dtypes1` and `shapes_dtypes2`, which are then combined or not appropriately?

diffrax/_brownian/tree.py
```diff
@@ -63,62 +64,113 @@
...
+    bar_H: Array
+
+
+class _SpaceTimeTimeLevyVal(_AbstractSpaceTimeLevyVal):
+    dt: Inexact[Array, ""]
+    W: Array
+    H: Array
+    bar_H: Array
+    K: Array
+    bar_K: Array
```
patrick-kidger, line 98: Possibly slightly simpler might be something like:
```python
class _LevyVal(eqx.Module):
    brownian_increment: AbstractBrownianIncrement
    bar_H: Optional[Array]
    bar_K: Optional[Array]
```
?
Possibly adding some `Generic[...]` typevars for each of the three arguments if required.

diffrax/_brownian/tree.py
```diff
@@ -63,62 +64,113 @@
...
+
+
 class _State(eqx.Module):
     level: IntScalarLike  # level of the tree
-    s: RealScalarLike  # starting time of the interval
+    s: Inexact[Array, ""]  # starting time of the interval
```
patrick-kidger, line 103: Times should always be real, not complex.

patrick-kidger (commented):
Awesome stuff! I've just quickly looked over this. I'll let you address my first round of comments, but my initial impression is that this is a fairly small change, so it should be pretty easy to land.

diffrax/_brownian/tree.py
andyElking: Fair enough, I suppose I was trying to do it very properly, but since it's never exposed to the user, you're right, I should go for the simpler option.

diffrax/_brownian/tree.py
andyElking: I was trying to avoid a myriad of `with jax.numpy_dtype_promotion("standard"):`, which spaghettify the code a bit, and I thought the times which are only used internally could safely be complex. But I guess it could be a source of bugs, so fair enough.

test/test_brownian.py
andyElking: Yes, good point.

diffrax/_autocitation.py
andyElking: Do you mean in the autocitation for the VBT? It is already in the doc of VBT, but I did add it to autocitation as well. I also added a little subsection on Levy areas in `docs/api/brownian/md`.

andyElking:
Thanks for the quick review! I made the changes you suggested.

I also added a little subsection on Levy areas in `docs/api/brownian/md`. Mainly to clear up how `solver.minimal_levy_area` interacts with VBT, etc.

diffrax/_autocitation.py
patrick-kidger: What I mean is that I think these citations should appear in the docstring, and be parsed out with `_parse_reference_multi`.

diffrax/_brownian/path.py
```diff
@@ -138,15 +141,24 @@ def _evaluate_leaf(
...
         key,
         shape: jax.ShapeDtypeStruct,
-        levy_area: type[Union[BrownianIncrement, SpaceTimeLevyArea]],
+        levy_area: type[
+            Union[BrownianIncrement, SpaceTimeLevyArea, SpaceTimeTimeLevyArea]
+        ],
         use_levy: bool,
     ):
         w_std = jnp.sqrt(t1 - t0).astype(shape.dtype)
+        key_w, key_hh, key_kk = jr.split(key, 3)
```
patrick-kidger, line 150: Can we do the split inside the `if` blocks, to avoid the overhead in the common case of no Levy area?

diffrax/_brownian/tree.py
```diff
@@ -63,6 +64,25 @@
...
+# taking the difference between two Levy areas.
+class _LevyVal(eqx.Module):
+    dt: RealScalarLike
+    W: Array
+    H: Optional[Array]
+    bar_H: Optional[Array]
+    K: Optional[Array]
+    bar_K: Optional[Array]
+
+    def __post_init__(self):
```
patrick-kidger, line 78: Nit: prefer `__check_init__` where possible. This is an Equinox-specific extension that (a) runs even when you inherit, and (b) isn't ignored if you define a custom `__init__` method, and (c) doesn't allow you to still mutate the class whilst it is running.

diffrax/_brownian/tree.py
```diff
@@ -72,55 +92,72 @@ class _State(eqx.Module):
...
     r"""Computes $(W_{s,u}, H_{s,u})$ from $(W_s, \bar{H}_{s,u})$ and
     $(W_u, \bar{H}_u)$, where $\bar{H}_u = u * H_u$.
 
     **Arguments:**
 
     - `_`: unused, for the purposes of aligning the `jtu.tree_map`.
-    - `x0`: `LevyVal` at time `s`.
-    - `x1`: `LevyVal` at time `u`.
+    - `x0`: `_AbstractLevyVal` at time `s`.
+    - `x1`: `_AbstractLevyVal` at time `u`.
```
patrick-kidger, line 103: Remove the `Abstract`.

diffrax/_brownian/tree.py
```diff
@@ -72,55 +92,72 @@ class _State(eqx.Module):
...
 
     **Returns:**
 
-    `LevyVal(W_su, H_su)`
+    `AbstractBrownianIncrement(W_su, H_su, K_su)`
     """
+    dtype = jnp.dtype(x0.W)
```
patrick-kidger, line 109: I'd prefer using `jnp.result_type` as the argument is an array. I forget exactly what goes wrong but I recall getting bit by `jnp.dtype` at some point before.

diffrax/_brownian/tree.py
```diff
@@ -72,55 +92,72 @@ class _State(eqx.Module):
...
-            u_bb_s = dt1 * w0 - dt0 * w1
-            bhh_su = bhh1 - bhh0 - 0.5 * u_bb_s  # bhh_su = H_{s,u} * (u-s)
-            hh_su = inverse_su * bhh_su
-        return SpaceTimeLevyArea(dt=su, W=w_su, H=hh_su)
+            bkk_su = (
+                x1.bar_K
+                - x0.bar_K
+                - su / 2 * x0.bar_H
+                + x0.dt / 2 * bhh_su
+                - (x1.dt - 2 * x0.dt) / 12 * u_bb_s
```
patrick-kidger, line 142: Are these meant to be e.g. `su / (2 * x0.bar_H)` or `(su / 2) * x0.bar_H`?

I usually prefer to add brackets to communicate intent, even if the language has the appropriate binding precedences to make them unnecessary.

diffrax/_brownian/tree.py
```diff
@@ -138,18 +175,13 @@ class VirtualBrownianTree(AbstractBrownianPath):
...
-        randomness required for certain stochastic Runge--Kutta solvers; see
-        [`diffrax.AbstractSRK`][] for more information.
-
-        An error will be thrown during tracing if Lévy area is required but is not
-        available.
-
-        The choice here will impact the Brownian path, so even with the same key, the
-        trajectory will be different depending on the value of `levy_area`.
+        The parameter `levy_area` can be set to one of:
+        - [`diffrax.BrownianIncrement`][] (default, generates the increment of W)
```
patrick-kidger, line 179: Nit: markdown requires a newline before starting a bulleted list.

diffrax/_brownian/tree.py
```diff
@@ -138,18 +175,13 @@ class VirtualBrownianTree(AbstractBrownianPath):
...
+        - [`diffrax.SpaceTimeLevyArea`][] (generates W and the space-time Levy area H)
+        - [`diffrax.SpaceTimeTimeLevyArea`][] (generates W, H and the space-time-time
+                                                Levy area K)
+        The choice of `levy_area` will impact the Brownian path, so even with the same
```
patrick-kidger, line 183: And likewise a newline afterwards.

diffrax/_brownian/tree.py
```diff
@@ -348,14 +401,21 @@ def _body_fun(_state: _State):
...
+                self.levy_area is SpaceTimeLevyArea
+                or self.levy_area is SpaceTimeTimeLevyArea
+            ):
                 assert _bhh_stu is not None and _bhh_st_tu is not None
                 _bhh = _split_interval(_cond, _bhh_stu, _bhh_st_tu)
-            elif self.levy_area is BrownianIncrement:
-                _bhh = None
             else:
-                assert False
+                _bhh = None
```
patrick-kidger, line 418: I wonder if this can be more neatly expressed via a `jtu.tree_map(_split_interval, ...)`?

Not a strong feeling though.

diffrax/_brownian/tree.py
```diff
@@ -369,14 +429,114 @@ def _body_fun(_state: _State):
...
+                ) / su3
+                wh_cov = -(sr_by_su_3 * ru_by_su * (sr2 - 3 * sr * ru + 6 * ru2)) / (
+                    2 * su
+                )
+                wk_cov = (sr_by_su**4) * ru_by_su * (sr - ru) / 12
+                hh_cov = (
+                    sr / 12 * (1 - sr_by_su_3 * (sr2 + 2 * sr * ru + 16 * ru2) / su2)
+                )
+                hk_cov = -(ru / 24) * sr_by_su_5
+                kk_cov = (sr / 720) * (1.0 - sr_by_su_5)
```
patrick-kidger, line 481: I'll be honest, I'm just going to trust you on this one 😅

diffrax/_brownian/tree.py
```diff
@@ -369,14 +429,114 @@ def _body_fun(_state: _State):
...
+            cov = jnp.array(
+                [
+                    [ww_cov, wh_cov, wk_cov],
+                    [wh_cov, hh_cov, hk_cov],
+                    [wk_cov, hk_cov, kk_cov],
+                ]
+            )
+
+            if self._spline == "sqrt":
+                # NOTE: not compatible with jnp.float16
```
patrick-kidger, line 492: I think we should still aim to handle this. Probably this can be done by casting the dtype?
```python
dtype_atleast32 = jnp.result_type(dtype, jnp.float32)
hat_y = jr.multivariate_normal(..., dtype=dtype_atleast32)
hat_y = hat_y.astype(dtype)
```

patrick-kidger: FWIW I think generating a multivariate normal shouldn't be too tricky if we wanted to just do it manually.

I also don't love that this is apparently calling out to SVD. Admittedly, though, I have just done a quick google for an explicit form for the square root of a 3x3 posdef matrix and it looks like there isn't one...

diffrax/_brownian/tree.py
```diff
@@ -369,14 +429,114 @@ def _body_fun(_state: _State):
...
+                hat_y = jnp.zeros(shape=shape + (3,), dtype=dtype)
+            else:
+                raise ValueError(
+                    f"When levy_area='space-time-time', only 'sqrt' and"
+                    f" 'zero' splines are permitted, got {self._spline}."
+                )
+
+            hat_w_sr, hat_hh_sr, hat_kk_sr = [
+                x.squeeze(axis=-1) for x in jnp.split(hat_y, 3, axis=-1)
+            ]
```
patrick-kidger, line 512: I thnk just `hat_w_sr, hat_hh_sr, hat_kk_sr = hat_y` will work?

diffrax/_solver/dopri5.py
```diff
@@ -90,9 +90,9 @@ class Dopri5(AbstractERK):
     """
 
     tableau: ClassVar[ButcherTableau] = _dopri5_tableau
-    interpolation_cls: ClassVar[
-        Callable[..., _Dopri5Interpolation]
-    ] = _Dopri5Interpolation
+    interpolation_cls: ClassVar[Callable[..., _Dopri5Interpolation]] = (
+        _Dopri5Interpolation
+    )
```
patrick-kidger, line 95: Can we skip the spurious reformatting?

(Let's put that in a separate PR if you like? Perhaps this is coming from bumping `ruff`.)

docs/api/brownian.md
```diff
@@ -22,3 +22,38 @@ SDEs are simulated using a Brownian motion as a control. (See the neural SDE exa
         members:
             - __init__
             - evaluate
+
+---
+
+## Levy areas
+
+Brownian controls can return certain types of Levy areas. These are iterated integrals
```
patrick-kidger, line 30: Levy should have a diacritic: Lévy.

I don't always include this when writing informally (including in your name, I know!) but I try to get it right in persisent documentation!

docs/api/brownian.md
```diff
@@ -22,3 +22,38 @@ SDEs are simulated using a Brownian motion as a control. (See the neural SDE exa
...
+│   └── BrownianIncrement
+└── AbstractSpaceTimeLevyArea
+    │   └── SpaceTimeLevyArea
+    └── AbstractSpaceTimeTimeLevyArea
+            └── SpaceTimeTimeLevyArea
+```
+For example if `solver.minimal_levy_area` returns an `AbstractSpaceTimeLevyArea`, then
+the Brownian motion (which is either an `UnsafeBrownianPath` or 
+a `VirtualBrownianTree`) should be initialized with `levy_area=SpaceTimeLevyArea` or 
+`levy_area=SpaceTimeTimeLevyArea`. Note that for the BM, a concrete class must be
```
patrick-kidger, line 46: Nit: I try to avoid the "BM" abbreviation in documentation.

diffrax/_autocitation.py
andyElking: I wanted to include the line `and Theorem 6.1.6 of`, but I suppose it's not strictly necessary.

diffrax/_brownian/path.py
andyElking: I wanted to keep the generated path the same regardless of the Levy area setting, but fair, I guess this isn't really relevant for UBP.

diffrax/_brownian/tree.py
andyElking: Oh, I understand, will do.

diffrax/_brownian/tree.py
andyElking: Oops, will do.

diffrax/_brownian/path.py
andyElking: And I think I was using the wrong key to generate the bm anyway, oops.

diffrax/_brownian/tree.py
andyElking: I meant the latter, but yes, fair point.

diffrax/_brownian/tree.py
andyElking: Sorry, I really should have known this by now 😅

diffrax/_brownian/tree.py
andyElking: Does that improve compilation time? Otherwise I rather wouldn't, because it would require a very complicated `is_leaf=...`.

diffrax/_brownian/tree.py
andyElking: I did the Cholesky decomp manually, but it's very complicated and in my very basic experiments it took slightly longer to compute it with that formula (so computing the decomposition directly from s, r, u), as opposed to first computing the cov matrix, and then doing SVD on that. The issue with Cholesky is that it is very imprecise when the matrix is close to singular (which happens when r is close to s or to u). SVD is the only good option for near-singular matrices. Alternatively we'd have to do a separate case for when r-s or u-r is small, which is probably not ideal.

andyElking: Yes, the type-casting idea is very good.

diffrax/_brownian/tree.py
andyElking: No, I don't think it will, because we're splitting by the last dimension, whereas tuple unpacking splits along the first.

diffrax/_solver/dopri5.py
andyElking: Yes, I'll drop it, sorry.

andyElking:
Hi Patrick, sorry this round of corrections took so long. I made the fixes you mentioned, including adding diacritics on Lévy, which changed some files outside this PR, but we're still only on 12 files changed, so I hope it's fine 😅. I also rebased on top of the current main.

patrick-kidger:
No worries -- and the great news is that I think this is ready to be merged :D
Can you rebase on top of the latest main, and we can get this in!

andyElking:
That's very good to hear, I rebased it and squashed some of the commits. I can squash all of them if needed.

andyElking:
Hi Patrick,
Sorry for yet another delay. Seems like the tests passed, so I hope this means this PR is ready for a merge. Let me know if there's any more fixes I should make.

patrick-kidger:
Aaaaaand merged! :D
This is really great stuff, it's awesome to see how much you're adding to Diffrax.

I've just spun up a new `dev` branch (currently equal to `main` + your PR) in preparation for the next release!

andyElking:
That's great news, thanks!

I will probably open a new PR with Langevin solvers sometime over the next few days. That one is a bit larger than this one, but in my opinion still more lightweight than the SRK PR. But feel free to finish the next release first and turn to Langevin PR afterwards if you prefer.

And after that I'll give you some peace as far as Diffrax is concerned. Upcoming projects are then Langevin MCMC, some signature stuff and finally diffusion models.

Issue #420, opened 2024-05-19
ShARK w/ additive noise


lockwo:
In the documentation on SDE solvers, it says some (e.g. ShARK) should be strong order 1.5 on additive noise SDEs. But in the further documentation (https://docs.kidger.site/diffrax/devdocs/srk_example/) it seems to be about ~1.2. Is that just a numerical effect that is expected?

patrick-kidger:
(Tagging @andyElking.)

I expect this is just a numerical oddity. This kind of messiness around the empirically-measured order is fairly common.

andyElking:
Hi both,
That is indeed a numerical oddity, which I cannot really fix, but I made it a bit better with the PR #424. There I also explained why I think this is happening.

lockwo:
Makes sense, thanks for that @andyElking

PR #424, opened 2024-05-21
Fixes for srk_example.ipynb

andyElking:
This is a small PR fixing two issues in `devdocs/srk_example.ipynb`. The second commit just fixes the name of SPaRK.

The first commit improves the issue that @lockwo mentioned in #420. The problem was that the convergence order "eventually" approaches 1.5, but at the given step size it didn't quite get into that regime yet. So, by moving a bit to the right on the x-axis (i.e. increasing the step sizes) the order increased from 1.2 to 1.33. I imagine if we increased the step sizes even further, we'd get closer to 1.5 (assuming no other numerical errors).
I also increased the number of samples from 100 to 1000, so the lines have less variance.

The reason this issue arose in the first place is that the SDE being simulated there is 7 dimensional and very complicated. The drift vector field is a 5-layer MLP (taking as input both time and position) and the diffusion VF is some weird function of time. This was needed because with simpler SDEs, the order of the solvers was actually significantly higher than 1.5, so the tests kept failing.

As to what exactly makes the SDE "complicated enough" to not converge too fast and how small a step size you need to get to the correct order, I unfortunately don't have any good answers.


patrick-kidger:
LGTM! Thank you for the quick fix :)

Issue #425, opened 2024-05-25
Custom saveat function passed negative time when integrating backwards


mathisgerdes:
Hello! I have encountered unexpected behavior, that to me looks like a bug.

Specifically, the time passed to a custom "SaveAt-fn" seems to be inverted when integrating backwards.
Consider this (trivial) example:
```python
saveat = diffrax.SaveAt(
    t1=False, 
    ts=jnp.linspace(1, 0, 5), 
    fn=lambda t, y, args: t  # save times
)

term = diffrax.ODETerm(lambda t, x, _: 1)
sol = diffrax.diffeqsolve(term, diffrax.Tsit5(), saveat=saveat, t0=1, t1=0, dt0=-1/50, y0=1.0)
```

I would expect `sol.ys` to now contain the same `ts` which I specified to save. However, instead I get negative values `[-1.  , -0.75, -0.5 , -0.25, -0.  ]`. Am I missing something obvious, or is this a bug? Is the `t` in the SaveAt `fn` different in meaning to the one appearing in the ODE term? Thanks!

Reproduced this behavior both locally and on colab, diffrax version 0.5.1.

patrick-kidger:
Thanks for the report! Agreed that this looks like a bug.

Diffrax performs backwards-in-time integration over `t0 > t1` by solving forwards-in-time from `-t0` to `-t1`. Things are flipped around at the start and end of the solve. This means that almost all of our internals don't have to worry about the direction of the solve, which makes them much simpler to write!

We have a `direction` value, that is either `1` or `-1`, which is used to multiply timelike values where appropriate. (See `_integrate.py`.) But it looks like we missed one! We're not multiplying `t` by `direction` before calling the user-specified `SaveAt.fn`.

I can probably fix this quite quickly, but if you're interested in getting involved then I'd also be happy to take a pull request on this! (See [here](https://github.com/patrick-kidger/diffrax/blob/main/CONTRIBUTING.md)!)

mathisgerdes:
Thanks for the swift reply! I will be busy for the next few days, but I would happily look into it afterward. If it's a quick fix for you, feel free to do it...

PR #427, opened 2024-05-29
add direction for saveat + test

lockwo:
Addresses https://github.com/patrick-kidger/diffrax/issues/425

### DIFF
--- a/diffrax/_adjoint.py
+++ b/diffrax/_adjoint.py
@@ -132,7 +132,6 @@ class AbstractAdjoint(eqx.Module):
         passed_solver_state,
         passed_controller_state,
         progress_meter,
-        direction,
     ) -> Any:
         """Runs the main solve loop. Subclasses can override this to provide custom
         backpropagation behaviour; see for example the implementation of
@@ -565,7 +564,6 @@ def _loop_backsolve_bwd(
     throw,
     init_state,
     progress_meter,
-    direction,
 ):
     assert discrete_terminating_event is None
 
@@ -574,7 +572,7 @@ def _loop_backsolve_bwd(
     # using them later.
     #
 
-    del perturbed, init_state, t1, progress_meter, direction
+    del perturbed, init_state, t1, progress_meter
     ts, ys = residuals
     del residuals
     grad_final_state, _ = grad_final_state__aux_stats
--- a/diffrax/_adjoint.py
+++ b/diffrax/_adjoint.py
@@ -429,6 +429,14 @@ def _solve(inputs):
     )
 
 
+# Unwrap jaxtyping decorator during tests, so that these are global functions.
+# This is needed to ensure `optx.implicit_jvp` is happy.
+if _vf.__globals__["__name__"].startswith("jaxtyping"):
+    _vf = _vf.__wrapped__  # pyright: ignore[reportFunctionMemberAccess]
+if _solve.__globals__["__name__"].startswith("jaxtyping"):
+    _solve = _solve.__wrapped__  # pyright: ignore[reportFunctionMemberAccess]
+
+
 def _frozenset(x: Union[object, Iterable[object]]) -> frozenset[object]:
     try:
         iter_x = iter(x)  # pyright: ignore
--- a/diffrax/_adjoint.py
+++ b/diffrax/_adjoint.py
@@ -132,6 +132,7 @@ class AbstractAdjoint(eqx.Module):
         passed_solver_state,
         passed_controller_state,
         progress_meter,
+        direction,
     ) -> Any:
         """Runs the main solve loop. Subclasses can override this to provide custom
         backpropagation behaviour; see for example the implementation of
@@ -564,6 +565,7 @@ def _loop_backsolve_bwd(
     throw,
     init_state,
     progress_meter,
+    direction,
 ):
     assert discrete_terminating_event is None
 
@@ -572,7 +574,7 @@ def _loop_backsolve_bwd(
     # using them later.
     #
 
-    del perturbed, init_state, t1, progress_meter
+    del perturbed, init_state, t1, progress_meter, direction
     ts, ys = residuals
     del residuals
     grad_final_state, _ = grad_final_state__aux_stats
--- a/diffrax/_integrate.py
+++ b/diffrax/_integrate.py
@@ -37,7 +37,7 @@ from ._progress_meter import (
     NoProgressMeter,
 )
 from ._root_finder import use_stepsize_tol
-from ._saveat import SaveAt, SubSaveAt
+from ._saveat import save_y, SaveAt, SubSaveAt
 from ._solution import is_okay, is_successful, RESULTS, Solution
 from ._solver import (
     AbstractImplicitSolver,
@@ -890,6 +890,8 @@ def diffeqsolve(
         return [x for x in out if x is not None]
 
     def _direction_replace(fn):
+        if fn is save_y:
+            return fn
         return lambda t, y, args: fn(direction * t, y, args)
 
     saveat = eqx.tree_at(
--- a/diffrax/_integrate.py
+++ b/diffrax/_integrate.py
@@ -881,24 +881,17 @@ def diffeqsolve(
 
     saveat = eqx.tree_at(_get_subsaveat_ts, saveat, replace_fn=_check_subsaveat_ts)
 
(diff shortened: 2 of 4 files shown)


patrick-kidger:
Awesome -- thanks for this!
I can see there's still some failing tests. I've not dug into them, but please let me know once this is ready for review :)

lockwo:
It seems as though that test fails on main. I was very confused by it (since I tried hard to make sure there were no closure issues) and when I just clone diffrax main and run `python -m pytest test/test_adjoint.py -k 'test_implicit'` I see the same error: `FAILED test/test_adjoint.py::test_implicit - AssertionError`

lockwo:
So you can also see (without just taking my work for it), this PR, which changes nothing as of yet (random 1 line README change), still fails tests: https://github.com/patrick-kidger/diffrax/pull/430.

patrick-kidger:
Fixed in #431! I'm not sure why this ever passed...
We'll get that merged in and then rebase this PR on top of the new main.

diffrax/_adjoint.py
```diff
@@ -482,7 +484,16 @@ def loop(
...
-        inputs = (args, terms, self, kwargs, solver, saveat, init_state)
+        inputs = (
+            args,
+            terms,
+            self,
+            kwargs,
+            solver,
+            saveat,
+            init_state,
+            kwargs.pop("direction"),
```
patrick-kidger, line 495: I think it'd be better to just make `direction` be another keyword argument to `loop`.
(Or for that matter could it just be left in `kwargs` and passed through? So no code changes here whatsoever?)

diffrax/_integrate.py
```diff
@@ -405,7 +406,7 @@ def _body_fun(_save_state):
                 _ts = _save_state.ts.at[_save_state.save_index].set(_t)
                 _ys = jtu.tree_map(
                     lambda __y, __ys: __ys.at[_save_state.save_index].set(__y),
-                    fn(_t, _y, args),
+                    fn(direction * _t, _y, args),
```
patrick-kidger, line 409: I think this change needs to include in all the other places we call `fn`, i.e. when saving at `t0`, `t1`, `steps`, dense`.

I think it'd better to simply wrap `.fn` inside of `saveat`, during the initial set-up for the solve. It's a bit of a mouthful, but something like this:
```python
def _subsaveat_direction_fn(x):
    if _is_subsaveat(x):
        direction_fn = lambda t, y, args: x.fn(direction * t, y, args)
        return eqx.tree_at(lambda x: x.fn, x, direction_fn)
    else:
        return x

saveat = jtu.tree_map(_subsaveat_direction, saveat, is_leaf=_is_subsaveat)
```

test/test_integrate.py
```diff
@@ -398,6 +398,30 @@ def g(t, y, args):
             assert tree_allclose(sol1.derivative(ti), -sol2.derivative(-ti))
 
 
+def test_reverse_time_save():
```
patrick-kidger, line 401: As above I think it's worth including other kinds of `saveat` here, which include `t0`, `t1` etc, to be sure that these cases don't slip through.

patrick-kidger (commented):
Great stuff! I've just merged the test fix, so if you'd like to rebase on top of that then that would be great.

diffrax/_integrate.py
lockwo: That's actually what I had originally, but I thought direction would then get closured in and cause issues. I will roll back to that version

test/test_integrate.py
lockwo: Expanded to include all the flags of saveat. Note that I specify max steps because if that is greater and steps is being logged, it multiples the jnp.inf by direction at the end which results in a mismatch. I don't see a great way to get the custom function to match that and given that those infinities are just padding/holding space, I don't consider it an actual issue

diffrax/_adjoint.py
lockwo: Moved to avoid direction as keyword altogether with just modifying the function

test/test_integrate.py
patrick-kidger: Hmm, that surprises me. The `jnp.inf` should just be the initial buffer values, I don't think they're touched afterwards. In particular I don't think the choice of `direction` should affect them?

Other than that this PR LGTM!

lockwo: Sepcifically, the output ys vs ts are [3.9 3.8 3.7 3.6 3.5 3.4 3.3 3.2 3.1 3.  2.9 2.8 2.7 2.6 2.5 2.4 2.3 2.2
 2.1 2.  1.9 1.8 1.7 1.6 1.5 1.4 1.3 1.2 1.1 1.  0.9 0.8 0.7 0.6 0.5 0.4
 0.3 inf inf inf inf inf inf inf inf inf inf inf inf inf inf inf inf inf
 inf inf inf inf inf inf inf inf inf inf inf inf inf inf inf inf inf inf
 inf inf inf inf inf inf inf inf inf inf inf inf inf inf inf inf inf inf
 inf inf inf inf inf inf inf inf inf inf]

vs

 [ 3.9  3.8  3.7  3.6  3.5  3.4  3.3  3.2  3.1  3.   2.9  2.8  2.7  2.6
  2.5  2.4  2.3  2.2  2.1  2.   1.9  1.8  1.7  1.6  1.5  1.4  1.3  1.2
  1.1  1.   0.9  0.8  0.7  0.6  0.5  0.4  0.3 -inf -inf -inf -inf -inf
 -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf
 -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf
 -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf
 -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf -inf
 -inf -inf]

The ts get a direction multiplied by them at the end, but it is initialized to `inf` without getting direction multiplied. I think this might be a bug, so I fixed it as well and now all the ts inf get initialized with respect to the direction (        ts = jnp.full(out_size, direction * jnp.inf, dtype=time_dtype)). Maybe not a bug since these `inf` are technically meaningless, but at least now the tests pass no problem and I can remove max steps.

lockwo: Hmm, I guess that actually breaks `assert tree_allclose(sol1.ts, -cast(Array, sol2.ts), equal_nan=True)` for steps=True because then the infs are not opposite signs. Not sure if they should be or not, since they are basically superfluous. I will defer to whatever you think. I can revert the last commit and keep max steps and the infs will be sign flipped, or I can make a change to what this failing test expects and the infs will always be positive.

patrick-kidger: Right. So I think this is indicative of a totally separate bug: in some cases we've been returning `-inf` all along:
```python
diffrax.diffeqsolve(diffrax.ODETerm(lambda t, y, args: -y), diffrax.Tsit5(), 1, 0, -0.1, 1, saveat=diffrax.SaveAt(steps=True), max_steps=50).ts
```
This shouldn't be the case. I think we should always return `+inf` for our padding!

It looks like the culprit is as you've identified, namely that we flip things here:

https://github.com/patrick-kidger/diffrax/blob/c6cc85c92b0e14e24092830584d0397a1990d091/diffrax/_integrate.py#L1018-L1020

I think we should adjust the failing test you've highlighted, so that the infs are always positive.

lockwo: Great, I changed that test, and all padding should be positive now.

diffrax/_integrate.py
```diff
@@ -881,6 +881,18 @@ def _check_subsaveat_ts(ts):
 
     saveat = eqx.tree_at(_get_subsaveat_ts, saveat, replace_fn=_check_subsaveat_ts)
 
+    def _subsaveat_direction_fn(x):
+        if _is_subsaveat(x):
+            if x.fn is not save_y:
```
patrick-kidger, line 886: Why this exclusion? I don't it should actually matter? (I'd rather not have an extra branch unless we need it.)

patrick-kidger (commented):
Ultranit aside this LGTM!

diffrax/_integrate.py
patrick-kidger: Ah, makes sense!
Alright -- merged! :D

PR #430, opened 2024-05-30
Implicit closure

lockwo:


patrick-kidger:
Thanks for the clear repro btw!

PR #434, opened 2024-06-06
Lineax VF

lockwo:
see https://github.com/patrick-kidger/diffrax/issues/370#issuecomment-2151248116


patrick-kidger:
Heads-up that this is failing as it's currently targeting `dev`. I believe that's actually the old `dev` branch from the previous release; can you update this to point to `main` instead?

(In case you're curious what's going on here: updates directly to `main` are fine right up until they discuss a new feature in the docs. At that point I spin off a `dev` branch, so that not-yet-released features don't pre-emptively appear in the docs -- which are rebuilt on every merge to `main`.)

PR #436, opened 2024-06-07
Lineax Control Terms

lockwo:
Addresses https://github.com/patrick-kidger/diffrax/issues/370

### DIFF
--- a/diffrax/_misc.py
+++ b/diffrax/_misc.py
@@ -1,4 +1,6 @@
+import warnings
 from collections.abc import Callable
+from functools import wraps
 from typing import Any, cast, Optional
 
 import jax
@@ -10,6 +12,10 @@ import optimistix as optx
 from jaxtyping import Array, ArrayLike, PyTree, Shaped
 
 from ._custom_types import BoolScalarLike, RealScalarLike
+
+
+# PendingDeprecationWarning are usually supressed by default
+warnings.simplefilter("always", PendingDeprecationWarning)
 
 
 _itemsize_kind_type: dict[tuple[int, str], Any] = {
@@ -189,3 +195,20 @@ def upcast_or_raise(
     elif config_value != "standard":
         assert False, f"Unrecognised `JAX_NUMPY_DTYPE_PROMOTION={config_value}`"
     return jnp.astype(x, promote_dtype)
+
+
+def pending_deprecation_warning(cls):
+    orig_init = cls.__init__
+
+    @wraps(orig_init)
+    def new_init(self, *args, **kwargs):
+        warnings.warn(
+            f"{cls.__name__} is pending deprecation and may be removed "
+            "in future versions. Consider using the new alternative "
+            "ControlTerm(lx.DiagonalLinearOperator(...)).",
+            PendingDeprecationWarning,
+        )
+        orig_init(self, *args, **kwargs)
+
+    cls.__init__ = new_init
+    return cls
--- a/diffrax/_misc.py
+++ b/diffrax/_misc.py
@@ -1,6 +1,4 @@
-import warnings
 from collections.abc import Callable
-from functools import wraps
 from typing import Any, cast, Optional
 
 import jax
@@ -12,10 +10,6 @@ import optimistix as optx
 from jaxtyping import Array, ArrayLike, PyTree, Shaped
 
 from ._custom_types import BoolScalarLike, RealScalarLike
-
-
-# PendingDeprecationWarning are usually supressed by default
-warnings.simplefilter("always", PendingDeprecationWarning)
 
 
 _itemsize_kind_type: dict[tuple[int, str], Any] = {
@@ -195,20 +189,3 @@ def upcast_or_raise(
     elif config_value != "standard":
         assert False, f"Unrecognised `JAX_NUMPY_DTYPE_PROMOTION={config_value}`"
     return jnp.astype(x, promote_dtype)
-
-
-def pending_deprecation_warning(cls):
-    orig_init = cls.__init__
-
-    @wraps(orig_init)
-    def new_init(self, *args, **kwargs):
-        warnings.warn(
-            f"{cls.__name__} is pending deprecation and may be removed "
-            "in future versions. Consider using the new alternative "
-            "ControlTerm(lx.DiagonalLinearOperator(...)).",
-            PendingDeprecationWarning,
-        )
-        orig_init(self, *args, **kwargs)
-
-    cls.__init__ = new_init
-    return cls
--- a/diffrax/_term.py
+++ b/diffrax/_term.py
@@ -358,18 +358,8 @@ class ControlTerm(_AbstractControlTerm[_VF, _Control]):
     """
 
     def prod(self, vf: _VF, control: _Control) -> Y:
-        if isinstance(
-            jtu.tree_leaves(
-                vf, is_leaf=lambda x: isinstance(x, lx.AbstractLinearOperator)
-            )[0],
-            lx.AbstractLinearOperator,
-        ):
-            return jtu.tree_map(
-                lambda _vf, _control: _vf.mv(_control),
-                vf,
-                control,
-                is_leaf=lambda x: isinstance(x, lx.AbstractLinearOperator),
-            )
+        if isinstance(vf, lx.AbstractLinearOperator):
+            return vf.mv(control)
         return jtu.tree_map(_prod, vf, control)
 
 
--- a/diffrax/_term.py
+++ b/diffrax/_term.py
@@ -13,7 +13,7 @@ from equinox.internal import ω
 from jaxtyping import ArrayLike, PyTree, PyTreeDef
 
 from ._custom_types import Args, Control, IntScalarLike, RealScalarLike, VF, Y
-from ._misc import upcast_or_raise
+from ._misc import pending_deprecation_warning, upcast_or_raise
 from ._path import AbstractPath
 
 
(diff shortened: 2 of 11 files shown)


lockwo:
New PR for https://github.com/patrick-kidger/diffrax/pull/434. Will take into account https://github.com/patrick-kidger/diffrax/issues/370#issuecomment-2155263828 then mark as ready for review

lockwo:
Supports lineax control term. Removes weakly diagonal from docs.

diffrax/_misc.py
```diff
@@ -12,6 +14,10 @@
 from ._custom_types import BoolScalarLike, RealScalarLike
 
 
+# PendingDeprecationWarning are usually supressed by default
+warnings.simplefilter("always", PendingDeprecationWarning)
+
```
patrick-kidger, line 19: Ah, this I didn't realise. Let's just use a `DeprecationWarning` then, to avoid having to mutate shared global state.

diffrax/_misc.py
```diff
@@ -189,3 +195,20 @@ def upcast_or_raise(
...
+        warnings.warn(
+            f"{cls.__name__} is pending deprecation and may be removed "
+            "in future versions. Consider using the new alternative "
+            "ControlTerm(lx.DiagonalLinearOperator(...)).",
+            PendingDeprecationWarning,
+        )
+        orig_init(self, *args, **kwargs)
+
+    cls.__init__ = new_init
+    return cls
```
patrick-kidger, line 214: Since I think we only use this once, let's inline it where it's used. (In particular as the error message mentions `ControlTerm(lx.DiagonalLinearOperator(...))`!)

diffrax/_term.py
```diff
@@ -288,7 +289,8 @@ def to_ode(self) -> ODETerm:
 - `vector_field`: A callable representing the vector field. This callable takes three
     arguments `(t, y, args)`. `t` is a scalar representing the integration time. `y` is
     the evolving state of the system. `args` are any static arguments as passed to
-    [`diffrax.diffeqsolve`][].
+    [`diffrax.diffeqsolve`][]. This `vector_field` can be a function that returns a
+    jax array, or returns any lineax `AbstractLinearOperator`.
```
patrick-kidger, line 293: Can you add a link to the docs for `lineax.AbstractLinearOperator`?
(Also whilst I'm here, nit jax->JAX.)

diffrax/_term.py
```diff
@@ -335,9 +357,18 @@ class ControlTerm(_AbstractControlTerm[_VF, _Control]):
     """
 
     def prod(self, vf: _VF, control: _Control) -> Y:
+        if isinstance(vf, lx.AbstractLinearOperator):
+            return jtu.tree_map(
+                lambda _vf, _control: _vf.mv(_control),
+                vf,
+                control,
+                is_leaf=lambda x: eqx.is_array(x)
```
patrick-kidger, line 365: I think the `eqx.is_array` bit is unnecessary -- anything matching this will already be a leaf.

patrick-kidger: Actually also, show we have a `tree_map` here or not? The surrounding check is already `isinstance(vf, lx.AbstractLinearOperator)`, so this tree-map is guaranteed to be trivial.

Off the top of my head I don't know if I have an opinion on whether `vf` should be either a `lx.AbstractLinearOperator` or a `PyTree[lx.AbstractLinearOperator]`. WDYT?

docs/usage/extending.md
```diff
@@ -24,7 +24,7 @@ The main points of extension are as follows:
 - **Custom controls** (e.g. **custom interpolation schemes** analogous to [`diffrax.CubicInterpolation`][]) should inherit from [`diffrax.AbstractPath`][].
 
 - **Custom terms** should inherit from [`diffrax.AbstractTerm`][].
-    - For example, if the vector field - control interaction is a matrix-vector product, but the matrix is known to have special structure, then you may wish to create a custom term that can calculate this interaction more efficiently than is given by a full matrix-vector product. For example this is done with [`diffrax.WeaklyDiagonalControlTerm`][] as compared to [`diffrax.ControlTerm`][].
+    - For example, if the vector field - control interaction is a matrix-vector product, but the matrix is known to have special structure, then you may wish to create a custom term that can calculate this interaction more efficiently than is given by a full matrix-vector product.
```
patrick-kidger, line 27: We should probably also highlight how unusual this is here -- that using a `ControlTerm` returning a `lx.AbstractLinearOperator` should be sufficient.

test/test_adjoint.py
```diff
@@ -338,14 +343,21 @@ def g(t, y, args):
         _, k1 = args
         return k1 * y
 
+    def g_lx(t, y, args):
+        _, k1 = args
+        return jtu.tree_map(lx.DiagonalLinearOperator, k1 * y)
```
patrick-kidger, line 348: Why the `tree_map`?

diffrax/_misc.py
lockwo: Changed

diffrax/_misc.py
lockwo: Ok, moved to be inside the weakly diagonal

diffrax/_term.py
lockwo: added link and changed jax

diffrax/_term.py
lockwo: True, I removed the `is_array`. Also true that the if/tree-map was unnecessary. I updated the if statement. 

Unless my understanding of lineax is wrong (which it could be, I'm new to that package), a PyTree[operator] seems like the way to go. Diffrax allows arbitrary pytrees as control, so the output of the vector field could previously be any pytree of arrays (that are just element wise multiplied). My understanding of how to do this in lineax was just to make a pytree of operators and treemap over them. If this is true, then I think PyTree[operator] is definitely the way to go, since pytree based controls are important. If there is a different way to do things in lineax then perhaps there are other approaches to consider here.

docs/usage/extending.md
lockwo: Added text emphasizing support and that this should be rare

test/test_adjoint.py
lockwo: Leftover from a older version I made probably, removed

diffrax/_term.py
patrick-kidger: So both approaches are equally expressive: you could write a linear operator that consumes a pytree and does whatever you like with it.

As such, doing a tree-map here is equivalent to making a block-diagonal operator.

Noodling it over, I think I'm probably leaning towards having just a single operator. This pushes more of the abstraction into Lineax (where it's easier to reason about I think), and means we can skip the potentially error-prone process of appropriately broadcasting PyTrees etc here. I always have to scrutinise tree-map code 10x as much as any other code!

diffrax/_term.py
```diff
@@ -360,6 +395,15 @@ class WeaklyDiagonalControlTerm(_AbstractControlTerm[_VF, _Control]):
         without the "weak". (This stronger property is useful in some SDE solvers.)
     """
 
+    def __init__(self, *args, **kwargs):
+        warnings.warn(
```
tttc3, line 399: Sorry to jump in here, but as `typing_extensions` is already a dependancy, can we use the [`@deprecated`](https://typing-extensions.readthedocs.io/en/latest/index.html#typing_extensions.deprecated) decorator on the class instead? This additionally allows your code editor of choice to highlight the deprecation at coding time: 
deprecation_decorator

To trigger the runtime warning, the `typing_extensions` version in the `pyproject.toml` would need to be bumped to `>=4.9.0` though. What do you think of this @patrick-kidger?

patrick-kidger: Thank you for the heads up -- I think that sounds like an excellent idea! Lets do it. @lockwo?

diffrax/_term.py
lockwo: I see, so for this sort of block diagonal approach, the user should create a custom operator that inherits and implement .mv using this sort of tree map. But in diffrax itself, the code doesn't know about the pytree just calls .mv?

patrick-kidger: Yup, exactly!

diffrax/_term.py
lockwo: Sg, added

diffrax/_term.py
lockwo: Ok, I changed it

lockwo:
Tests are failing because tests are failing on main: https://github.com/patrick-kidger/diffrax/pull/441

diffrax/_term.py
lockwo: I haven't use the deprecated wrapper before, and I am unable to get it to actually surface a warning. I don't see a lot of docs on the matter. Do you have any advice @tttc3 ? That is the only failing test right now (failing since it doesn't raise a warning).

lockwo:
The rebase makes the tests pass @patrick-kidger (the one fail is due to the `@deprecated` flag not raising an error which I am looking into)

diffrax/_term.py
lockwo: I see runtime warnings are not expected here (https://discuss.python.org/t/pep-702-marking-deprecations-using-the-type-system/23036/21?page=2) I will remove the test

tttc3: I would expect the warning to raise as the decorator is a backward compatible port of [`@warnings.deprecate`](https://docs.python.org/3.13/library/warnings.html#warnings.deprecated) from python 3.13.  For example, my original example raises the following warning at runtime:
```bash
demo.py:11: DeprecationWarning: This is deprecated.
  use_deprecated = dont_use()
```

From what I can see, there is some nuanced interaction between the `eqx.Module` initialisation process and the `if cls is arg` check used in `deprecated` to trigger the warning.

patrick-kidger:
Great stuff! This PR looks ready to go to me. If you can rebase on top of the latest `dev` then I'd love to merge this.

diffrax/_term.py
patrick-kidger: Ah, `eqx.Module` does indeed do some fairly cunning stuff during the initialisation -- it swaps out the class with a slightly different one that unfreezes it during `__init__` only. This is what makes our modules be frozen most of the time, but still support `self.foo = foo` assignment during `__init__`.

I think in this case, raising a warning is more important than using the `@deprecated` decorator -- so best to put this back as an explicit `warnings.warn` I think!

lockwo:
> Great stuff! This PR looks ready to go to me. If you can rebase on top of the latest `dev` then I'd love to merge this.

Rebased!

patrick-kidger:
Hmm, GitHub doesn't seem to believe so?
"This branch cannot be rebased due to conflicts."

Screenshot 2024-06-26 at 22 37 00

lockwo:
Wow, weird. It looks totally fine on my end?

Screenshot 2024-06-26 at 1 50 25 PM

Let me try opening a new PR maybe?

lockwo:
Ok, seems like my git locally somehow got messed up with the remote tracking. Let me know if this fixes it

patrick-kidger:
And merged! Thank you for all your hard work getting this in :D I'm really glad to be able to take advantage of Lineax like this now :)

PR #449, opened 2024-06-26
Langevin PR

andyElking:
Hi Patrick,

This PR contains all the new Langevin solvers. All of these inherit from `AbstractLangevinSRK` in `langevin_srk.py`.
Another important addition is LangevinTerm in `_term.py`. I explained why it is needed in a comment bellow.

I haven't added the new solvers to the docs and autocite yet, because 1) the relevant paper is not on arxiv yet, but might be in a month or two and 2) I expect you might suggest several changes, so I don't want to redo the docs many times. Still, I think the docstrings and comments are quite comprehensive.

I'm making this PR now so you have ample time to have a look at it, but I will be away for the next few weeks, so there is aboslutely no hurry.

Best,
Andraž

### DIFF
--- a/diffrax/__init__.py
+++ b/diffrax/__init__.py
@@ -75,6 +75,7 @@ from ._solver import (
     AbstractSRK as AbstractSRK,
     AbstractStratonovichSolver as AbstractStratonovichSolver,
     AbstractWrappedSolver as AbstractWrappedSolver,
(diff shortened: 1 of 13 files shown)


patrick-kidger:
(I haven't closed this intentionally, that just happens when the target branch is removed. Can you re-open this against main?)

PR #454, opened 2024-07-01
Complex fixes in SDEs

Randl:
Some fixes imported from https://github.com/patrick-kidger/diffrax/pull/418
(SDEs still do not pass all tests)

### DIFF
--- a/diffrax/_solver/milstein.py
+++ b/diffrax/_solver/milstein.py
@@ -211,7 +211,7 @@ class ItoMilstein(AbstractItoSolver):
         leaves_ΔwΔw = []
         for i1, l1 in enumerate(leaves_Δw):
             for i2, l2 in enumerate(leaves_Δw):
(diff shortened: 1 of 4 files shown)


patrick-kidger:
These look reasonable to me!

I just want to check which argument is being conjugated, though. [There isn't a convention on which argument of a sequilinear form is conjugated](https://en.wikipedia.org/wiki/Sesquilinear_form#Convention). Ultimately this is a choice that we should (a) perform consistently across the whole ecosystem and (b) document carefully. I've just taken a quick look at Lineax and I *think* we're doing the right thing there too, but just wanted to check?

Randl:
I was doing the first argument conjugation everywhere (that's what I'm used to). If nobody else added conjugations, we should be good.

patrick-kidger:
Alright! First-argument conjugation it is :D

PR #453, opened 2024-07-01
Langevin PR

andyElking:
Hi Patrick,

Thanks for the heads up, here's the reuploaded PR (with another quick fix).

This PR contains all the new Langevin solvers. All of these inherit from AbstractLangevinSRK in langevin_srk.py.
Another important addition is LangevinTerm in _term.py. I explained why it is needed in a comment bellow.

I haven't added the new solvers to the docs and autocite yet, because 1) the relevant paper is not on arxiv yet, but might be in a month or two and 2) I expect you might suggest several changes, so might as well write the docs once the rest is stationary. Still, I think the docstrings and comments are quite comprehensive.

I'm making this PR now so you have ample time to have a look at it, but I will be away for the next few weeks, so there is aboslutely no hurry.

Best,
Andraž

PR #463, opened 2024-07-12
KL Divergence for Latent SDEs

lockwo:
A continuation of https://github.com/patrick-kidger/diffrax/pull/402 with the new 0.6.0 lineax changes


diffrax/_integrate.py
```diff
@@ -124,6 +125,8 @@ def _term_compatible(
     contr_kwargs: PyTree[dict],
 ) -> bool:
     error_msg = "term_structure"
+    if isinstance(y, KLState):
+        y = y.y
```
patrick-kidger, line 129: I don't think this can be allowed unfortunately :(
The contents of `_integrate.py` try to avoid ever special-casing particular inputs. Rather, they try to simply interact with the abstractions provided by ABCs.
(As you scroll through this file you'll find that there are a few exceptions to this, but I think they are all exclusively minor performance optimisations.)

The general principle is that someone should be able to implement equivalent functionality on top of Diffrax, without ever needing to change its internals.

diffrax/_solver/kl.py
```diff
@@ -0,0 +1,291 @@
...
+    """
+    Compute the KL divergence.
+    """
+    drift1 = drift_term1.vf(t0, y0, args)
+    drift2 = drift_term2.vf(t0, y0, args)
+    drift = jtu.tree_map(operator.sub, drift1, drift2)
+
+    diffusion = diffusion_term.vf(t0, y0, args)  # assumes same diffusion
+
+    divergences = lx.linear_solve(diffusion, drift, solver=linear_solver).value
```
patrick-kidger, line 61: This is assuming that `diffusion` is a linear operator.
Probably the best thing to do is just an `isinstance` check and raise an explicit error if not? (Or convert to `MatrixLinearOperator`?)

diffrax/_solver/kl.py
```diff
@@ -0,0 +1,291 @@
...
+
+    kl_divergence = jtu.tree_map(lambda x: 0.5 * jnp.sum(x**2), divergences)
+    kl_divergence = jtu.tree_reduce(operator.add, kl_divergence)
+
+    return KLState(drift1, jnp.squeeze(kl_divergence))
```
patrick-kidger, line 66: Avoid using `jnp.squeeze` without specifying a dimension. Otherwise you get different behaviour based on whether or not the input just-so-happens to have `n=1` in an array of shape `(n, 1)`.

In this case it looks to me like this should already be a scalar anyway, however?

diffrax/_solver/kl.py
```diff
@@ -0,0 +1,291 @@
...
+
+
+class _KLDrift(AbstractTerm):
+    drift1: ODETerm
+    drift2: ODETerm
+    diffusion: ControlTerm
+    linear_solver: lx.AbstractLinearSolver
+
+    def vf(self, t: RealScalarLike, y: Y, args: Args) -> KLState:
+        y = y.y
```
patrick-kidger, line 76: I think this should be `y: KLState`?

diffrax/_solver/kl.py
```diff
@@ -0,0 +1,291 @@
...
+class _KLControlTerm(AbstractTerm):
+    control_term: ControlTerm
+
+    def vf(self, t: RealScalarLike, y: Y, args: Args) -> KLState:
+        y = y.y
+        vf = self.control_term.vf(t, y, args)
+        return KLState(vf, jnp.array(0.0))
+
+    def contr(self, t0: RealScalarLike, t1: RealScalarLike, **kwargs) -> KLState:
+        return KLState(self.control_term.contr(t0, t1), jnp.array(0.0))
```
patrick-kidger, line 97: I don't think this needs to output a `KLState`. The output of this is only used as an input to `prod`/`vf_prod`, which we control.

diffrax/_solver/kl.py
```diff
@@ -0,0 +1,291 @@
...
+
+    for every time interval. This is useful for KL based latent SDEs. The output
+    of the solution.ys will be a tuple containing (ys, kls) where kls is the KL
+    divergence integration at that time. Unless the noise is diagonal, this
+    inverse can be extremely costly for higher dimenions.
+
+    The input must be a `MultiTerm` composed of the first SDE with drift `f`
+    and diffusion `g` and the second either a SDE or just the drift term
+    (since the diffusion is assumed to be the same). For example, a type
+    of: `MuliTerm(MultiTerm(ODETerm, _DiffusionTerm), ODETerm)`.
```
patrick-kidger, line 135: As per this comment:
https://github.com/patrick-kidger/diffrax/pull/402#pullrequestreview-2013511154
and also the updated term docs:
https://docs.kidger.site/diffrax/api/terms/
then this outer `MultiTerm` isn't really in-keeping. We're not adding all of these extra terms on to the same evolving state.

Bearing in mind that the rest of Diffrax has to see this as just another SDE solve.

I think this one might take a bit more iteration to get to something that's obeying the abstractions in the way they're designed, I'm afraid.

diffrax/_solver/kl.py
```diff
@@ -0,0 +1,291 @@
...
+        self,
+        terms: PyTree[AbstractTerm],
+        t0: RealScalarLike,
+        t1: RealScalarLike,
+        y0: Y,
+        args: Args,
+        solver_state: _SolverState,
+        made_jump: BoolScalarLike,
+    ) -> tuple[Y, Optional[Y], DenseInfo, _SolverState, RESULTS]:
+        terms1, terms2 = terms.terms
```
patrick-kidger, line 208: Note that this is inconsistent with the type annotation above, btw.

diffrax/_solver/kl.py
```diff
@@ -0,0 +1,291 @@
...
+            drift_term1, is_leaf=lambda x: isinstance(x, ODETerm)
+        )
+        drift_term2 = jtu.tree_map(
+            lambda x: x if isinstance(x, ODETerm) else None,
+            terms2,
+            is_leaf=lambda x: isinstance(x, ODETerm),
+        )
+        drift_term2 = jtu.tree_leaves(
+            drift_term2, is_leaf=lambda x: isinstance(x, ODETerm)
+        )
```
patrick-kidger, line 224: I don't think these should be necessary -- we can just assume the appropriate terms have been passed outside of any tree structure. (E.g. see how the other SDE-specific solvers work.)

diffrax/_solver/kl.py
```diff
@@ -0,0 +1,291 @@
...
+
+        drift_term1 = eqx.error_if(
+            drift_term1, len(drift_term1) != 1, "First SDE doesn't have one ODETerm!"
+        )
+        drift_term2 = eqx.error_if(
+            drift_term2, len(drift_term2) != 1, "Second SDE doesn't have one ODETerm!"
+        )
```
patrick-kidger, line 231: Note that these are compiletime errors, so you don't need to make them runtime errors!

patrick-kidger (commented):
Okay, I really like the example here!

I'm afraid this might still take a bit more iteration to get to something clean enough to merge, though -- see my comments. :)

diffrax/_integrate.py
lockwo: The reason this was here was because the term checking would fail since the term itself is unaware of the KL state. I was thinking of ways around this, and it seems like the easiest way would be to require the term to be aware of it. 

Much of the design was focused on the user not actively controlling the KLstate, but maybe it makes more sense to require them to provide it and for their term to actually consume it, rather than trying to totally hide it and make it so they don't have to do anything. 

I will push a change enforcing that requirement.

diffrax/_solver/kl.py
lockwo: Changed

diffrax/_solver/kl.py
lockwo: Removed

diffrax/_solver/kl.py
lockwo: Correct, fixed

diffrax/_solver/kl.py
lockwo: Understandable, changed

diffrax/_solver/kl.py
lockwo: Fixed

diffrax/_solver/kl.py
lockwo: Removed

diffrax/_solver/kl.py
lockwo: I think I get what you're saying, multiterm implies a single differential equation "unit". So the composed multi terms is bad form. However, I'm not sure I see the difficulty going forward, I can replace it with tuple (multiterm, multiterm) or even tuple (multiterm, ode term). Which seems to adhere to this principle of multiterm = sde unit, since we are integrating two simultaneous SDEs, while also falling in line with other solvers (such as implicit Euler as you remarked). 

On the terms vs solver approach, I am open to both. I think in my many iterations/experimentations I found the solver approach more in line with my thinking about the nature of the problem, specifically the original idea of (terms, kl_term) I didn't see as appealing since the KL_term relies on information from the other term and I didn't see a clean way to do that. However, having terms with a term wrapper is very doable (but may not mesh with the repo as well).

lockwo:
> Okay, I really like the example here!
> 
> I'm afraid this might still take a bit more iteration to get to something clean enough to merge, though -- see my comments. :)

Happy to iterate on cleaning it, I think the biggest question is the design one (on solvers, terms, and how to represent the problem in diffrax). Once that is resolved, I can iterate quickly to get the rest in :)

diffrax/_solver/kl.py
lockwo: Maybe introducing a totally new term is ok (given the remarks in https://github.com/patrick-kidger/diffrax/pull/453), in which case the approach of a KLTerm (rather than a solver), is doable. Given the restricted nature of terms so far, I originally thought that wasn't in line with the package

lockwo:
Ok, I took the feedback from Andraz's Langevin PR regarding terms and incorporated it into this PR. I think it made things simpler and more in line with the diffrax philosophy, let me know what you think. Basically, now like Langevin, there's just a function that accepts multi term and returns a multi term of private terms that can be consumed by any solver.

Issue #474, opened 2024-08-01
Additive SDE throws error with SRK style solvers


ParticularlyPythonicBS:
Hi,
Can you help me debug why this SDE would throw errors for SRK solvers, but works and integrates fine with ERK and Milstein? 
Here is a simplified version of the code:
```python
import os
import multiprocessing

os.environ["XLA_FLAGS"] = "--xla_force_host_platform_device_count={}".format(
    multiprocessing.cpu_count()
)
# os.environ["JAX_TRACEBACK_FILTERING"] = "off"
# os.environ["EQX_ON_ERROR"] = "breakpoint"

import jax
import diffrax as dfx
import jax.numpy as jnp
import time

SEED = 0
KEY = jax.random.PRNGKey(SEED)

m = 0.01 # inertia
gamma = 0.1 # viscosity

amplitude = 0.42 # amplitude of the driving force
omega = 1 # frequency of the driving force
drive_period = 2 * jnp.pi / omega 

alpha = -1 # linear spring constant
beta = 1 # cubic spring constant

sigma = 0.123 # noise intensity

x0 = 1.0 # initial position
v0 = 0.0 # initial velocity
state0 = jnp.array([x0, v0])

t_min = 0.0
t_max = 2**(10) * drive_period
dt = 2 **(-8) * drive_period

def functional_duffing(t: float, state: jnp.array,
                    args: list[float])->jnp.array:
    x,v = state
    dx = v
    
    gamma, alpha, beta, amplitude, omega, m = args

    driving = amplitude * jnp.cos(omega * t)
    damping = gamma * v
    spring = alpha * x + beta * x ** 3
    dv = (driving - damping - spring)/m

    dstate = jnp.array([dx, dv])
    return dstate

KEY, noise_key = jax.random.split(KEY)
term = dfx.ODETerm(functional_duffing)
args = [gamma, alpha, beta, amplitude, omega, m]

brownian_noise= dfx.VirtualBrownianTree(t_min, t_max, tol=1e-3, shape=(), key=noise_key)
def noise(t, y, args):
    return jnp.array([0, sigma])

noise_term = dfx.ControlTerm(noise, brownian_noise)
terms = dfx.MultiTerm(term, noise_term)
solver = dfx.ShARK()
saveat = dfx.SaveAt(ts = jnp.arange(t_min, t_max, dt))

begin = time.time()
[...]

Issue #461, opened 2024-07-10
Coupled SDE System Implementation


aspannaus:
Hi all, 

thanks for the great library. I'm having an issue implementing a coupled system of SDEs. I'm getting an ```ValueError: `terms` must be a PyTree of `AbstractTerms` (such as `ODETerm`)``` error. The system is: 

```math
\begin{aligned}
		\frac{\mathrm{d} S(t)}{\mathrm{d} t} &= -\beta(t)S(t)\frac{I(t)}{N} \mathrm{d} t, \\
		\frac{\mathrm{d} I(t)}{\mathrm{d}t} &= (\beta(t)S(t)\frac{I(t)}{N} - \gamma(t) I(t)) \mathrm{d} t,\\
		\frac{\mathrm{d} R(t)}{\mathrm{d}t} &= \gamma(t) I(t)\, \mathrm{d} t,\\
		\frac{\mathrm{d} \log\beta(t)}{\mathrm{d}t} &= w_3\mathrm{d} B_w(t),\\
        \frac{\mathrm{d} \log\gamma(t)}{\mathrm{d}t} &= u_3 \mathrm{d}B_u(t)
\end{aligned}
```

The code is
```

import jax.numpy as jnp
import jax.random as jr
import matplotlib.pyplot as plt

import diffrax

def sde_drift(t, y, args):
    N, _ = args
    beta = jnp.exp(y[3])
    gamma = jnp.exp(y[4])
    dS = -(beta * y[0] * y[1]) / N
    dI = (beta * y[0] * y[1]) / N - y[1] * gamma
    dR = y[1] * gamma
    # only diffusion, no drift
    dbeta = 0.0  # jnp.array([0.0])
    dgamma = 0.0  # jnp.array([0.0])
    dy = jnp.array([dS, dI, dR, dbeta, dgamma])

    return dy

def sde_diffusion(t, y, args):
    _, sigma_1 = args
    y1, y2, y3, y4, y5 = y
    diagonal = jnp.array([0.0, 0.0, 0.0, sigma_1 * y4, sigma_1 * y5])
    return diagonal 


def sde():

    t0 = 0
    t1 = 100
    dt0 = 0.1
    y0 = jnp.array([3990.0, 10.0, 0.01, jnp.log(0.25), jnp.log(0.05)])
    args = (4000.0, 0.2)

[...]

Issue #446, opened 2024-06-24
Can't use Equinox inside `term`


pascal-mueller:
I have this code solving a PDE. If I set the force inside `equations` to a numerical value, everything is fine but if I try to replace it with a neural network, I get:
```

% python osc.py
Traceback (most recent call last):
  File ".../project_3/osc.py", line 71, in <module>
    solution = dfx.diffeqsolve(
               ^^^^^^^^^^^^^^^^
  File ".../project_3/venv/lib/python3.12/site-packages/equinox/_jit.py", line 206, in __call__
    return self._call(False, args, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../project_3/venv/lib/python3.12/site-packages/equinox/_module.py", line 1053, in __call__
    return self.__func__(self.__self__, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../project_3/venv/lib/python3.12/site-packages/equinox/_jit.py", line 200, in _call
    out = self._cached(dynamic_donate, dynamic_nodonate, static)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../project_3/venv/lib/python3.12/site-packages/jax/_src/traceback_util.py", line 179, in reraise_with_filtered_traceback
    return fun(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^
  File ".../project_3/venv/lib/python3.12/site-packages/jax/_src/pjit.py", line 327, in cache_miss
    outs, out_flat, out_tree, args_flat, jaxpr, attrs_tracked = _python_pjit_helper(
                                                                ^^^^^^^^^^^^^^^^^^^^
  File ".../project_3/venv/lib/python3.12/site-packages/jax/_src/pjit.py", line 175, in _python_pjit_helper
    attrs_tracked) = _infer_params(jit_info, args, kwargs)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../project_3/venv/lib/python3.12/site-packages/jax/_src/pjit.py", line 627, in _infer_params
    jaxpr, consts, out_shardings_flat, out_layouts_flat, attrs_tracked = _pjit_jaxpr(
                                                                         ^^^^^^^^^^^^
[...]

PR #478, opened 2024-08-07
Improve errors

lockwo:
Previously, the abstract term error, now:

https://github.com/patrick-kidger/diffrax/issues/446: `ValueError: Error while tracing ODETerm(vector_field=<function equations>).vf: matmul input operand 1 must have ndim at least 1, but it has ndim 0` 

https://github.com/patrick-kidger/diffrax/issues/461: `ValueError: The brownian increment BrownianIncrement(dt=f32[], W=f32[5]) does not have the minimal Levy Area <class 'diffrax._custom_types.AbstractSpaceTimeLevyArea'>.`

https://github.com/patrick-kidger/diffrax/issues/474: `ValueError: The brownian increment BrownianIncrement(dt=f32[], W=f32[]) does not have the minimal Levy Area <class 'diffrax._custom_types.AbstractSpaceTimeLevyArea'>.`

Just generally propagating the actual error

### DIFF
--- a/diffrax/_integrate.py
+++ b/diffrax/_integrate.py
@@ -176,7 +176,10 @@ def _term_compatible(
 
                 contr = ft.partial(term.contr, **term_contr_kwargs)
                 # Work around https://github.com/google/jax/issues/21825
-                control_type = eqx.filter_eval_shape(contr, 0.0, 0.0)
+                try:
+                    control_type = eqx.filter_eval_shape(contr, 0.0, 0.0)
+                except ValueError as e:
+                    raise ValueError(f"Error while tracing {term}.contr: " + str(e))
                 control_type_compatible = eqx.filter_eval_shape(
                     better_isinstance, control_type, control_type_expected
                 )
--- a/diffrax/_integrate.py
+++ b/diffrax/_integrate.py
@@ -1014,10 +1014,15 @@ def diffeqsolve(
     del timelikes
 
     # Backward compatibility
-    if isinstance(
-        solver, (EulerHeun, ItoMilstein, StratonovichMilstein)
-    ) and _term_compatible(
-        y0, args, terms, (ODETerm, AbstractTerm), solver.term_compatible_contr_kwargs
+    if (
+        isinstance(solver, (EulerHeun, ItoMilstein, StratonovichMilstein))
+        and _term_compatible(
+            y0,
+            args,
+            terms,
+            (ODETerm, AbstractTerm),
+            solver.term_compatible_contr_kwargs,
+        )[0]
     ):
         warnings.warn(
             "Passing `terms=(ODETerm(...), SomeOtherTerm(...))` to "
--- a/diffrax/_integrate.py
+++ b/diffrax/_integrate.py
@@ -166,7 +166,7 @@ def _term_compatible(
                 vf_type_expected, control_type_expected = term_args
                 try:
                     vf_type = eqx.filter_eval_shape(term.vf, 0.0, yi, args)
-                except ValueError as e:
+                except Exception as e:
                     raise ValueError(f"Error while tracing {term}.vf: " + str(e))
                 vf_type_compatible = eqx.filter_eval_shape(
                     better_isinstance, vf_type, vf_type_expected
@@ -178,7 +178,7 @@ def _term_compatible(
                 # Work around https://github.com/google/jax/issues/21825
                 try:
                     control_type = eqx.filter_eval_shape(contr, 0.0, 0.0)
-                except ValueError as e:
+                except Exception as e:
                     raise ValueError(f"Error while tracing {term}.contr: " + str(e))
                 control_type_compatible = eqx.filter_eval_shape(
                     better_isinstance, control_type, control_type_expected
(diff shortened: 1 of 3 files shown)


patrick-kidger:
(Let me know when you want a review btw, I'm ignoring this for now whilst it's a draft -- which I'm taking as a "you don't need to review this yet" signal)

lockwo:
> (Let me know when you want a review btw, I'm ignoring this for now whilst it's a draft -- which I'm taking as a "you don't need to review this yet" signal)

Yea, that's exactly what I mean by draft, you can ignore it. But with the weird backward edge case resolved, it is ready for you to take a look at

diffrax/_integrate.py
```diff
@@ -136,18 +137,22 @@ def _check(term_cls, term, term_contr_kwargs, yi):
                 for term, arg, term_contr_kwarg in zip(
                     term.terms, get_args(_tmp), term_contr_kwargs
                 ):
-                    if not _term_compatible(yi, args, term, arg, term_contr_kwarg):
+                    if not _term_compatible(
+                        yi, args, term, arg, term_contr_kwarg, bwd_compat
+                    ):
```
patrick-kidger, line 142: I think this is now wrong, as the return type is a 2-tuple rather than a boolean?

I think perhaps we should instead change the name and signature of this function to be `_assert_term_compatible(...) -> None`, and have it raise an error if it finds an issue?
In a couple of cases (here; in the backward compatibility), we can then try-except the error in question. (Here to do a `raise ValueError(...) from e`; in the backward compability just to treat it as a boolean.)

lockwo: Sure, changing to assert makes sense, then I can make the try catch block explicitly around the backward compatibility code rather than as a flag. It's not hurdle keeping it, but I'm curious what's the motivation for this specific backward compatible section? Diffrax (and equinox) have introduced breaking/non backward compatible changes (I think new events did for example, and equinox is about with jax, and I'm sure other cases).

patrick-kidger: Really just that I try to avoid backward compatibility breaks if it's reasonable to do so.

The change with events, for example, includes a similar backward compatibility converter. That one was a little imperfect because we no longer provide the same information as the old events API (which was far too broad), but was still a best-effort!

diffrax/_integrate.py
```diff
@@ -162,32 +163,37 @@ def _check(term_cls, term, term_contr_kwargs, yi):
...
 
     try:
         with jax.numpy_dtype_promotion("standard"):
             jtu.tree_map(_check, term_structure, terms, contr_kwargs, y)
-    except ValueError:
+    except Exception as e:
         # ValueError may also arise from mismatched tree structures
-        return False
-    return True
+        raise ValueError("Terms are not compatible with solver! " + str(e))
```
patrick-kidger, line 196: I think this needs either a `from e` after the end, to chain together exceptions.

diffrax/_solver/srk.py
```diff
@@ -363,7 +363,11 @@ def make_zeros_aux(leaf):
         # Now the diffusion related stuff
         # Brownian increment (and space-time Lévy area)
         bm_inc = diffusion.contr(t0, t1, use_levy=True)
-        assert isinstance(bm_inc, self.minimal_levy_area)
+        if not isinstance(bm_inc, self.minimal_levy_area):
+            raise ValueError(
+                f"The brownian increment {bm_inc} does not have the "
```
patrick-kidger, line 368: Nit: capital `B` Brownian, as it is someone's name.

diffrax/_integrate.py
```diff
@@ -1006,29 +1012,35 @@ def _promote(yi):
...
+                "Passing `terms=(ODETerm(...), SomeOtherTerm(...))` to "
+                f"{solver.__class__.__name__} is deprecated in favour of "
+                "`terms=MultiTerm(ODETerm(...), SomeOtherTerm(...))`. This means that "
+                "the same terms can now be passed used for both general "
+                "and SDE-specific solvers!",
+                stacklevel=2,
+            )
+            terms = MultiTerm(*terms)
+        except Exception as _:
+            pass
```
patrick-kidger, line 1034: This would be better structured as
```python
try:
    _assert_term_compatible(...)
except ValueError:
    pass
else:
    warnings.warn(...)
    terms = MultiTerm(*terms)
```
to avoid placing too much in the `try` block.

diffrax/_integrate.py
lockwo: Ah yes, the beauty of python 3

patrick-kidger:
Great! This LGTM.

Looks like one of the tests has gone a bit wonky, I'm not sure why. Anyway, happy to merge this once things are green.

(Whilst I'm here, I'm also noticing that our tests have quite a few warnings in them. Some of these are expected -- fixed upstream in Equinox, which should have a new release any day now. The one about `WeaklyDiagonalControlTerm` could probably do with being suppressed, though.)

patrick-kidger:
And merged! :)

diffrax/_integrate.py
lockwo: I admire your commitment to backwards compatibility for pre 1.0 releases, in the crucible I was forged (qiskit) there were so many breaking changes in minor releases pre 1.0 that they literally made a YouTube series called breaking changes to help explain them all every minor release.

Issue #483, opened 2024-08-10
Why are step_ts and jump_ts treated differently here?


andyElking:
Hi Patrick,

Am I correct in saying that the only differences between step_ts and jump_ts are the following:

-` jump_ts` cannot be integers, but must be floats (so that you can do `prevbefore` and `nextafter`)
- `_clip_jump_ts` also returns `made_jump`, which is used to determine whether we need to do `_t1 = nextafter(nextafter(t1))`.

But in addition to those discrepancies, it seems Diffrax treats them differently in one other way as well, which I am not sure I understand. Namely, the line below uses `prev_dt=prev_dt` if the step was clipped due to a jump, but `prev_dt=t1-t0` if the step was clipped due to `step_ts`. I don't see why we should make a distinction between these two cases.

https://github.com/patrick-kidger/diffrax/blob/b977dce78382a4a5dc3e84e4bf25dfa2a8ae2bb2/diffrax/_step_size_controller/adaptive.py#L561

I would go even further and say that the line should just say `prev_dt=t1-t0` in all cases. This is because the error of the current step depends on `t1-t0`, rather than on `prev_dt`, so I feel like keeping `prev_dt` in controller state is not needed. Here is what could go wrong with the current setup: 

Say `prev_dt=0.1`, but due to `jump_ts` it was clipped to `t1-t0 = 0.01`. Also assume that the error was large and the step gets rejected and assume that the controller computes `factor=0.5`. Then the next step-size proposal will be `0.05`, which is bigger than the step that was just taken, so it will again be clipped by `jump_ts` to `0.01`, resulting in an infinite loop. Instead the new step proposal should just be `(t1-t0)*factor = 0.005`, which would presumably result in a smaller error and move forward.

On the other hand if the step was clipped to a much smaller size than was intended (i.e. `t1-t0 << prev_dt`), then this will usually reflect in the error being small accordingly, resulting in a large `factor`. This means that  `(t1-t0)*factor` would be again a reasonably large step-size proposal, whereas `prev_dt*factor` would be disproportionately massive.

Let me know if I missed something.

patrick-kidger:
So I don't think this should ever be an infinite loop -- as the next time around then `prev_dt` (pulled out of the controller state) should be `0.05`, and we'll keep on shrinking the step.

[...]

PR #484, opened 2024-08-12
Added JumpStepWrapper

andyElking:
Hi Patrick,

I factored the `jump_ts` and `step_ts` out of the `PIDController` into `JumpStepWrapper` (I'm not very set on this name, lmk if you have ideas). I also made it behave as we discussed in #483. In particular, the following three rules are maintained:

1. We always have `t1-t0 <= prev_dt` (this is checked via `eqx.error_if`), with inequality only if the step was clipped or if we hit the end of the integration interval (we do not explicitly check for that).
2. If the step was accepted, then `next_dt` must be `>=prev_dt`.
3. If the step was rejected, then `next_dt` must be `< t1-t0`.

We achieve this in a very simple way here: https://github.com/patrick-kidger/diffrax/blob/78b122adf39b2f8d26a79d0ac239a2fb675653a1/diffrax/_step_size_controller/jump_step_wrapper.py#L119-L123

The next step is to add a parameter `JumpStepWrapper.revisit_rejected_steps` which does what you expect. That will appear in a future commit in this same PR.


andyElking:
I now also added the functionality to revisit rejected steps. In addition, I also imporved the runtime of `step_ts` and `jump_ts`, because the controller no longer searches the whole array each time, but keeps an index of where in the array it was previously.

Also I think there was a bug in the PID controller, where it would sometimes reject a step, but have `factor>1`. To remedy this I modified the following:
https://github.com/patrick-kidger/diffrax/blob/501bed52fb3726aaee41798aaab46d2bdff1a680/diffrax/_step_size_controller/adaptive.py#L569-L574
I think possibly something smaller than just `self.safety` would make even more sense, I feel like if a step is rejected the next step should be at least 0.5x smaller. But I'm not an expert.

I added a test for revisiting steps and it all seems to work. I also sprinkled in a bunch of `eqx.error_if` statements to make sure the necessary invariants are always maintained. But this is a bit experimental, so maybe there are some bugs I didn't test for.

I think I commented the code quite well, so hopefully you can easily notice if I made a mistake somewhere.

P.S.: Sorry for bombarding you with PRs. As far as I'm concerned this one is very low priority, I can use the code even if it isn't merged into diffrax proper.

andyElking:
Hi @patrick-kidger,
I got rid of some `eqx.error_if`s that I added to my `JumpStepWrapper` and redid the timing benchmarks. My new implementation was already faster than the old `PIDController` before, but now this is way more significant, especially when `step_ts` is long (think `>100`). Surprisingly, it is faster even when it has to revisit rejected steps. See https://github.com/patrick-kidger/diffrax/blob/345e23ad34abe095d25b8be8d621360afd592165/benchmarks/jump_step_timing.py#L126-L128

diffrax/_step_size_controller/adaptive.py
```diff
@@ -353,35 +349,14 @@ def dynamics(t, y, args):
     dtmin: Optional[RealScalarLike] = None
     dtmax: Optional[RealScalarLike] = None
     force_dtmin: bool = True
-    step_ts: Optional[Real[Array, " steps"]] = eqx.field(
-        default=None, converter=_none_or_array
-    )
-    jump_ts: Optional[Real[Array, " jumps"]] = eqx.field(
-        default=None, converter=_none_or_array
-    )
```
patrick-kidger, line 361: So I think it's important that we don't break the existing `PIDController`. If at all possible I really try to avoid backward-incompatible changes.

In this case I think we have two choices:
1. Keep all of the code here, but if `self.step_ts is not None` or `self.jump_ts is not None`, then emit a warning in `__check_init__` to prefer the new implementation instead.
2. Add a custom metaclass with a custom `__call__`, so that if `step_ts` or `jump_ts` is passed, then in fact we'll automatically rewrap them into the new implementation. (That is, `PIDController(..., jump_ts=...)` will actually return a `JumpStepWrapper(PIDController(...), ...)` ! Python allows for some crazy things.)

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,410 @@
...
+# rejected `i_rjct -=1` and `rjct_buff[i_rjct] = t1`. Each time a step ends at
+# `t1 == rjct_buff[i_rjct]`, we increment `i_rjct` by 1 (even if the step was
+# rejected, in which case we will re-add `t1` to the stack immediately).
+# We clip the next step to `t1_next = min(t1_next, rjct_buff[i_rjct])`.
+# If `i_rjct < 0` then an error is raised.
+
+
+class JumpStepWrapper(
+    AbstractStepSizeController[_JumpStepState[_ControllerState], _Dt0]
+):
```
patrick-kidger, line 151: I think we should make this -- and the wrapped `.controller` -- specifically be `AbstractAdaptiveStepSizeController`. We do have a couple of places that check for adaptivity via `isinstance(...., AbstractAdaptiveStepSizeController)`.

(More generally this is something where in retrospect I probably shouldn't have lifted certain properties, like adaptivity, into the type system, as it makes wrapping harder. At least in this case I think it's fine, I don't really anticipate use cases for this with fixed step size controllers -- do you?)

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,410 @@
...
+            If this is > 0, then the controller will revisit rejected steps. This is
+            useful for SDEs, where the solution is guaranteed to be correct only if the
+            SDE is evaluated at all times at which the Brownian motion (BM) is
+            evaluated. Since the BM is also evaluated at rejected steps, we must later
+            evaluate the SDE at these times as well.
+        """
+        self.controller = controller
+        self.step_ts = _none_or_array(step_ts)
+        self.jump_ts = _none_or_array(jump_ts)
+        self.rejected_step_buffer_len = rejected_step_buffer_len
```
patrick-kidger, line 190: Can we default this to `max_steps`, do you think? (Maybe it is set as `None`?)

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,410 @@
...
+        self.callback_on_reject = _callback_on_reject
+        self.__check_init__()
```
patrick-kidger, line 192: This isn't necessary, Equinox will call this for you.

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,410 @@
...
+
+    def wrap(self, direction: IntScalarLike):
+        step_ts = None if self.step_ts is None else self.step_ts * direction
+        jump_ts = None if self.jump_ts is None else self.jump_ts * direction
+        return eqx.tree_at(
+            lambda s: (s.step_ts, s.jump_ts),
+            self,
+            (step_ts, jump_ts),
+            is_leaf=lambda x: x is None,
+        )
```
patrick-kidger, line 210: We should call `.wrap` of the controller we are wrapping too, to give it an opportunity to do anything else it wants to do.

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,410 @@
...
+        dt_proposal = t1 - t0
+        tdtype = jnp.result_type(t0, t1)
+
+        if self.step_ts is not None:
+            # Upcast step_ts to the same dtype as t0, t1
+            step_ts = upcast_or_raise(
+                self.step_ts,
+                jnp.zeros((), tdtype),
+                "`PIDController.step_ts`",
```
patrick-kidger, line 234: We should change the name `PIDController` here.

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,410 @@
...
+                "time (the result type of `t0`, `t1`, `dt0`, `SaveAt(ts=...)` etc.)",
+            )
+        else:
+            step_ts = None
+
+        if self.jump_ts is not None:
```
patrick-kidger, line 240: Nit: remove the `not` and flip the branches. We can save four characters ;)

(By which I mean, I find the logic easier to follow.)

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,410 @@
...
+                self.jump_ts,
+                jnp.zeros((), tdtype),
+                "`PIDController.jump_ts`",
+                "time (the result type of `t0`, `t1`, `dt0`, `SaveAt(ts=...)` etc.)",
+            )
+        else:
+            jump_ts = None
+
+        if self.rejected_step_buffer_len > 0:
+            rjct_buff = jnp.zeros(
```
patrick-kidger, line 252: Can we spell this out? I try to avoid less-readable abbreviated variable names.

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,410 @@
...
+
+            # If the step was rejected, then we need to store the rejected time in the
+            # rejected buffer and decrement the rejected index.
+            i_rjct = jnp.where(keep_step, i_rjct, i_rjct - 1)
+            i_rjct = eqx.error_if(
+                i_rjct,
+                i_rjct < 0,
+                "Maximum number of rejected steps reached. "
+                "Consider increasing JumpStepWrapper.rejected_step_buffer_len.",
+            )
```
patrick-kidger, line 357: And in particular we should be able to skip this check if we set this value to `max_steps`.

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,410 @@
...
+            rjct_buff = jnp.where(keep_step, rjct_buff, rjct_buff.at[i_rjct].set(t1))
```
patrick-kidger, line 358: Note that this is a very expensive way to describe this operation! You're copying the whole buffer. XLA will *sometimes* optimize this out -- because I added that optimization to it! -- but not always.

Better is to do `rjct_buff.at[i_rjct].set(jnp.where(keep_step, rjct_buff[i_rjct], t1))`

patrick-kidger: Other than that, I think we may need to extend the API here slightly -- we should be able to mark state like this as being a buffer for the purposes of:

https://github.com/patrick-kidger/diffrax/blob/0679807bf7bda0de3884a78cf20ce2dd4d8ade4c/diffrax/_integrate.py#L621

which is needed to avoid spurious copies during backpropagation.

(You can see that both of these comments are basically us having to work around limitations of the XLA compiler.)

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,410 @@
...
+            # location altogether.
+            jump_keep = made_jump & keep_step
+            next_t0 = static_select(
+                jump_keep, eqxi.nextafter(eqxi.nextafter(next_t0)), next_t0
+            )
+
+        if TYPE_CHECKING:
+            assert isinstance(
+                next_t0, RealScalarLike
+            ), f"type(next_t0) = {type(next_t0)}"
```
patrick-kidger, line 389: I think we should be able to do this check even outside of type checking?

diffrax/_step_size_controller/adaptive.py
andyElking: Yes, I was just about to ask your advice on how to achieve the second point 😁.
Will do!

diffrax/_step_size_controller/jump_step_wrapper.py
andyElking: I agree, probably making it adaptive-only makes sense.

diffrax/_step_size_controller/jump_step_wrapper.py
andyElking: In practice it rarely needs to be more than 10, so `max_steps` would be way overkill in my opinion. But I agree, using `None` to turn it off is more transparent.

diffrax/_step_size_controller/jump_step_wrapper.py
andyElking: Oh, thanks. This was actually added by copilot and I wasn't sure, so I left it 😅.

diffrax/_step_size_controller/jump_step_wrapper.py
andyElking: Oh yes of course, completely forgot about that.

diffrax/_step_size_controller/jump_step_wrapper.py
andyElking: Haha oops

diffrax/_step_size_controller/jump_step_wrapper.py
andyElking: Fair!

diffrax/_step_size_controller/jump_step_wrapper.py
andyElking: I'll run some timing tests to see if a long buffer significantly hinders performance. If not, I'm happy to change it to `max_steps`.

diffrax/_step_size_controller/jump_step_wrapper.py
andyElking: The first one makes sense, I should have seen this.

I don't really know what you want me to do in your second comment. And frankly `diffeqsolve` is something I haven't even started digesting yet. Are you telling me `rejected_buffer` should be one of the `outer_buffers`, meaning that I should make it an instance of `SaveState` or sth like that? I would apprecaite a bit more guidance.

Also damn how you managed to write all this code is beyond me. Even trying to begin understanding it seems a lot!

diffrax/_step_size_controller/jump_step_wrapper.py
andyElking: I get this:
```
AssertionError: type(next_t0) = <class 'jax._src.interpreters.partial_eval.DynamicJaxprTracer'>
```

diffrax/_step_size_controller/jump_step_wrapper.py
andyElking: So it seems that using a long buffer doesn't impact performance at all. So I think it could be fine to just make a binary parameter `revisit_rejected` and if it is `True`, then the buffer length is `max_steps`. However when vmaping over this that could create a significant but unnecessary memory overhead, so maybe giving the option to change this buffer length might be reasonable. What do you think?

andyElking:
Thanks for the review! I made all the edits I could and I left some comments where I need guidance (no hurry though, this is not high priority for me). Also, should I get rid of `prev_dt` entirely, as you suggested in #483?

andyElking:
Working on the NeuralSDE now and I realised JumpStepWrapper can be useful to wrap a constant step controller as well, for the purpose of making it step to save_ts on top of whichever constant steps it is making. Yes it is more efficient/proper to just do a union of the original constant steps and the `save_ts` and then feed that into diffrax.StepTo. But there are situations where that could be very inconvenient to do in code, so you can just wrap your constant step controller with a `JumpStepWrapper`. And I am pretty sure the overhead should be quite minimal (with my implementation the cost per step is no longer O(log n), but just O(1)), so this isn't so bad.

This ties into the SDE interpolation business, where essentially the only good way of interpolating SDEs seems to be just making the solver step to whichever ts you want to save at. I do intend to eventually implement that, but in the meantime this is a useful workaround, I would say.

We don't really have to advertise this use case, but can nevertheless just change the type of `JumpStepWrapper.controller` from `AbstractAdaptiveStepController` to `AbstractStepController`.
What do you think?

diffrax/_step_size_controller/jump_step_wrapper.py
patrick-kidger: Ah, good point about the vmap case.

I think the parameter `revisit_rejected` could be of type `bool | int`. If `False` then don't revisit. If `True` then do revisit and use a buffer of size `max_steps`. If an integer then do revisit and use a buffer of that size. Gives an easy choice whilst also offering control if that's really necessary for some reason.

diffrax/_step_size_controller/jump_step_wrapper.py
patrick-kidger: Haha, you're too kind!

As for my second comment -- I think I've realised that I was wrong. Let me explain. Our backpropagation involves saving copies of our state in checkpoints. Let's suppose we set `RecursiveCheckpointAdjoint(checkpoints=max_steps)`, so that's `O(max_steps)` memory right? Well, not quite: our updating buffer here is potentially of length `max_steps` (as per the debate above), and we're saving a copy of it in every checkpont, so we'd actually be using `O(max_steps^2)` memory!

The simple solution to this will just be to set the size of this buffer to e.g. 100 by default, and just allow those copies to be made. And given the behaviour you have here -- in which you potentially overwrite values -- then that is actually what's necessary as well.

As for the complicated solution that I was wrong about: let's consider the case of `SaveAt(steps=True)`. This also involves a buffer of length `max_steps`, that we save into as we go along. Fortunately, this one has a useful extra property, which is that we never overwrite a value. That means we don't actually need to copy our buffer for every checkpoint! We can use a single buffer that is shared across all checkpoints, getting gradually filled in. To support this case then we actually have a special argument `eqxi.while_loop(..., buffers=...)`, to declare which elements of our loop state have this behaviour. Unfortunately that's not the case here because we do overwrite the values. (And side-note the presence of this `buffers` parameter is the reason I've not made this public API in Equinox, because the buffer-ness is completely unchecked and it's very easy to shoot yourself in the foot.)

diffrax/_step_size_controller/jump_step_wrapper.py
patrick-kidger: Oh, this is a silly known Python bug. You need `isinstance(next_to, typing.get_args(RealScalarLike))`. Python's unions have the wrong `isinstance` behaviour.

patrick-kidger:
> Also, should I get rid of prev_dt entirely, as you suggested in #483?

If it's easy to do that in a separate commit afterwards then I would say yes. A separate commit just so it's easy to revert if it turns out we were wrong about something here :D

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,424 @@
...
+        # This is just a logging utility for testing purposes
+        if self.callback_on_reject is not None:
+            jax.debug.callback(self.callback_on_reject, keep_step, t1)
+
+        # Check whether we stepped over an element of step_ts/jump_ts/rejected_buffer
+        # This is all still bookkeeping for the PREVIOUS STEP.
+        if step_ts is not None:
+            # If we stepped to `t1 == step_ts[i_step]` and kept the step, then we
+            # increment i_step and move on to the next t in step_ts.
+            step_inc_cond = keep_step & (t1 == _get_t(i_step, step_ts))
```
patrick-kidger, line 344: I don't think I'm comfortable with this `==` between floating point numbers.

More generally speaking I think there could be cases in which the times being passed here do not perfectly align with the times that the adaptive step size controller suggested on the previous step (e.g. because of further wrapping of the step size controller), so I think this kind of logic is wrong anyway. I think you need something more like the `jump_ts` branch below, where you just want to snap `i_step` to the correct value. (Nothing that the correct value should be determinable statically, we only have state here for efficiency purposes.)

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,424 @@
...
+            i_step = jnp.where(step_inc_cond, i_step + 1, i_step)
+
+        if jump_ts is not None:
+            next_jump_t = _get_t(i_jump, jump_ts)
+            jump_inc_cond = keep_step & (t1 >= eqxi.prevbefore(next_jump_t))
+            i_jump = jnp.where(jump_inc_cond, i_jump + 1, i_jump)
```
patrick-kidger, line 350: Likewise, if someone else is fiddling with the times then this seems to me like it might be fragile.

Recalling the previous implementation with its inefficient use of searchsorted. I think the robust approach here might be to write something with the same API as that, but whose implementation is just a simple linear search forwards or backwards from the current position (which is a 'hint' about where to start searching).

Most of the time that will just iterate once and be done, as here. But in the edge cases it should now do the right thing.

andyElking:
Thanks for the review, Patrick! I'll probably make the fixes sometime in the coming week. I am also making progress on the ML examples for the Single-seed paper, but it is slower now, due to my internship.

diffrax/_step_size_controller/jump_step_wrapper.py
andyElking: Okay, I see. Thanks for the in-depth explanation! So let's see if I understand this correctly:

- I should set the default value to be 100 and keep the type as `Optional[int]`.
- If this was not getting rewritten, then I should make it register as a buffer in the outer while loop of diffeqsolve. But, because it does get rewritten, I should not do that(??). Still, I am curious, if I did want to register it as a buffer, how would I accomplish that? Is it indeed by making it an instance of `SaveState`, or is it something else entirely?

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,424 @@
...
+    rejected_buffer: Optional[Array],
+) -> RealScalarLike:
+    if rejected_buffer is None:
+        return t1
+    _t1 = _get_t(i_reject, rejected_buffer)
+    _t1 = jnp.minimum(_t1, t1)
+    return _t1
+
+
+# EXPLANATION OF STEP_TS AND JUMP_TS
```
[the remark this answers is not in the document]
andyElking: I moved this entire explanation into the docstring. Not sure if you intended me to move it all, but I can easily move it back. I also rewrote the little `!!! warning` in `stepsize_controller.md` about what requirements need to be fulfilled in order for an adaptive SDE solve to be guaranteed to converge to the correct solution.

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,424 @@
...
+    prev_dt: RealScalarLike
+    step_index: IntScalarLike
+    jump_index: IntScalarLike
+    rejected_index: IntScalarLike
+    rejected_buffer: Optional[Array]
+    step_ts: Optional[Array]
+    jump_ts: Optional[Array]
+    inner_state: _ControllerState
+
+    def get(self):
```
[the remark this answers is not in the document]
andyElking: That makes sense. I did it mostly for convenience, but it's better to stick with the good practice.

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,424 @@
...
+            jump_ts,
+            inner_state,
+        ) = controller_state.get()
+
+        # Let the controller do its thing
+        (
+            keep_step,
+            next_t0,
+            next_t1,
+            _,
```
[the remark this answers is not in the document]
andyElking: I've given this a hard think and I think it is nearly impossible to do this perfectly. We can maintain correctness of the DE solution by allowing `made_jump` to sometimes be a false positive (e.g. the inner JSW recorded `jump_next_step=True` in the previous step, but then the outer JSW further clipped the proposal, so the jump didn't actually happen). That just makes us do one unnecessary VF evaluation at the fanthom jump point, but if I understand correctly the final solution should still be correct.

andyElking:
I am very confused about what the correct value of `made_jump` should be when the step was rejected. By my understanding the original PID controller also got this wrong. However, I'm not sure whether getting this wrong is actually a significant issue, so I want your thoughts on it.

Suppose there is a jump at t=2. I will present 2 possible scenarios, in both of which I think something goes wrong (although maybe diffeqsolve might correct for the issue in scenario B). I wrote them as if JSW and the controller are separate, but the same holds for just the old PID controller.

====== scenario A =======

1. We start with a step from `t0 = 0` to `t1 = 1`, the controller decides the step will be kept and makes a proposal `next_t0 = 1`, `next_t1=2.5` which gets clipped to `next_t1 = 2` and so the JSW writes `jump_next_step=True` in its state.
2. We take the next step from `t0 = 1` to `t1 = 2`. Suppose the step is rejected, so `next_t0 = 1`. But the way the code works now, it will still set `made_jump = previous_state.jump_next_step`, which has been set to `True`. But that is wrong because `made_jump` should reflect whether there is a jump at `next_t0` so that the solver knows whether to reevaluate the VF at the start of the next step. Still this is not horrible, since it just causes one extra evaluation of the VF, but the full solution should still be correct. Right?

====== scenario B =======

1. We start with a step from `t0 = 0` to `t1 = 1`, the controller decides the step will be kept and makes a proposal `next_t0 = 1`, `next_t1=2.5` which gets clipped to `next_t1 = 2` and so the JSW writes `jump_next_step=True` in its state.
2. We take the next step from `t0 = 1` to `t1 = 2`. Suppose the step is accepted and the controller proposes `next_t0 = 2` and `next_t1 = 4` the controller returns (correctly) `made_jump = previous_state.jump_next_step (=True)` and `ControllerState(..., jump_next_step=False)`.
3. We take the next step from `t0 = 2` to `t1 = 4`. Note that because `made_jump` was set to `True` last iteration, the FSAL VF was reevaluated at `t = 2`. Suppose the step is rejected and the new proposal is `next_t0 = 2` and `next_t1 = 3`. The controller returns  `made_jump = previous_state.jump_next_step (=False)`. This is a problem because there is a jump at `next_t0=2`, so `made_jump` should have been `True`. However that might be fine if the value of the VF which was recomputed at the start of this step is kept for the next step (but I remember thinking that solver state is discarded when the step is rejected, so the recomputed VF also gets discarded).


Another way of seeing this all is through this:

- `jump_next_step = (there is a jump at next_t1)`
- `previous_state.jump_next_step = (there is a jump at t1)`
- `made_jump` should be `True` iff there is a jump at `next_t0`

Hence setting `made_jump = previous_state.jump_next_step` only works when the current step is being accepted, so `next_t0 = t1`. When rejecting the step, the new `made_jump` should be equal to the previous step's `made_jump` which we no longer have access to. The solution would be to add yet another thing to the state, but first I wanted to confirm with you to see if I'm getting this right. What are your thoughts?

patrick-kidger:
So I think the `made_jump`-on-rejected-step is handled through this line, outside the stepsize controller:

https://github.com/patrick-kidger/diffrax/blob/daec89c8fc9e72e4c2b32def2527a96b80f18b9d/diffrax/_integrate.py#L390

I made the decision to handle some of the step-rejection logic in the main `integrate.py` loop, on the basis that for those pieces it should be the same for all stepsize controllers.

So I *think* this fine? Do double-check my logic though! :p

Other than that, one thing I am noticing is that this `next_made_jump` business is pretty annoying to deal with in the stepsize controller, and I think could probably also be factored out to happen in the main `integrate.py` loop.

andyElking:
Great, that's exactly the line I was looking for (I must admit I looked in `_integrate.py` only briefly). But yes this is precisely the piece of the puzzle that my logic was missing, so I think we both independently arrived at the same conclustion that this is probably correct.

Thinking about it now, the `next_made_jump` situation might be way easier than I thought it was. I'll try to write the code and then I'll write a proof that it does the right thing for any number of stacked JSWs.

diffrax/_step_size_controller/jump_step_wrapper.py
andyElking: I implemented a linear search as you suggest and now use it to determine `i_step` and `i_jump`.

However, I think that for the rejected step business using linear search is incorrect, because `t1` should never be greater than `rejected_t = _get_t(i_reject, rejected_buffer)`. So depending on what you think is more appropriate we can either use `==` or  `jnp.isclose(t1, rejected_t, atol=1e-12)`. I also added the code below, but I can remove it if you don't think it is necessary. I could also add a parameter that only activates this callback when we are doing tests. Let me know.

``` python
i_reject = eqx.error_if(
    i_reject,
    t1 > rejected_t,
    "Jumped over a rejected time. Please report this as a bug.",
)
```

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,424 @@
...
+def _none_or_array(x):
+    if x is None:
+        return None
+    else:
+        return jnp.asarray(x)
+
+
+def _get_t(i: IntScalarLike, ts: Array) -> RealScalarLike:
+    i_min_len = jnp.minimum(i, len(ts) - 1)
+    return jnp.where(i == len(ts), jnp.inf, ts[i_min_len])
```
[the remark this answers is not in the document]
andyElking: For each of the tests (except backprop) in `test_adaptive_stepsize_controller.py` I added `@pytest.mark.parametrize("backwards", [False, True])` which just swaps the order of `t0` and `t1`. This also revealed two other problems:

1.  `step_ts` and `jump_ts` need to be sorted and in particular need to be re-sorted after multiplying them with `direction` in `wrap`. I added sorting both in `JSW.__init__` and `JSW.wrap`.
2. VBT complains if `t0 >= t1`. I don't think this is necessarily a problem but it could be confusing to some users, so let me know if you want to revisit that design decision and I can try to do something about it.

But other than that it all seems to work perfectly when solving backwards. To be fair the way you dealt with the whole backwards solve business is very clean and makes this work perfectly without any alterations.

diffrax/_step_size_controller/pid.py
```diff
@@ -600,10 +542,12 @@ def _scale(_y0, _y1_candidate, _y_error):
         factor2 = 1 if _zero_coeff(coeff2) else prev_inv_scaled_error**coeff2
         factor3 = 1 if _zero_coeff(coeff3) else prev_prev_inv_scaled_error**coeff3
         factormin = jnp.where(keep_step, 1, self.factormin)
+        # If the step is not kept, next step must be smaller, so factor must be <1.
+        factormax = jnp.where(keep_step, self.factormax, self.safety)
```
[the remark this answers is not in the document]
andyElking: Just pushed a rebased version of all this, where these changes are in separate commits.

diffrax/_step_size_controller/jump_step_wrapper.py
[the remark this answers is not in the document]
andyElking: I handled this all inside the JSW in the new commit and I wrote a long comment about why I think the implementation is correct. But please do double check it.

andyElking:
Hi Patrick!

I just pushed a new version of this PR, rebased on top of the most current main. I think I addressed everything you asked me to fix.

As it stands this contains 3 commits, contatining:

1. Everything except 2. and 3.
2. The changes to `at_dtmin` and `factormax` in `pid.py`
3. Removing `prev_dt` from JSW.

I left some conversations unresloved. I did try to fix the things mentioned in those, but I am not sure whether what I did was the best way to tackle that so I wanted to hear your opinion.

Also the test are failing because pyright doesn't know how to import `typeguard`, which has nothing to do with my changes.

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,463 @@
...
+        # next_t1 \in (next_t0, original_next_t1) and hence there cannot be a jump
+        # in inner_jump_ts at next_t1. So the jump_at_next_t1 only depends on
+        # jump_at_next_t1.
+        # On the other hand if original_next_t1 == next_t1, then we just take an
+        # OR of the two.
+        jump_at_next_t1 = jnp.where(
+            next_t1 == original_next_t1,
+            jump_at_original_next_t1,
+            jump_at_next_t1 | jump_at_original_next_t1,
+        )
```
patrick-kidger, line 444: Hmm, I don't think I completely believe this. Can we have the following:
- the PID controller proposes `t1`.
- the inner JSW wants to clip to a jump `b < t1`.
- the outer JSW wants to clip to a step (not a jump!) `a < b`
?

In this case then we will have `next_t1 != original_next_t1`, an `jump_at_original_next_t1 == True`... but overall we want `made_jump == False`?

patrick-kidger: +can we have a test for two tested JSW, including the above scenario? It doesn't need to be a full diffeqsolve, just directly calling `adapt_step_size` and checking that we get the right output.

patrick-kidger (commented):
Okay! I think I really like this.

First of all, I think I'm basically happy with pretty much everything outside of `jump_step_wrapper.py`. The changes here are pleasingly simple ^^

For `jump_step_wrapper.py`, I think my main question is around whether the rejected-step-buffer should actually be part of this wrapper at all -- since that handles SDEs with any kind of step rejection, which I think is completely orthogonal to clipping steps? (Not sure how I didn't notice this before!) I've also commented on a few other more minor points.

By the way, what did you think of the idea of moving `next_made_jump` into `_integrate.py`? It doesn't have to be now -- happy for that to be a separate PR -- just checking your thoughts on whether it is a generalisable thing.

Finally: merry Christmas, and a happy new year! :D

diffrax/_step_size_controller/jump_step_wrapper.py
patrick-kidger: FWIW if you want `error_if` here then I'd make it a testing-only branch. During regular user runtime I try to avoid using it for these kinds of asserts as it's actually quite slow.

diffrax/_step_size_controller/jump_step_wrapper.py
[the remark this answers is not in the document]
patrick-kidger: Okay, great! Then I'm very glad to have these tests :D 

For VBT I think I'm happy with either approach -- autoswitching `t0`, `t1` or just raising an error. No strong feelings on what is the better UX.

diffrax/_step_size_controller/jump_step_wrapper.py
andyElking: I think I bungled the code completely, because it doesn't do what's written in the comment at all. But I believe the comment (if implemented correctly) is correct.

I think the code should be:

```
jump_at_next_t1 = jnp.where(
    next_t1 == original_next_t1,
    jump_at_next_t1 | jump_at_original_next_t1,
    jump_at_next_t1,
)
```

And this indeed works in the case you brought up as well.

Given this, I don't think moving `next_made_jump` into integrate is strictly necessary at this point.

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,463 @@
...
+    Y,
+)
+from .._misc import static_select, upcast_or_raise
+from .._solution import RESULTS
+from .._term import AbstractTerm
+from .base import AbstractAdaptiveStepSizeController, AbstractStepSizeController
+
+
+_ControllerState = TypeVar("_ControllerState")
+_Dt0 = TypeVar("_Dt0", None, RealScalarLike, Optional[RealScalarLike])
```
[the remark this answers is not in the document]
andyElking: Oh, I see, thanks for this explanation! I remember getting confused about this before, and I think I looked it up more than once, but I never got such a good explanation!

diffrax/_step_size_controller/pid.py
```diff
@@ -576,7 +510,9 @@ def _scale(_y0, _y1_candidate, _y_error):
 
         scaled_error = self.norm(jtu.tree_map(_scale, y0, y1_candidate, y_error))
         keep_step = scaled_error < 1
+        # Automatically keep the step if we're at dtmin.
         if self.dtmin is not None:
+            at_dtmin = at_dtmin | (prev_dt <= self.dtmin)
             keep_step = keep_step | at_dtmin
```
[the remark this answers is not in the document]
andyElking: Yes, good point. Looking at the code, I don't really see a reason to keep it, but I'll get rid of it in a separate commit just to be sure.

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,463 @@
...
+    Explanation of `step_ts` and `jump_ts`:
+
+    The `step_ts` and `jump_ts` are used to force the solver to step to certain times.
+    They mostly act in the same way, except that when we hit an element of `jump_ts`,
+    the controller must return `made_jump = True`, so that the diffeqsolve function
+    knows that the vector field has a discontinuity at that point, in which case it
+    re-evaluates it right after the jump point. In addition, the
+    exact time of the jump will be skipped using eqxi.prevbefore and eqxi.nextafter.
+    So now to the explanation of the two (we will use `step_ts` as an example, but the
+    same applies to `jump_ts`):
```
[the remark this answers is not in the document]
andyElking: Yes, I intended to keep that as a comment, not part of the docs, but you suggested I put it in the docstring and I wasn't sure what exactly you wanted. I don't have strong opinions here, so feel free to rewrite it however you wish.

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,463 @@
...
+    `rejected_step_buffer_len` and a counter `i_reject`. The "stack" are all the items
+    in `rejected_buffer[i_reject:]` with `rejected_buffer[i_reject]` being the top of
+    the stack.
+    When `i_reject == rejected_step_buffer_len`, the stack is empty.
+    At the start of the run, `i_reject = rejected_step_buffer_len`. Each time a step is
+    rejected `i_reject -=1` and `rejected_buffer[i_reject] = t1`. Each time a step ends
+    at `t1 == rejected_buffer[i_reject]`, we increment `i_reject` by 1 (even if the
+    step was rejected, in which case we will re-add `t1` to the stack immediately).
+    We clip the next step to `t1_next = min(t1_next, rejected_buffer[i_reject])`.
+    If `i_reject < 0` then an error is raised.
```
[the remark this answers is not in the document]
andyElking: Hmm I see what you mean in the first comment. So essentially as an attribute that user would have to manually set to `True`? That could probably work. Then again, we did put up several warnings and beyond that we do have to assume the user isn't stupid anyway. But if you would like me to implement such a thing, I can try.

For answer to second comment see my more in-depth response below.

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,463 @@
...
+
+    def cond_up(_i):
+        return (_i < len(ts)) & (ts[_i] <= t)
+
+    def cond_down(_i):
+        return (_i > 0) & (ts[_i - 1] > t)
+
+    i = hint
+    i = jax.lax.while_loop(cond_up, lambda _i: _i + 1, i)
+    i = jax.lax.while_loop(cond_down, lambda _i: _i - 1, i)
```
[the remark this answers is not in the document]
andyElking: At most one of the two loops will trigger, so I am pretty sure doing it this way is faster and cleaner than your second suggestion (unless I'm getting it completely wrong??). And this is probably the safest option, but yeah I think we can easily just have the upwards loop, everything should still work if my logic is correct. Well in fact everything worked perfectly without any loops at all, the reason we added this is to be extra sure there aren't any edge cases. So I'd say if we want to be safe, let's be completely safe and have both loops. In my experiments the slowdown from this is relatively small.

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,463 @@
...
+            inner_state,
+            result,
+        ) = self.controller.adapt_step_size(
+            t0, t1, y0, y1_candidate, args, y_error, error_order, st.inner_state
+        )
+        next_t1 = original_next_t1
+
+        # This is just a logging utility for testing purposes
+        if self.callback_on_reject is not None:
+            jax.debug.callback(self.callback_on_reject, keep_step, t1)
```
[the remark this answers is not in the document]
andyElking: Sounds good! I don't think the order matters in the test I use this for, but I suppose might as well do it properly.

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,463 @@
...
+            next_t0 = static_select(
+                jump_keep, eqxi.nextafter(eqxi.nextafter(next_t0)), next_t0
+            )
+
+        if TYPE_CHECKING:  # if i don't seperate this out pyright complains
+            assert isinstance(
+                next_t0, RealScalarLike
+            ), f"type(next_t0) = {type(next_t0)}"
+        else:
+            isinstance(next_t0, get_args(RealScalarLike))
```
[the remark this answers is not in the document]
andyElking: Haha oops

diffrax/_step_size_controller/jump_step_wrapper.py
```diff
@@ -0,0 +1,463 @@
...
+            # location altogether.
+            jump_keep = st.jump_at_next_t1 & keep_step
+            next_t0 = static_select(
+                jump_keep, eqxi.nextafter(eqxi.nextafter(next_t0)), next_t0
+            )
+
+        if TYPE_CHECKING:  # if i don't seperate this out pyright complains
+            assert isinstance(
+                next_t0, RealScalarLike
+            ), f"type(next_t0) = {type(next_t0)}"
```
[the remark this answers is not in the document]
andyElking: Oh, I see.

diffrax/_step_size_controller/jump_step_wrapper.py
andyElking: Hmm, yes I'll add a test.

andyElking:
Hi Patrick!

Sorry for the long silence, the last few weeks have been very busy.

I made the corrections you suggested.

Regarding splitting off revisiting steps from `JumpStepWrapper`, I agree that this feature is perpendicular to `jump_ts` and `step_ts`, but I don't see the harm in implementing several different features with one class... Certainly we want step-revisiting to be part of some sort of wrapper which can wrap either PID or potentially a custom step controller (I had such a situation in the paper with James I think), so it shouldn't be part of PID either.
So then the alternative is to have two wrappers -- `JumpStepWrapper` and `RevisitingWrapper`  -- which might work, but might also introduce more confusion and more potential edge cases.

So unless you feel very strongly about separating these two features, I would prefer not to add extra hurdles to this PR and conclude it in the near future. I think Owen has been itching to get this done as fast as possible as well.

diffrax/_step_size_controller/jump_step_wrapper.py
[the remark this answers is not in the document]
patrick-kidger: > Hmm I see what you mean in the first comment. So essentially as an attribute that user would have to manually set to True?

Yup exactly. Just as an extra preventative footgun.

patrick-kidger:
> So unless you feel very strongly about separating these two features, I would prefer not to add extra hurdles to this PR and conclude it in the near future. I think Owen has been itching to get this done as fast as possible as well.

Leaving it here on that basis sounds totally reasonable to me. Let's just tweak the name from `JumpStepWrapper` to something else to reflect this though.

test/test_adaptive_stepsize_controller.py
```diff
@@ -65,6 +75,7 @@ def run(**kwargs):
     sol_no_jump_ts = run()
     sol_with_jump_ts = run(jump_ts=[7.5])
     assert sol_no_jump_ts.stats["num_steps"] > sol_with_jump_ts.stats["num_steps"]
+    print(sol_no_jump_ts.stats["num_steps"], sol_with_jump_ts.stats["num_steps"])
```
patrick-kidger, line 78: !

patrick-kidger (commented):
Okay, I started reviewing this... and now it's way too late in the evening for me to parse the math on this! :D 
The one big thing I want to check for is that the numerics in `JumpStepWrapper` are doing the correct thing. I'll do that tomorrow and then let's aim to merge this. :)

patrick-kidger:
Okay @andyElking, I've finished going over this PR and I've pushed changes back to this branch in one commit. Changes are:

- Renamed `JumpStepWrapper` to `ClipStepSizeController` to reflect the three different kinds of clipping it does.
- Simplify the internals such that `st.jump_at_next_t1` does not need to be stored: instead I examine the previous times we were stepping to. I think this should be much simpler to reason about! (The previous implementation here / in `PIDController` was always something I found pretty mind-bending.) More pragmatically, I'm more certain that this will do the right thing when nesting multiple wrappers inside each other.
- Switched from using `eqx.error_if` to returning an error code, which should be more efficient.
- Adjusted documentation and added your paper to `diffrax.citation` :)

Lmk what you think!

andyElking:
I read your changes and it all LGTM. Indeed the trick of just checking equality over all `ts` in `_bump_next_t0` is very inventive and much cleaner! You really made the whole code waaay simpler!

Oh, I never really thought of using the `RESULT` business to report errors. To be fair, now it makes much more sense to me why it's there in the first place.

I completely agree with the renaming as well.

Thanks for adding the citation!

Also I find the `_set_metaclass = dict(metaclass=_MetaPID)` trick pretty funny. Crazy what sorts of workarounds are sometimes needed in python 😁

Thanks for finishing this up!

patrick-kidger:
Awesome. In that case: now merged :) Thank you for your efforts driving this one!

PR #496, opened 2024-08-20
Adding weak levy area

lockwo:
We don't add this to the VBT because we don't know how to (to quote James "Perhaps some fancy coupling is possible here, but it's probably quite difficult."). But one thing we are still thinking about was the best approach to generalize this to non vector Wiener processes. Because we need this antisymmetric matrix, but we can't do b + b.T necessarily. Maybe the solver is ill-defined for Weiner processes greater than 1D @frankschae


patrick-kidger:
If you want Davie approximations to Levy area then I think we had that in torchsde. It does exist!

lockwo:
> If you want Davie approximations to Levy area then I think we had that in torchsde. It does exist!

This torchsde package sounds pretty neat, I might have to check it out ;).

But yea, this is basically the same Levy area as in torchsde (well, it's a work in progress). The difficulty seems to come in for weak schemes that have non-previsible adaptive steps, then (per James) sampling A is hard.

PR #498, opened 2024-08-23
Add Stochastic Theta method

lockwo:
cc: @frankschae

### DIFF
--- a/diffrax/__init__.py
+++ b/diffrax/__init__.py
@@ -110,6 +110,7 @@ from ._solver import (
     SPaRK as SPaRK,
     SRA1 as SRA1,
     StochasticButcherTableau as StochasticButcherTableau,
+    StochasticTheta as StochasticTheta,
     StratonovichMilstein as StratonovichMilstein,
     Tsit5 as Tsit5,
 )
--- a/diffrax/_solver/__init__.py
+++ b/diffrax/_solver/__init__.py
@@ -50,4 +50,5 @@ from .srk import (
     AbstractSRK as AbstractSRK,
     StochasticButcherTableau as StochasticButcherTableau,
 )
+from .stochastic_theta import StochasticTheta as StochasticTheta
 from .tsit5 import Tsit5 as Tsit5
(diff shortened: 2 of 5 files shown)


patrick-kidger:
I don't think I'm completely comfortable with this solver!

The main point is adaptivity -- famously, adaptive Euler fails to converge to the Ito solution. I think this is opening a similar footgun.

Naming-wise I think this is also easy to muddle up with the other θ-method, that converges to an arbitrary point between or beyond the Ito and Stratonovich solutions.

On balance I think this has too many footguns for inclusion in Diffrax itself. But thank you anyway!

frankschae:
> Naming-wise I think this is also easy to muddle up with the other θ-method, that converges to an arbitrary point between or beyond the Ito and Stratonovich solutions.

One easy way to deal with this would be to implement only the two most common cases (trapezoidal rule for θ=0.5 and drift-implicit Euler rule θ=1) instead of allowing every θ in [0, 1]. I think both of these solvers would be good additions due to the implicit treatment of the drift, see https://epubs.siam.org/doi/abs/10.1137/S003614299834736X for a stability analysis.

> The main point is adaptivity -- famously, adaptive Euler fails to converge to the Ito solution. I think this is opening a similar footgun.

Good point. Is there a good way to make this non-adaptive while allowing a user to define an error estimate and controller if they want to? In particular, we actually want to support/implement the following weak adaptive scheme for SDEs with small noise:

https://epubs.siam.org/doi/abs/10.1137/030601429

which is based on the stochastic θ method. Could just hardcode the error estimate/controller to follow this scheme and write explicitly that this is meant for weak solutions of SDEs with small noise?

lockwo:
> One easy way to deal with this would be to implement only the two most common cases (trapezoidal rule for θ=0.5 and drift-implicit Euler rule for θ=1)

Very true, we can reduce the subspace of theta if we need to. Drift implicit is low hanging fruit (given there is already implicit).

>Good point. Is there a good way to make this non-adaptive while allowing a user to define an error estimate and controller if they want to? In particular, we actually want to support/implement the following weak adaptive scheme for SDEs with small noise:

This was with the end goal, so we can also split this into Drift implicit with no error estimate (and 0.5 theta), and the RW scheme which has a error estimate in a separate PR

patrick-kidger:
I think for now I'm going to say this is out-of-scope for Diffrax, I'm afraid. I think the details here are subtle enough that I don't think I can commit to supporting them going forward. :)

PR #515, opened 2024-10-18
small fix of docs in all three and a return type in quicsort

andyElking:
Noticed a few little issues while I was implementing UBU (for some unrelated work).

### DIFF
--- a/diffrax/_solver/align.py
+++ b/diffrax/_solver/align.py
@@ -46,8 +46,9 @@ _ErrorEstimate = UnderdampedLangevinTuple
 class ALIGN(AbstractFosterLangevinSRK[_ALIGNCoeffs, _ErrorEstimate]):
     r"""The Adaptive Langevin via Interpolated Gradients and Noise method
     designed by James Foster. This is a second order solver for the
-    Underdamped Langevin Diffusion, the terms for which can be created using
-    [`diffrax.make_underdamped_langevin_term`][]. Uses two evaluations of the vector
+    Underdamped Langevin Diffusion, and accepts terms of the form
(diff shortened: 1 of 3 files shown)


patrick-kidger:
LGTM!

I was wondering why these didn't trigger an error in our documentation-building process, and it looks like we haven't actually included these solvers in the documentation at all! I'd missed that that was the case -- probably they are meant to be there?

andyElking:
I think they are present in the docs, at the bottom of sde_solvers.md. Maybe you don't build the docs on the dev branch?? To be fair I was very surprised this went unnoticed, since I did build the docs each time before committing (after you got me used to doing that). Not sure...

patrick-kidger:
Ah, right! I was indeed just looking on the wrong branch, whoops.

I should probably do a new release -- once this PR is in I'll do so.

(And I'm aware thte the `JumpStepWrapper` implementation still needs my attention... I've been meaning to find some time to get back to that 😅)

andyElking:
No hurry with the `JumpStepWrapper`, I am done with what I needed it for – but of course I am happy to make edits once you review it. And also no hurry with a new release, as far as I'm concerned 😊.

I just saw that `test_implicit_closure_convert` failed for some reason, but I am confident that has nothing to do with my edits 😅. Also it passes fine when I run it locally.

patrick-kidger:
Yup, this is just JAX 0.4.34 doing its thing and breaking us all over the show :D I've now got PRs out for both Optimistix and Lineax:

https://github.com/patrick-kidger/optimistix/pull/87
https://github.com/patrick-kidger/lineax/pull/111

andyElking:
Oh, I see. I don't envy you for having to fix that.

PR #524, opened 2024-11-13
check-and-clip linear_rescale to [0,1]

mattlevine22:
Protect against floating point errors that sometimes occur on GPUs (https://github.com/jax-ml/jax/issues/24807). Previously observed that linear_rescale was sometimes returning 1.000001 for `t==t1`, and this was creating an error in the `evaluate` method of `VirtualBrownianTree` during SDE solves on GPU. Did not observe issue on CPUs.

### DIFF
--- a/diffrax/_misc.py
+++ b/diffrax/_misc.py
@@ -73,17 +73,22 @@ def fill_forward(
 
 
 def linear_rescale(t0, t, t1) -> Array:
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Closing as superseded by #526! Thanks @mattlevine22 for putting this one together :)

PR #528, opened 2024-11-19
Reversible Solvers

sammccallum:
Hey Patrick,

Here's an implementation of Reversible Solvers! This includes:

1. make any `AbstractRungeKutta` method in diffrax algebraically reversible - see `diffrax.Reversible`
2. backpropagate through the solve in constant memory and get exact gradients (up to floating point errors) - see `diffrax.ReversibleAdjoint`

Main details I should highlight here:

1. The current implementation relies on the _SolverState type of `AbstractRungeKutta` methods. Specifically, as the reversible method switches between evaluating the vector field at y and z, we ensure the `fsal` is correct by evaluating the vector field outside of the base Runge Kutta step. In principle this is unnecessary but required to fit with the behaviour of `AbstractRungeKutta` solvers; any ideas for how to avoid this?

3. To backpropagate through the reversible solve we require knowledge of the `ts` that the solver visited. As this is not known a priori for adaptive step sizes, I've added a (teeny weeny) bit of infrastructure to the State in `_integrate.py`. This allows us to save the ts that the solver stepped to which we make available to `ReversibleAdjoint` as a residual. The added State follows exactly the implementation of saving dense_ts and is only triggered when `adjoint=ReversibleAdjoint`.

Best,
Sam


diffrax/_solver/reversible.py
```diff
@@ -0,0 +1,114 @@
...
+    The convergence order of the reversible solver is inherited from the wrapped
+    Runge-Kutta method.
+
+    Backpropagation through the reversible solver implies very low memory usage and
+    exact gradient calculation (up to floating point errors). This is implemented in
+    [`diffrax.ReversibleAdjoint`][] and passed to [`diffrax.diffeqsolve`][] as
+    `adjoint=diffrax.ReversibleAdjoint()`.
+    """
+
+    solver: AbstractRungeKutta
```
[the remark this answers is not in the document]
patrick-kidger: Also what about `Euler`, which isn't implemented as an `AbstractRungeKutta` but which does have the correct properties?

(It's done separately to be able to use as example code for how to write a solver.)

diffrax/_adjoint.py
```diff
@@ -852,3 +857,206 @@ def loop(
...
+        passed_controller_state,
+        event,
+        **kwargs,
+    ):
+        # `is` check because this may return a Tracer from SaveAt(ts=<array>)
+        if eqx.tree_equal(saveat, SaveAt(t1=True)) is not True:
+            raise ValueError(
+                "Can only use `adjoint=ReversibleAdjoint()` with "
+                "`saveat=SaveAt(t1=True)`."
+            )
```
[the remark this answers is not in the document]
patrick-kidger: FWIW I imagine `SaveAt(steps=True)` is probably much easier.

diffrax/_adjoint.py
```diff
@@ -852,3 +857,206 @@ def loop(
...
+
+        if not isinstance(solver, Reversible):
+            raise ValueError(
+                "Can only use `adjoint=ReversibleAdjoint()` with "
+                "`Reversible()` solver."
+            )
```
patrick-kidger, line 1044: Could we perhaps remove `Reversible` from the public API altogether, and just have `solver = Reversible(solver)` here? Make the `Reversible` solver an implementation detail of the adjoint.

diffrax/_solver/reversible.py
```diff
@@ -0,0 +1,114 @@
...
+        self,
+        terms: PyTree[AbstractTerm],
+        t0: RealScalarLike,
+        t1: RealScalarLike,
+        y0: Y,
+        args: Args,
+        solver_state: _SolverState,
+        made_jump: BoolScalarLike,
+    ) -> tuple[Y, Optional[Y], DenseInfo, _SolverState, RESULTS]:
+        (first_step, f0), z0 = solver_state
```
patrick-kidger, line 87: This will fail for non-FSAL Runge-Kutta solvers.
(Can you add a test for one of those to be sure we get correct behaviour?)

diffrax/_solver/reversible.py
```diff
@@ -0,0 +1,114 @@
...
+        f0 = self.func(terms, t0, z0, args)
+        step_z0, z_error, dense_info, _, result1 = self.solver.step(
+            terms, t0, t1, z0, args, (first_step, f0), made_jump
+        )
+        y1 = (self.l * (ω(y0) - ω(z0)) + ω(step_z0)).ω
+
+        f0 = self.func(terms, t1, y1, args)
+        step_y1, y_error, _, _, result2 = self.solver.step(
+            terms, t1, t0, y1, args, (first_step, f0), made_jump
+        )
```
patrick-kidger, line 103: On these two evaluations of `.step` -- take a look at `eqx.internal.scan_trick`, which might allow you to collapse these two callsites into one. That can be used to half compilation time!

sammccallum:
Thanks very much for the review and suggestions!

I'll reply to individual comments inline but here is an overview:

1. I really like the idea for removing `Reversible` from the public API and just making the adjoint auto-wrap the original solver. It feels very JAX-like, as we just augment the forward-trace so that we can backpropagate with improved properties. This has been added.

2. Can now use any `AbstractSolver`, apart from solvers that are already algebraically reversible (raises a ValueError if passed).

3. Add functionality for `SaveAt(steps=True)`.

4. Improved docs so that people know what they can use ReversibleAdjoint with (any solver, ODEs/CDEs/SDEs, adaptive time steps).

5. Improved tests

diffrax/_solver/reversible.py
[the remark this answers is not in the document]
sammccallum: Yep, you're right - any single step method should work. The reversible solver now works with any `AbstractSolver`.

See the discussion on `fsal` for more info.

diffrax/_solver/reversible.py
```diff
@@ -0,0 +1,114 @@
...
+
+    @property
+    def root_find_max_steps(self):
+        return self.solver.root_find_max_steps  # pyright: ignore
+
+    def order(self, terms: PyTree[AbstractTerm]) -> Optional[int]:
+        return self.solver.order(terms)
+
+    def strong_order(self, terms: PyTree[AbstractTerm]) -> Optional[RealScalarLike]:
+        return self.solver.strong_order(terms)
```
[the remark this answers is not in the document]
sammccallum: We have no theory here (does James' intuition count?), but numerically it works for SDEs! I've added SDEs to the docstring.

There's the detail that the second solver step (that steps backwards in time) should use the same Brownian increment as the first solver step. I believe this is handled by `VirtualBrownianTree`.

diffrax/_solver/reversible.py
```diff
@@ -0,0 +1,114 @@
...
+    algebraically reversible.
+
+    The convergence order of the reversible solver is inherited from the wrapped
+    Runge-Kutta method.
+
+    Backpropagation through the reversible solver implies very low memory usage and
+    exact gradient calculation (up to floating point errors). This is implemented in
+    [`diffrax.ReversibleAdjoint`][] and passed to [`diffrax.diffeqsolve`][] as
+    `adjoint=diffrax.ReversibleAdjoint()`.
+    """
```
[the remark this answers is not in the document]
sammccallum: Thanks, good point - added. 

Removing `Reversible` from public API helps with control here.

diffrax/_adjoint.py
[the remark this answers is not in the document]
sammccallum: I've added functionality for `SaveAt(steps=True)`, but `SaveAt(ts=...)` is a tricky one.

Not a solution, but some thoughts:

The `ReversibleAdjoint` computes gradients accurate to the numerical operations taken, rather than an approximation to the 'idealised' continuous-time adjoint ODE. This is then tricky when the numerical operations include interpolation and not just ODE solving.

In principle, the interpolated `ys` are just a function of the stepped-to `ys`. We can therefore calculate gradients for the stepped-to `ys` and let AD handle the rest. This would require the interpolation routine to be separate to the solve routine, but I understand the memory drawbacks of this setup.

I imagine there isn't a huge demand to decouple the solve from the interpolation - but if it turns out this is relevant for other cases I'd be happy to give it a go!

diffrax/_adjoint.py
sammccallum: I really like this idea :D

I've removed `Reversible` from the public API and any `AbstractSolver` passed to `ReversibleAdjoint` is auto-wrapped. There is now a `_Reversible` class within the `_adjoint` module that is exclusively used by `ReversibleAdjoint`. Do you think this is an appropriate home for the `_Reversible` class or should I keep it elsewhere?

diffrax/_solver/reversible.py
sammccallum: See comment below.

diffrax/_solver/reversible.py
sammccallum: CMIIW, I'm not sure we can use the scan trick here as the function return signature is different for each solver step?

That is, we only want to update the `original_solver_state` and `dense_info` when taking the forward-in-time step. So we don't return these on the backward-in-time step. IIUC, collapsing the two calls into one would require the returned carry to be the same on both calls?

diffrax/_solver/reversible.py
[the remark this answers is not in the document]
sammccallum: Just added a check and test for `UnsafeBrownianPath` in light of the above.

And thinking about this further, we require the same conditions as `BacksolveAdjoint`, namely that the solver converges to the Stratonovich solution, so I've added a check and test for this.

diffrax/_adjoint.py
patrick-kidger: For consistency I'd probably put it in `_solvers/reversible.py` -- I've generally tended to organize things in this way, e.g. `_terms.py::AdjointTerm` is used as part of `BacksolveAdjoint`.

But that's only for consistency, one could imagine an alternate layout where these all lived next to their consumers instead.

diffrax/_solver/reversible.py
```diff
@@ -0,0 +1,114 @@
...
+        made_jump: BoolScalarLike,
+    ) -> tuple[Y, Optional[Y], DenseInfo, _SolverState, RESULTS]:
+        (first_step, f0), z0 = solver_state
+
+        # TODO The solver steps switch between evaluating from z0
+        # and y1. Therefore, we re-evaluate f0 outside of the base
+        # solver to ensure the vf is correct.
+        # Can we avoid this re-evaluation?
+
+        f0 = self.func(terms, t0, z0, args)
```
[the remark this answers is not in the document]
patrick-kidger: Awesome! It makes me very happy that these are all possible, that really simplifies our lives ^^

diffrax/_solver/reversible.py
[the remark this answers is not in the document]
patrick-kidger: Hmm FWIW my intuition is also that this should work for SDEs. Sounds like a follow-up paper to be written :)

But... until that theory exists, I think I'd feel more comfortable issuing an error here instead, to try to minimize the possibility of footguns. Most users treat solvers like oracles, and I try to cater to that unfootgunable UX!

diffrax/_solver/reversible.py
patrick-kidger: So it's usually possible to make this work even in arbitrary cases by welding two copies of your state together and then using a `jnp.where` or `lax.cond` to route between them based on which step you're on.

That said this is a pretty fiddly optimization, and I probably shouldn't have suggested it just yet! Once we're happy with everything else then we could do this later, but until then the un-scan-trick'd code is much easier to read.

diffrax/_integrate.py
```diff
@@ -1320,6 +1341,18 @@ def _outer_cond_fn(cond_fn_i):
...
+    if isinstance(adjoint, ReversibleAdjoint):
+        if max_steps is None:
+            raise ValueError(
+                "`max_steps=None` is incompatible with `ReversibleAdjoint`"
+            )
+        reversible_ts = jnp.full(max_steps + 1, jnp.inf, dtype=time_dtype)
+        reversible_save_index = 0
+    else:
+        reversible_ts = None
+        reversible_save_index = None
```
patrick-kidger, line 1354: I think what might be simpler here is to have:
```python
if max_steps is None:
    reversible_ts = None
    reversible_save_index = None
else:
    reversible_ts = jnp.full(...)
    reversible_save_index = 0
```
so that we are *always* saving this information if possible.

The benefit of this is that in principle someone else could write their own `ReversibleAdjoint2` and have it work without needing to be special-cased here inside the main diffeqsolve implementation: it would just consume the information made available to it.

Finally the `ValueError` can be moved inside the implementation of `ReversibleAdjoint`, if the necessary `reversible_ts` information is not available.

This shouldn't really impose any performance penalty (a very small compile-time one only) because for any other adjoint method it will just be DCE'd.

diffrax/_adjoint.py
```diff
@@ -852,3 +865,427 @@ def loop(
...
+                    odes},
+            author={Zhuang, Juntang and Dvornek, Nicha C and Tatikonda, Sekhar and
+            Duncan, James S},
+            journal={arXiv preprint arXiv:2102.04668},
+            year={2021}
+        }
+        ```
+    """
+
+    l: float = 0.999
```
patrick-kidger, line 1096: Can we pick a more descriptive name for this parameter? E.g. when writing an optimizer then mathematically we may conventionally use λ but in code we would often use a variable with a name like `learning_rate`.

diffrax/_adjoint.py
```diff
@@ -852,3 +865,427 @@ def loop(
...
+        # `is` check because this may return a Tracer from SaveAt(ts=<array>)
+        if (
+            eqx.tree_equal(saveat, SaveAt(t1=True)) is not True
+            and eqx.tree_equal(saveat, SaveAt(steps=True)) is not True
+            and eqx.tree_equal(saveat, SaveAt(t0=True, steps=True)) is not True
+        ):
+            raise ValueError(
+                "Can only use `diffrax.ReversibleAdjoint` with "
+                "`saveat=SaveAt(t1=True)` or `saveat=SaveAt(steps=True)`."
+            )
```
patrick-kidger, line 1120: I'm guessing that `SaveAt(steps=True, t1=True)` or `SaveAt(t0=True, steps=True, t1=True)` should also be allowed?

diffrax/_adjoint.py
```diff
@@ -852,3 +865,427 @@ def loop(
...
+        tprev = init_state.tprev
+        tnext = init_state.tnext
+        y = init_state.y
+
+        init_state = eqx.tree_at(
+            lambda s: s.solver_state,
+            init_state,
+            solver.init(terms, tprev, tnext, y, args),
+            is_leaf=_is_none,
+        )
```
patrick-kidger, line 1163: Note that the `_Reversible.init` here will re-call the underlying original `solver.init`, I think unnecessarily? I think we should be able to do just `init_state = eqx.tree_at(lambda s: s.solver_state, init_state, (init_state.solver_state, init_state.y))`, and then set `class _Reversible: def init(...): assert False`.

diffrax/_adjoint.py
```diff
@@ -852,3 +865,427 @@ def loop(
...
+    def term_structure(self):
+        return self.solver.term_structure
+
+    @property
+    def interpolation_cls(self):  # pyright: ignore
+        return self.solver.interpolation_cls
+
+    @property
+    def term_compatible_contr_kwargs(self):
+        return self.solver.term_compatible_contr_kwargs
```
patrick-kidger, line 1231: I think this, and `term_structure`, should not be necessary -- they are checked at the start of `diffeqsolve`, which we are already past. I think it may be neater to set these to `assert False`?

diffrax/_adjoint.py
```diff
@@ -852,3 +865,427 @@ def loop(
...
+        self,
+        terms: PyTree[AbstractTerm],
+        t0: RealScalarLike,
+        t1: RealScalarLike,
+        y0: Y,
+        args: Args,
+    ) -> _SolverState:
+        if isinstance(self.solver, AbstractRungeKutta):
+            object.__setattr__(self.solver.tableau, "fsal", False)
+            object.__setattr__(self.solver.tableau, "ssal", False)
```
patrick-kidger, line 1257: Note that this will produce a bug if we do:
```python
solver = Tsit5()
diffeqsolve(..., solver, ReversibleAdjoint())
diffeqsolve(..., solver, RecursiveCheckpointAdjoint())
```
as the `solver` is modified in-place.

I think I have a better solution: can we unconditionally pass `made_jump=True` into `self.solver.step`? This is our API to indicate to solvers that something has changed, and that their state may be out-of-date. Technically right now it's used to indicate jumps in the vector field, but we could re-use it (or add another flag) to indicate exogenous jumps in `y`.
Alternatively it may be safer for now to only allow `AbstractRungeKutta` here, and not general solvers -- it's not clear to me that any of this will really work with multi-step solvers like `LeapfrogMidpoint`, for example.

diffrax/_adjoint.py
```diff
@@ -852,3 +865,427 @@ def loop(
...
+
+        step_y1, y_error, _, _, result2 = self.solver.step(
+            terms, t1, t0, y1, args, original_solver_state, made_jump
+        )
+        z1 = (ω(y1) + ω(z0) - ω(step_y1)).ω
+
+        solver_state = (original_solver_state, z1)
+        result = update_result(result1, result2)
+
+        return y1, _add_maybe_none(z_error, y_error), dense_info, solver_state, result
```
patrick-kidger, line 1286: Why add the error estimates together?

diffrax/_adjoint.py
```diff
@@ -852,3 +865,427 @@ def loop(
...
+
+    diff_args = eqx.filter(args, eqx.is_inexact_array)
+    diff_terms = eqx.filter(terms, eqx.is_inexact_array)
+    diff_z1 = eqx.filter(z1, eqx.is_inexact_array)
+    grad_args = jtu.tree_map(jnp.zeros_like, diff_args)
+    grad_terms = jtu.tree_map(jnp.zeros_like, diff_terms)
+    grad_z1 = jtu.tree_map(jnp.zeros_like, diff_z1)
+    del diff_args, diff_terms, diff_z1
+
+    def grad_step(state):
```
patrick-kidger, line 963: Hmm, I'm trying to figure out how `solver_state` is handled here. I don't think it is correct?

`solver_state` is some completely arbitrary information that is propagated forward step-by-step, internal to the solver. In particular we don't have an API for reconstructing this backwards in time reversibly.

patrick-kidger (commented):
Okay, sorry for the (long) delay! 😅

So this mostly looks like it's in a good place to me. I think my one big concern is handling `solver_state` -- as each solver is free to define any kind of state it likes, then (a) there need not be any way to reconstruct it reversibly (so the backward pass goes wrong), and (b) it may encode information from previous `y`,`z`-values that aren't actually the ones being propagated (so the forward pass goes wrong).

I think right now you're dodging this just because of the particular choice of solvers being considered, i.e. (forced-to-be-)non-FSAL Runge--Kutta solvers.

I think we've got a couple of possible options:
- We could expand the `AbstractSolver` API a little bit. For example a method to construct the previous state reversibly, and a flag to indicate that the `y`-value has changed and the solver might need to consider its state invalidated.
- We could also special-case down to e.g. just `AbstractRungeKutta`, and hardwire all the things about it that we know how to handle.

I am weakly leaning towards the first option, as it could allow us to 'already reversible' solvers like `ReversibleHeun` or `LeapfrogMidpoint` (c.f. #541). Presumably we'd also need a way to signal to `ReversibleAdjoint` whether or not to wrap things in a `_Reversible` solver or whether the solver is already natively reversible. But I also realise that these are all pretty complicated research questions, and I don't want to presume upon your appetite for tackling them!

Let me know what you think / also happy to have a longer chat about this via email etc if it's easier.

sammccallum:
A description of the performance bug with the current ReversibleAdjoint:

- vmapping over diffeqsolve creates a significant slow down

From the example below, we see that for `batch_size=1` ReversibleAdjoint is faster than RecursiveCheckpointAdjoint (as expected). But for `batch_size=1000` ReversibleAdjoint is slower than RecursiveCheckpointAdjoint.



```python3
import time

import diffrax as dfx
import equinox as eqx
import jax
import jax.numpy as jnp
import jax.random as jr

class VectorField(eqx.Module):
    mlp: eqx.nn.MLP

    def __init__(self, y_dim, width_size, depth, key):
        self.mlp = eqx.nn.MLP(y_dim, y_dim, width_size, depth, key=key)

        return self.mlp(y)


@eqx.filter_jit
def solve(model, y0, adjoint):
    term = dfx.ODETerm(model)
    solver = dfx.Euler()
    t0 = 0.0
    t1 = 10.0
    dt0 = 0.01
    sol = dfx.diffeqsolve(
        term,
        solver,
        t0,
        t1,
        dt0,
        y0,
        saveat=dfx.SaveAt(t1=True),
        adjoint=adjoint,
        max_steps=1000,
    )
    return sol.ys


@eqx.filter_value_and_grad
def grad_loss(model, y0, adjoint):
    ys = eqx.filter_vmap(solve, in_axes=(None, 0, None))(model, y0, adjoint)
    return jnp.mean(ys**2)


def measure_runtime(y0, model, adjoint):
    tic = time.time()
    loss, grads = grad_loss(model, y0, adjoint)
    toc = time.time()
    print(f"Compile time: {(toc - tic):.5f}")

    repeats = 10
    tic = time.time()
    for i in range(repeats):
        loss, grads = jax.block_until_ready(grad_loss(model, y0, adjoint))
    toc = time.time()
    print(f"Runtime: {((toc - tic) / repeats):.5f}")


y_dim = 100
width_size = 100
depth = 4
model = VectorField(y_dim, width_size, depth, key=jr.PRNGKey(10))

print("Batch Size = 1")
print("--------------")
y0 = jnp.ones((1, y_dim))
print("Recursive")
adjoint = dfx.RecursiveCheckpointAdjoint()
measure_runtime(y0, model, adjoint)
print("Reversible")
adjoint = dfx.ReversibleAdjoint()
measure_runtime(y0, model, adjoint)

print("\nBatch Size = 1000")
print("-----------------")
y0 = jnp.ones((1000, y_dim))
print("Recursive")
adjoint = dfx.RecursiveCheckpointAdjoint()
measure_runtime(y0, model, adjoint)
print("Reversible")
adjoint = dfx.ReversibleAdjoint()
measure_runtime(y0, model, adjoint)
```
```
Batch Size = 1
--------------
Recursive
Compile time: 3.99825
Runtime: 0.37017
Reversible
Compile time: 1.60509
Runtime: 0.18221

Batch Size = 1000
-----------------
Recursive
Compile time: 5.35720
Runtime: 3.15009
Reversible
Compile time: 4.87897
Runtime: 3.89896
```

FWIW, I don't think this is specifically a problem with vmap, but a problem of scale. For example, if we keep `batch_size=1` but increase the `width_size` of the MLP then a similar slow down is observed.

```
Width Size = 100
--------------
Recursive
Compile time: 3.75692
Runtime: 0.33692
Reversible
Compile time: 1.63070
Runtime: 0.18080

Width Size = 1000
--------------
Recursive
Compile time: 3.18077
Runtime: 0.61647
Reversible
Compile time: 2.06301
Runtime: 0.65720
```

In principle, this is very wrong as we are only changing the cost of each step $n$. The complexity remains a comparison between $O(n)$ for Reversible and $O(n \log n)$ for RecursiveCheckpointing where $n$ is fixed. Any ideas would be massively appreciated!

(The quoted runtimes are on GPU)

patrick-kidger:
Hmm, that's rather weird! I've just taken a look over both your MWE and your code and nothing immediately jumps out at me as wrong.

A few immediate thoughts about how you might tackle this:
- Can you try doing all of this without ever explicitly calling `jax.vmap`? Just setting up multiple independent systems altogether and treating them as a single ODE solve. When we do vmap-of-loop or vmap-of-cond then there can be some performance pessimations that occur, and I notice that you're unconditionally using vmap even on the `batch_size=1` case.
- Use [a profiler](https://jax.readthedocs.io/en/latest/profiling.html). If you want an example of what debugging with that looks like, [here is a recent Equinox issue](https://github.com/patrick-kidger/equinox/issues/920) on a similar topic. FWIW for slowdowns of the nature, then it's moderately common that the underlying root cause is a buffer somewhere getting unnecessarily copied between loop iterations; something to keep an eye out for.
- Check the asymptotics of how the time varies as your problem size changes.

PR #530, opened 2024-11-27
Update getting-started.md

ddrous:
Hi @patrick-kidger,

I think there's a small typo in the Getting-Started docs on the SDE integration interval. Here's a smaller fix !

Cheers,
R. Des


patrick-kidger:
LGTM! Thank you for the fix :)

PR #537, opened 2024-12-09
Forward mode "adjoint" 

johannahaffner:
Here you go! This is the pragmatic solution, without support or test coverage for integer inputs and only a small comment explicating that forward mode is not really an adjoint, even though its diffrax interface is that of an adjoint.

Changes with respect to the last PR: 

- renamed to `ForwardMode` everywhere
- add a sentence to the docstring that explains that this is not really an adjoint, but keep inheriting from `AbstractAdjoint`
- remove stub for "forward gradient with int" from `test_adjoint.py` and explain that since JAX does not offer this option, we're not writing our own workaround to test it either

_On the last point: if I understood this correctly, then supporting this would entail writing a gradient-computation directly from a JVP with custom "unit pytrees". This is somewhat annoying for mixed array and non-array types. 
I'm happy to try again if computing gradients with respect to integer elements of a PyTree is an expected use case (maybe arising from composed/layered transformations of a solve) that requires test coverage._

Earlier comments [here](https://github.com/patrick-kidger/diffrax/pull/536#pullrequestreview-2488189344). 

(This is now rebased on main.)

### DIFF
--- a/diffrax/_adjoint.py
+++ b/diffrax/_adjoint.py
@@ -184,7 +184,9 @@ class RecursiveCheckpointAdjoint(AbstractAdjoint):
     !!! info
 
         Note that this cannot be forward-mode autodifferentiated. (E.g. using
-        `jax.jvp`.) Try using [`diffrax.DirectAdjoint`][] if that is something you need.
+        `jax.jvp`.) Try using [`diffrax.DirectAdjoint`][] if you need both forward-mode
+        and reverse-mode autodifferentiation, and [`diffrax.ForwardMode`][] if you need
+        only forward-mode autodifferentiation.
 
     ??? cite "References"
 
@@ -333,6 +335,8 @@ class DirectAdjoint(AbstractAdjoint):
 
     So unless you need forward-mode autodifferentiation then
     [`diffrax.RecursiveCheckpointAdjoint`][] should be preferred.
+    If you need only forward-mode autodifferentiation, then [`diffrax.ForwardMode`][] is
+    more efficient.
     """
 
     def loop(
@@ -855,11 +859,13 @@ class BacksolveAdjoint(AbstractAdjoint):
 
 
 class ForwardMode(AbstractAdjoint):
-    """Differentiate through a differential equation solve during the forward pass.
-    (So it is not really an adjoint - it is a different way of quantifying the
-    sensitivity of the output to the input.)
+    """Supports forward-mode automatic differentiation through a differential equation
+    solve. This works by propagating the derivatives during the forward-pass - that is,
+    during the ODE solve, instead of solving the adjoint equations afterwards.
+    (So this is really a different way of quantifying the sensitivity of the output to
+    the input, even if its interface is that of an adjoint for convenience.)
 
-    ForwardMode is useful when we have many more outputs than inputs to a function - for
+    This is useful when we have many more outputs than inputs to a function - for
     instance during parameter inference for ODE models with least-squares solvers such
     as `optimistix.Levenberg-Marquardt`, that operate on the residuals.
     """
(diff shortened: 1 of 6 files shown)


diffrax/_adjoint.py
```diff
@@ -852,3 +852,38 @@ def loop(
         )
         final_state = _only_transpose_ys(final_state)
         return final_state, aux_stats
+
+
+class ForwardMode(AbstractAdjoint):
+    """Differentiate through a differential equation solve during the forward pass.
```
patrick-kidger, line 858: More precisely, this _enables support for forward-mode autodifferentiation_. (Not that it does the autodiff for you itself.)

diffrax/_adjoint.py
```diff
@@ -852,3 +852,38 @@ def loop(
...
+    (So it is not really an adjoint - it is a different way of quantifying the
```
patrick-kidger, line 859: I might specify what an adjoint is here, to make this clear.

docs/requirements.txt
```diff
@@ -3,6 +3,7 @@ mkdocs==1.3.0            # Main documentation generator.
 mkdocs-material==7.3.6   # Theme
 pymdown-extensions==9.4  # Markdown extensions e.g. to handle LaTeX.
 mkdocstrings==0.17.0     # Autogenerate documentation from docstrings.
+mkdocs-autorefs==1.0.1   # Automatically generate references to other pages.
```
patrick-kidger, line 6: This shouldn't be necessary any more, as it's already included below.

patrick-kidger (commented):
Nita aside this LGTM! Once you fix things up I'll merge this into a new dev branch.

diffrax/_adjoint.py
johannahaffner: Right! :D

diffrax/_adjoint.py
johannahaffner: Gave it a shot!

docs/requirements.txt
johannahaffner: Done, sorry for the oversight

johannahaffner:
Completed the above fixes :) 

I also added references to `ForwardMode` to the docstrings of the other adjoints, where they point to `DirectAdjoint` for forward-mode automatic differentiation. I'm recommending it for the case where only forward-mode automatic differentiation is required, but I'm not sure if I should hedge/tone down my statement that `ForwardMode` is more efficient than `DirectAdjoint`. That is - I'm not sure if this statement holds in general. What do you think? 

While going though the errata raised by the other adjoints I was also wondering if there is a case `ForwardMode` cannot handle, such as the `UnsafeBrownianPath` for which some of the other adjoints error out.

johannahaffner:
I benchmarked these a little bit by computing the Jacobian of the output `ys` for a variable number of states, once with one parameter for each state and once with a single parameter for all states. _(So the Jacobians had shape `(num_timepoints, num_states, num_states)` in the first case and `(num_timepoints, num_states, 1)` in the second case.)_

I did not include any couplings between states, but this roughly makes what I have seen on my own data (with coupled ODEs), where the `DirectAdjoint` takes about 1.5x as long as `ForwardMode`.

### Equal number of states and parameters, uncoupled ODE
Largest Jacobian (1024, 512, 512). DirectAdjoint takes ~2.3x as long at high and low end.

Screenshot 2024-12-21 at 12 44 30

### One parameter for all states, uncoupled ODE
Largest Jacobian (4096, 512, 1). DirectAdjoint takes ~1.15x as long at high end, ~2.4x as long at low end

Screenshot 2024-12-21 at 13 05 50

patrick-kidger:
LGTM! And merged :)

As for `DirectAdjoint` -- yup, it's totally expected that this will be slower than `ForwardMode`. Honestly `DirectAdjoint` is kind of a terrible choice, it only exists because of the very occasional need to support both forward and reverse mode through the same solve. (Hopefully one day JAX implements `jvp`-of-`custom_vjp`.)

PR #544, opened 2024-12-16
Import Jaxpr and ClosedJaxpr from jax.exend

jakevdp:
`jax.core.Jaxpr` and `jax.core.ClosedJaxpr` will issue deprecation warnings starting in the next JAX release (0.4.38). They have been available from `jax.extend.core` since JAX v0.4.27, so given that diffrax requires `jax>=0.4.28`, the new import paths can be used unconditionally.


patrick-kidger:
Thanks Jake! I'll go through and do something similar for the other libraries too.

PR #547, opened 2024-12-27
Update _term.py

ricor07:
Allowing args into grad_f for ULD

### DIFF
--- a/test/test_term.py
+++ b/test/test_term.py
@@ -5,6 +5,7 @@ import jax.numpy as jnp
 import jax.random as jr
 import jax.tree_util as jtu
 import pytest
+from diffrax._term import UnderdampedLangevinDriftTerm
 from jaxtyping import Array, PyTree, Shaped
 
 from .helpers import tree_allclose
--- a/test/test_term.py
+++ b/test/test_term.py
@@ -158,3 +158,39 @@ def test_weaklydiagonal_deprecate():
         _ = diffrax.WeaklyDiagonalControlTerm(
             lambda t, y, args: 0.0, lambda t0, t1: jnp.array(t1 - t0)
         )
+
+def test_underdamped_langevin_drift_term_args():
+    """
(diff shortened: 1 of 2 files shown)


lockwo:
Probably worth adding a test to make sure the args work without mucking anything up

ricor07:
Hello, I added a test. Let me know what you think.

test/test_term.py
```diff
@@ -158,3 +158,39 @@ def test_weaklydiagonal_deprecate():
         _ = diffrax.WeaklyDiagonalControlTerm(
             lambda t, y, args: 0.0, lambda t0, t1: jnp.array(t1 - t0)
         )
+
+def test_underdamped_langevin_drift_term_args():
+    """
+    Test that the UnderdampedLangevinDriftTerm handles `args` in grad_f correctly.
+    """    
+    from diffrax._term import UnderdampedLangevinDriftTerm
```
patrick-kidger, line 166: Can you remove this private import?

patrick-kidger:
This pretty much LGTM!

ricor07:
Is it now ok?

patrick-kidger:
@ricor07 heads-up that tests are failing.

Given that these are pre-commit checks take a look at [CONTRIBUTING.md](https://github.com/patrick-kidger/diffrax/blob/main/CONTRIBUTING.md), which describes how to set up pre-commit so that these are automatically applied locally.

Looking at your code I think your new test will fail, by the way: `UnderdampedLangevinDriftTerm` is not defined.

ricor07:
Please, have a look now

patrick-kidger:
So the rest of the file is consistent with using
```python
import diffrax

diffrax.Foo
```
so I think you should do the same. That is to say, `diffrax.UnderdampedLangevinDriftTerm`

ricor07:
You're welcome! I'd like to work on something else, what do you suggest me? Is there a diffrax Slack or Discord?

Issue #551, opened 2025-01-01
CNF Issues


lockwo:
Running the CNF issue on main has two issues:
- A warning with the latest version of imageio 2.36.1 (maybe also good to add a .extras/examples section to the toml to ensure compatible versions for those additional packages?)
```
DeprecationWarning: Starting with ImageIO v3 the behavior of this function will switch to that of iio.v3.imread. To keep the current behavior (and make this warning disappear) use `import imageio.v2 as imageio` or call `imageio.v2.imread` directly.
  img = jnp.asarray(imageio.imread(path))
```
- Error in execution
```
Cell In[12], line 1
----> 1 main(in_path="../imgs/cat.png")

Cell In[11], line 86, in main(in_path, out_path, batch_size, virtual_batches, lr, weight_decay, steps, exact_logp, num_blocks, width_size, depth, print_every, seed)
     84 while step < steps:
     85     start = time.time()
---> 86     value, model, opt_state, step, loss_key = make_step(
     87         model, opt_state, step, loss_key
     88     )
     89     end = time.time()
     90     if (step % print_every) == 0 or step == steps - 1:

    [... skipping hidden 19 frame]

Cell In[11], line 79, in main.<locals>.make_step(model, opt_state, step, loss_key)
     77 value = value / virtual_batches
     78 grads = jax.tree_util.tree_map(lambda a: a / virtual_batches, grads)
---> 79 updates, opt_state = optim.update(grads, opt_state, model)
     80 model = eqx.apply_updates(model, updates)
     81 return value, model, opt_state, step, loss_key

File ~/miniforge3/envs/dev_diffrax/lib/python3.10/site-packages/optax/_src/combine.py:59, in chain.<locals>.update_fn(updates, state, params, **extra_args)
     57 new_state = []
...

ValueError: Expected None, got 0.05.

In previous releases of JAX, flatten-up-to used to consider None to be a tree-prefix of non-None values. To obtain the previous behavior, you can usually write:
  jax.tree.map(lambda x, y: None if x is None else f(x, y), a, b, is_leaf=lambda x: x is None)
```

patrick-kidger:
Hmm! So the first warning I don't have strong feelings about. We could ignore it until imageio v3 comes out (as the `imageio.v2` namespace might not exist for older versions) or switch the namespace as suggested (to remove the warning).

As for the latter, I think it might be sufficient to do:
```diff
- updates, opt_state = optim.update(grads, opt_state, model)
[...]

PR #554, opened 2025-01-02
Fix CNF

lockwo:
https://github.com/patrick-kidger/diffrax/issues/551


patrick-kidger:
LGTM! I should note that I didn't actually check that my suggested fix in #551, actually, well, fixes the error... :p 
Have you done so? If so, very happy with this. (I don't think these examples are covered by the tests, since some of them take a long time to run.)

lockwo:
I didn't run the full thing, but I ran one iteration on my laptop without error so I assume it's fine. 

I also have examples that are very slow to run, my solution was to make the iterations/size a parameter then make that tiny on the GitHub actions version just to test if the examples runs through.

patrick-kidger:
Alright, SGTM! And you're right that handling examples like that would be the better thing for us to do.

Issue #548, opened 2024-12-31
UnderdampedLangevinDiffusionTerm inheritance


lockwo:
Should `UnderdampedLangevinDiffusionTerm` inherit from `_AbstractControlTerm` (in theory, not talking about whether this reduces LoC). The reason I ask is because working on the stateful control, I'm parsing terms to check if they are control terms (which is just done via inheritance check), and given that the diffusion term is a control term, seems like it should inherit (even if there's just some new "AbstractControlTerm" which doesn't have any code but unified inheritance). If this is not the case, should there just be a hard coded check for UnderdampedLangevinDiffusionTerm, or what is the best practice to parse some terms and determine which are control terms?

patrick-kidger:
So `_AbstractControlTerm` is a private implementation detail of `ControlTerm` and `WeaklyDiagonalControlTerm`, just to factor out shared code. Although FWIW as we can now pass `lineax.DiagonalLinearOperator`s to `ControlTerm` then I'm thinking I may simplify things by:
- folding `_AbstractControlTerm` into `ControlTerm`
- maintaining backward compatibility by just turn `WeaklyDiagonalControlTerm` into a stub with a `__new__` that returns a `ControlTerm`.

On to the nature of your question: there's really no such thing as a 'control term' -- these are not any different from regular terms. I think what you want is to have a pytree-of-states corresponding to the pytree-of-terms. And it will simply happen that e.g. `ODETerm` has `None` as its state.

lockwo:
Yea I have it such that it's a pytree of states where ODE is just None state. It was mostly during an initialization step, the ODEterm has no control member variable to do an initialization operation with (which could be checked via attributes). That is to say, when creating the pytree of states, it basically wants to call the AbstractPath's init method (which an ODETerm doesn't have, so I would return None for that state) and so my first instinct was just to parse which ones I call init for (which are the control terms). Writing a new init method for the term seemed unnecessary at first since it would be just calling the control.init anyway.

lockwo:
To be concrete, sometimes the code to the effect of (just my intermediate hack)

```python
    def _path_init(term):
        if isinstance(term, _AbstractControlTerm) or isinstance(
            term, UnderdampedLangevinDiffusionTerm
        ):
            if isinstance(term.control, AbstractPath):
                return term.control.init(t0, end, y0, args, max_steps)
            return None
        elif isinstance(term, MultiTerm):
            return jax.tree.map(
                lambda x: _path_init(x, end),
[...]

Issue #517, opened 2024-10-21
Performance issue with SDE solver


pierreguilmin:
Hello,

When solving the (trivial) SDE $d y_t = -y_t\ dt + 0.2\ dW_t$, the Diffrax Euler solver is ~200x slower than a naive for loop. Am I doing something wrong? The speed difference is consistent across various SDEs, solvers, time steps `dt`, and number of trajectories, and it appears to be specific to SDE solvers.
```python
import diffrax as dx
import jax
import jax.numpy as jnp
from matplotlib import pyplot as plt

# === simulation parameters
key = jax.random.PRNGKey(42)
t0 = 0
t1 = 1
y0 = 1.0
ndt = 101
dt = (t1 - t0) / (ndt - 1)
drift = lambda t, y, args: -y
diffusion = lambda t, y, args: 0.2

# === diffrax euler
brownian_motion = dx.VirtualBrownianTree(t0, t1, tol=1e-3, shape=(), key=key)
solver = dx.Euler()
terms = dx.MultiTerm(dx.ODETerm(drift), dx.ControlTerm(diffusion, brownian_motion))
saveat = dx.SaveAt(ts=jnp.linspace(t0, t1, ndt))

@jax.jit
def diffrax_simu():
    return dx.diffeqsolve(terms, solver, t0, t1, dt0=dt, y0=y0, saveat=saveat).ys

# === homemade euler
@jax.jit
def homemade_simu():
    dWs = jnp.sqrt(dt) * jax.random.normal(key, (ndt,))

    def step(y, dW):
        dy = drift(None, y, None) * dt + diffusion(None, y, None) * dW
        return y + dy, y

    return jax.lax.scan(step, 1.0, dWs)[-1]

# === plot a single trajectory
y = diffrax_simu()
plt.plot(y)
y = homemade_simu()
plt.plot(y)

# === benchmark
%timeit diffrax_simu().block_until_ready()
%timeit homemade_simu().block_until_ready()
```

```text
5.39 ms ± 261 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
19.7 μs ± 899 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

lockwo:
I get them to be a lot closer by using `UnsafeBrownianPath`, which has less overhead than VBT. Diffrax is still a bit slower with this change on my machine, but the difference is smaller (and probably due to other overheads that diffrax does to enable more features).

[...]

Issue #490, opened 2024-08-16
Why can't UBP be backproped?


lockwo:
In the docs it says "You do not need to backpropagate through the differential equation." for UBP usage. However, this doesn't seem to be theoretically necessary, you can just backprop through the solver with the added noise right? What's the motivation requiring this to be the case?

patrick-kidger:
The difficulty is that backprop via `RecursiveCheckpointAdjoint` involves recomputing intermedaite states from checkpoints. I'm not confident that this will be bitwise identical to the original forward pass -- e.g. perhaps due to nondeterminism in convolutions.

Meanwhile, UBP casts float->int->key and uses that to sample its random noise. This means that floating point fluctuations will produce entirely different Brownian motion samples.

lockwo:
I guess it seems to me that these are solvable problems in engineering (I hope), but there isn't any theoretical limitations (like if we want to just store every sample made on the forward pass, and then on the backward pass reuse these samples/use a bridge because we have a lot of memory since neural networks aren't in play, that seems possible maybe not with a UBP, but with something simpler than a VBT). Maybe @frankschae can help articulate this better, and why we are interested in this direction (the motivation is basically, VBT is potentially overkill for us and might introduce slowdowns we don't need, if there is some distinction between UBP, VBT, and some object that does sampling differentiably, has some misc. stuff for weak solvers we are doing, and has bridges then that third object might be the most interesting).

patrick-kidger:
Yup, this all makes sense.

If our controls consumed a step counter or something then one possible approach would just be to `jr.fold_in` the step index into a key, and that should operate reliably. Or if controls were stateful then we could `jr.split` a key each step. The current approach does what it does just because right now, a control consumes only the times at which it is evaluated.

I think this dovetails with the VBT discussion in #489 -- maybe we should think about modifying the way controls are handled, and if we pick our abstractions right we can tackle this issue whilst we're at it.

lockwo:
I think in general, having some concept of state makes sense for general controls. Even if all existing controls can be done without it, we are working on more flexible controls that might take advantage of that (and it would allow more implementations of custom controls in an easier way). That being said, exactly how to do it is open. I think once we have a weak solver PR open to motivate this, it can provide more concrete examples.

lockwo:
[...]

Issue #489, opened 2024-08-16
VBT vs brownian path slowdown


lockwo:
While implementing some weak solver schemes, I noticed that when I used the VBT as opposed to the UBP (unsafe brownian path) it was substantially (like ~5 mins vs ~10 seconds) slower. Using an UBP is fine in this case for us (since its a fixed step solver, and we aren't differentiating through the equation), but in the future is not ideal. Below is a MVC, but in summary:
- This is not the real solver, I ripped out most of everything just to make the code smaller
- These numbers seem a bit small/microbenchmark-y but I have verified them on some larger problems as well, I am just using this small problem for speed and demonstration. 
- VBT is 10x slower here, and seems dominated by the line `u += g1 @ (_dW + chi1)`. Specifically, if I comment that out, I see a decrease in speed from ~11s to 3s. But in UBP it only goes from like 1.1s to 0.8s (so it isn't just the elimination of a matmul making that whole speed gap). 
- The surprising thing is not that VBT is slower (I figured it would come with some overhead), but that it seems to scale as well. Specifically, if I decrease dt, it's not just some constant overhead but seems to increase. Maybe this is expected, but even for small problems with large dts this becomes prohibitive (see the original 5 min vs 10 seconds). 

All of this is a bit surprising since I just call the diffusion control once. Is there a way of using VBT's or integrating them into new solvers that avoids this slowdown, or am I just making some mistake in my usage of the VBT?

Here is the full code:

```python
import jax
from jax import numpy as jnp
import diffrax
from typing import ClassVar

_NORMAL_ONESIX_QUANTILE = -0.9674215661017014


def calc_threepoint_random(x):
    return jnp.where(
        jnp.abs(x) > -_NORMAL_ONESIX_QUANTILE,
        jnp.where(x < _NORMAL_ONESIX_QUANTILE, -1.0, 1.0),
        0.0,
    )


def calc_twopoint_random(x):
    return jnp.where(x > 0, 1.0, -1.0)


class Solver(diffrax.AbstractSolver):

    term_structure: ClassVar = diffrax.AbstractTerm
    interpolation_cls: ClassVar = diffrax.LocalLinearInterpolation

    def func(self, terms, t0, y0, args):
        return terms.vf(t0, y0, args)

    def init(self, terms, t0, t1, y0, args):
        return None

    def step(self, terms, t0, t1, y0, args, solver_state, made_jump):
        drift = terms.terms[0]
        diffusion = terms.terms[1]
[...]

PR #559, opened 2025-01-05
Stateful Controls

lockwo:
This is the v1 of the stateful control PR.

This PR introduces a suite of internally breaking changes which allow for stateful paths. To the non-advanced user, this will likely have no impact on the way they interact with diffrax. It should not impact ODE users at all. SDEs users may be impacted as they might want to try the new "UnsafeBrownianPath" (or as I am calling it "DirectBrownianPath" since it isn't really that unsafe any more), but this should just allow for faster code. I tried several approaches to this problem, but I think based on the conversations in other issues (e.g. https://github.com/patrick-kidger/diffrax/issues/548), that this is at least someone an aligned vision.

Functionaly, this was just a lot of adding arguments to things to support an extra path_state variable. Hence the internal breaking changes. However, these breaking changes come with some decent motivation:
- Flexibility and speed: this allows for SDE users to see meaningful speedups (on smaller problems where the VF isn't something massive). VBT is necessarily (and unecessarily https://github.com/patrick-kidger/diffrax/issues/489) slower than UBP, and in theory is only necessary for adaptive solvers that reject steps (I think? Also maybe stateful paths allow some computations to carry brownian bridge states to work with rejecting steps, I think Frank had some jax code that did that but wasn't in diffrax). Thus for anyone running non adaptive solvers (or running solvers that use pre-visible step size adaptation, not currently in diffrax but we have several of those), VBT is not necessary. In addition to the speedups users get from being able to use UBP where before they had to use VBT, UBP is also faster. UBP is faster in both of its new modes, one in which it just splits the key as it goes along and (much faster) in the case where we precompute brownian motion. This addresses https://github.com/patrick-kidger/diffrax/issues/517 and https://github.com/patrick-kidger/diffrax/issues/490.
- Generalization: paths can be stateful, and this is a more complete and generalizable picture that will likely unlock new functionality (and previously, the UBP relied on a bit of a hack to split the keys to not be stateful). Being applied at the AbstractPath level also checks off one point of https://github.com/patrick-kidger/diffrax/issues/217.
- Also fixes https://github.com/patrick-kidger/diffrax/issues/570

Currently, this PR is missing a few things though.
- Direct adjoint edge case bug
- There's some comments/questions I left for myself (and reviewers) that I still need to address (I can extract some of the main ones to be sure to get feedback)


patrick-kidger:
I can see this is a draft, LMK when you want a review! Happy to take a look on specific points if you want direct feedback earlier.

(Although FWIW I think it'd be cleanest if #484 is the next 'big' thing we merge, just to avoid us having too many 'big' things in flight at the same time.)

lockwo:
> (Although FWIW I think it'd be cleanest if https://github.com/patrick-kidger/diffrax/pull/484 is the next 'big' thing we merge, just to avoid us having too many 'big' things in flight at the same time.)

I'm good with that, although I think that's in @andyElking's control ;)

lockwo:
One question you could answer @patrick-kidger . 

Background: With this PR, the BacksolveAdjoint doesn't work with the new UBP (which is not new, it didn't before either). However, with these changes, it seems to me you it could work with static step size solvers (since you can just take the reverse of the brownian path). For adaptive solvers that reject steps, that already doesn't work for UBP, and for pre visible adaptive solvers maybe that works on the back solve (but if it takes different steps, I'm not sure it would work). My question ii, is it worth adding support for UBP with back solve adjoint (it would require some more book keeping to have the path know more about whether its being run forward/reverse, which it's basically agnostic to currently, so that it could provide the same path, but I think it would be implementable), or does this not matter as a new feature (since I have generally perceived your opinion of backsolveadjoint as detailed below)?

patrick-kidger:
I think 'no' to adding support if it's going to be fiddly / a codepath that could potentially have more bugs. (You have correctly determined my feelings on the matter in the picture!)

But 'yes' if you think it's easy enough to do with only minor adjustments, and you feel that you'd like to add this support.

I think backprop with UBP and RecursiveCheckpointAdjoint is really a lot more interesting though :)

lockwo:
> Backprop with UBP and RecursiveCheckpointAdjoint is definitely the main goal here though!

Indeed, it currently works with RecursiveCheckpointAdjoint and ForwardMode (and DirectAdjoint, but not if DirectAdjoint exceeds a certain number of recursive calls, still working on that one)

lockwo:
Ok, I'm marking this as no longer draft now. There's a few things that I still need to fix (some typing remarks I left, and the directadjoint long run failure edge case but I will need to work up the motivation to dig into bounded loop implementations this weekend lol), but I think for the most part my idea of what this PR should look like is there. 

There's a few open questions, most of them are marked, but I'm putting them here too in case I deleted the comments:
- How should `derivative` be handled with evaluate vs call?
- Should stateful functions (as opposed to natural stateful paths) be supported in control terms?

PR #567, opened 2025-01-13
Add exponential Euler.

GJBoth:
This is a (very...) rough PR to see if there's interest in adding support for exponential integrators. Here's a good [source](https://na.math.kit.edu/download/papers/acta-final.pdf) with some background. Their main use is for stiff, oscillatory problems where neither IMEX nor stiff methods work well, such as brain dynamics or biochemical reaction networks. 

I added the so-called exponential Euler method, which uses forward Euler to evaluate the integral, and which according to this [paper ](https://dl.acm.org/doi/10.1007/s10543-014-0485-1) also works for the stochastic version (additive, commutative noise). 

In short, this solver
- handles both the ODE and SDE version.
- Has a term structure that is compatible with other solvers.
- Has a specialised version for when the linear operator is diagonal. 

The solver is quite straightforward, but the approach requires splitting the equation into terms, a linear and non-linear one. Since the `ODETerm` does not seem to support lineax terms in its vector field (not sure why?), I added a simple `LinearODETerm` which does allow this. Finally, I added the `make_operators` function which specialises to when the operator is diagonal. 

If you like it I can clean this up and merge it.

### DIFF
--- a/diffrax/_term.py
+++ b/diffrax/_term.py
@@ -788,6 +788,16 @@ class AdjointTerm(AbstractTerm[_VF, _Control]):
         return dy, da_y, da_diff_args, da_diff_term
 
 
+class LinearODETerm(ODETerm):
+    operator: lx.MatrixLinearOperator | lx.DiagonalLinearOperator
+
+    def __init__(self, operator: lx.MatrixLinearOperator | lx.DiagonalLinearOperator):
+        self.operator = operator
+        vf = lambda t, y, args: self.operator.mv(y)
+        super().__init__(vector_field=vf)
+
+
+
 # The Underdamped Langevin SDE trajectory consists of two components: the position
 # `x` and the velocity `v`. Both of these have the same shape.
 # So, by UnderdampedLangevinX we denote the shape of the x component, and by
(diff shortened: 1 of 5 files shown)


diffrax/_term.py
```diff
@@ -788,6 +788,16 @@ def _to_vjp(_y, _diff_args, _diff_term):
         return dy, da_y, da_diff_args, da_diff_term
 
 
+class LinearODETerm(ODETerm):
+    operator: lx.MatrixLinearOperator | lx.DiagonalLinearOperator
+
+    def __init__(self, operator: lx.MatrixLinearOperator | lx.DiagonalLinearOperator):
+        self.operator = operator
+        vf = lambda t, y, args: self.operator.mv(y)
+        super().__init__(vector_field=vf)
```
patrick-kidger, line 797: Btw, I try to avoid subclassing concrete classes -- c.f. https://docs.kidger.site/equinox/pattern/

Also, does this strictly need to be an ODE term? I think in other problems, we might be interested in expressing linear terms in this way (for any `operator: lx.AbstractLinearOperator`) regardless of the control.

patrick-kidger (commented):
Okay, left some high-level comments, but basically I think I like this! I'd be interested in taking this as a PR.

PR #575, opened 2025-01-26
JAX 0 5 0 fixes

patrick-kidger:
Should unblock #569. JAX changed it randomness a little bit in this release, and our tests were a bit too sensitive.

### DIFF
--- a/test/test_brownian.py
+++ b/test/test_brownian.py
@@ -123,7 +123,7 @@ def test_shape_and_dtype(ctr, levy_area, use_levy, getkey):
 def test_statistics(ctr, levy_area, use_levy):
     # Deterministic key for this test; not using getkey()
     key = jr.PRNGKey(5678)
-    num_samples = 60000
+    num_samples = 600000
     keys = jr.split(key, num_samples)
     t0, t1 = 0.0, 5.0
     dt = t1 - t0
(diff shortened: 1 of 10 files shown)


test/test_brownian.py
```diff
@@ -122,7 +122,7 @@ def is_tuple_of_ints(obj):
 @pytest.mark.parametrize("use_levy", (True, False))
 def test_statistics(ctr, levy_area, use_levy):
     # Deterministic key for this test; not using getkey()
-    key = jr.PRNGKey(5678)
+    key = jr.PRNGKey(5679)
```
johannahaffner, line 125: Is this some dark magic that I don't get or are you just bumping the seed to keep track of something? :D

patrick-kidger: Ah, this is a total hack on my part. JAX changed how its randomness is generated. It turns out previously we got a little bit lucky with tests passing here. Now it turns out we get unlucky. Instead of making the test more robust... I just bumped the seed.

I'm feeling appropriately called out, I've just rebased to put this back and tweaked the test to be better instead :P

johannahaffner: haha I see! Thanks for the fix :)

lockwo: Is there a reason to use PRNGKey over key?

PR #580, opened 2025-01-28
Update pyproject.toml to meet poetry conventions

joharkit:
in python-poetry ~=3.9 is interpreted as >=3.9<3.10 [2], though it should be >=3.9,<4.0 [2] https://python-poetry.org/docs/dependency-specification/


joharkit:
Since python 4.0 is not yet published, could be also left away, I guess

patrick-kidger:
LGTM! Thank you for the contribution :)

PR #588, opened 2025-02-09
Removes typeguard dependency from pyproject.toml

qodfathr:
Although `dev` was recently updated to no longer require `typeguard`, the dependency is still declared in `pyproject.toml`. This change removes that dependency.


patrick-kidger:
Good spot! Thanks for the fix - merged.

dhirschfeld:
Just ran into problems with the `typeguard` pin. It would be great to have this in a release :pray:

patrick-kidger:
Just opened #600, we'll do a release :)

PR #599, opened 2025-03-09
Fix backwards-in-time solve for AbstractSRKs

sammccallum:
Fixes sign of `h` in `AbstractSRK`.

### DIFF
--- a/diffrax/_solver/srk.py
+++ b/diffrax/_solver/srk.py
@@ -352,10 +352,7 @@ class AbstractSRK(AbstractSolver[_SolverState]):
         else:
             ignore_stage_g = jnp.array(self.tableau.ignore_stage_g)
 
-        # time increment
-        h = t1 - t0
-
-        # First the drift related stuff
+        # # First the drift related stuff
         a = self._embed_a_lower(self.tableau.a, dtype)
         c = jnp.asarray(
             np.insert(self.tableau.c, 0, 0.0), dtype=complex_to_real_dtype(dtype)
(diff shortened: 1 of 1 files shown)


sammccallum:
Fix #599

patrick-kidger:
LGTM!
Can you (a) target `dev` and (b) add a test? We can bundle this in to the upcoming #600 and make this part of the next release :)

sammccallum:
Right, turns out this is a slightly more involved bug! The Lévy area term $H_{s, t}$ isn't correctly handled when solving backwards-in-time too. I believe this is due to VBT returning $-H_{s, t}$ (along with $-W_{s, t}$ and $-K_{s, t}$) whereas $H_{s, t}$ should remain unchanged in sign.

I think there is a choice for whether this should be directly handled by VBT or by the solvers themselves? Tagging 'the SDE lads' @andyElking and @lockwo for their thoughts?

Btw, the original fix (`dt` -> `-dt`) fixes all the solvers to converge backwards-in-time but with strong order 1.0 (because we need to handle $H_{s, t}$ as well)!

sammccallum:
Yeah, that's what I thought and it semi-works: flipping the sign of `H` after it is returned from `diffusion.contr` gets all the additive noise methods (`SEA`, `SRA1`, `ShARK`) back to order 1.5. But the commutative noise methods (`GeneralShARK`, `SlowRK`, `SPaRK`) are still stuck at order 1.0 (for an additive noise SDE).

patrick-kidger:
Okay! I'll let you folks figure out how you want to tackle this, and since it looks like that might take a little time then I'll do a release on #600 in the mean time anyway.

PR #604, opened 2025-03-18
Owen/control revamp

lockwo:


alessandrofasse:
I would be very interested in this. Is there any plan to continue on it :)?

lockwo:
> I would be very interested in this. Is there any plan to continue on it :)?

I will do my best to keep it up to date (as I do with all my PRs, even those that are a year old https://github.com/patrick-kidger/diffrax/pull/463). We have some internal tools that also depend on this branch (so I think it should be a little stable).
 
I think the only outstanding real bug/issue is the weird edge case that cause a assertion error with direct adjoint on problems with enough steps (which I haven't dug into since I haven't been working a ton with diffrax this year so far). 

However, this introduces quite some changes, so the timeline is long.

I have a fork of diffrax that has additional features/solvers and I was thinking this could be part of it as well, although with how much other stuff is going on, I'm not sure that timeline is any shorter lol.

alessandrofasse:
Dear Lockwo, 

thanks for the comment and the explanation. I can imagine that this has a long tail. I think I am not enough into the topic to support you here. But in any case, you work is much appreciated! If there are any updates I would be happy to use it. As of now I think I need to drop `diffrax` for stochastic integration because it introduces for me a factor of being 20x slower, which is in my case a lot.

patrick-kidger:
Yup, this branch is still something I have an eye on. I'm tentatively thinking that after reversible adjoints are in (#603) then this is a good candidate for the next 'big' thing I'm thinking about.

Right now I'm not sure if that will mean stateful controls or something else, since the main downstream goal is a performance improvement for SDEs, and perhaps this is a thing we can obtain in some other way. Either way, a problem for future Patrick 😄

PR #608, opened 2025-03-21
Fixed incompatibility between step_ts/jump_ts and implicit solvers.

patrick-kidger:
Tagging @andyElking for context -- looks like we missed making this specifically an adaptive controller.


andyElking:
Oh I see. Thanks for fixing!

Issue #606, opened 2025-03-20
Slow jit of diffeqsolve since v0.6.1


HeSchatz:
While updating my python (3.11) environment from diffrax v0.5.1 and jax v0.4.29 to the latest versions I've ran into some problems regarding times for jit compilation of the diffeqsolve function and the solve itself. By backtracking the individual releases, I have noticed two major increases in compile times, one when updating from diffrax v0.6.0 to v0.6.1 and one when updating jax from v0.4.38 to v0.5.3 (I did not check all intermediate releases yet). The duration for the ODE solve itself increases significantly when updating from v0.6.1 to v0.6.2.

Below are the jit and solve times for integration of a neural ode model with time-dependent inputs. The solver used is Tsit5 with a PI step size controller. All runs report the exact same amount of solver steps taken. I'm using double precision due to issues with very small integration step sizes.

| jax version | diffrax version | jit + solve | solve
| --- | --- | --- | --- |
| v0.4.29 | v0.6.0 | ~1.5 s | ~0.4 s |
| v0.4.29 | v0.6.1 | ~30 s | ~0.4 s |
| v0.4.38 | v0.6.2 | ~25 s | ~1.5 s |
| v0.4.38 | v0.7.0 | ~25 s | ~1.5 s |
| v0.5.3 | v0.7.0 | ~ 500 s | ~ 1 s |

For jax v0.5.3 following message is printed after some time:
````
********************************
[Compiling module jit_diffeqsolve] Very slow compile? If you want to file a bug, run with envvar XLA_FLAGS=--xla_dump_to=/tmp/foo and attach the results.
********************************
````

Scanning through the changelog I could not find any changes that may cause this behaviour. Do you have any idea what might cause these problems?

The code used for testing is below. I can also provide the data and test model if needed. If you are wondering why I'm vmapping the ODEs rhs instead of the solve, I have found that this is significantly faster for my use case (atleast it used to be, maybe it's time to re-evaluate...). However, as only one (very long) input trajectory is used in the test case below, this shouldn't be relevant to this issue.

````
import json
from time import time

import diffrax as dfx
import equinox as eqx
import jax
import jax.nn as jnn
import jax.numpy as jnp

jax.config.update("jax_enable_x64", True)

class ODEFunc(eqx.Module):
    mlp: eqx.nn.MLP

    def __init__(self, state_size, input_size, width_size, depth, *, key, **kwargs):
        super().__init__(**kwargs)
[...]

PR #614, opened 2025-04-04
fix for t0==t1 when initializing ys, as opposed to inside the loop

dkweiss31:
Eliminates code introduced in https://github.com/patrick-kidger/diffrax/pull/494 whereby `ys` was updated inside of the loop to `y0` in the case of `t0==t1`. Now `ys` is instead initialized appropriately with `y0`. This is in response to performance issues observed in https://github.com/patrick-kidger/diffrax/issues/606.

### DIFF
--- a/diffrax/_integrate.py
+++ b/diffrax/_integrate.py
@@ -788,58 +788,8 @@ def loop(
                 save_state = _save(tfinal, yfinal, args, subsaveat.fn, save_state)
         return save_state
 
-    def _save_if_t0_equals_t1(subsaveat: SubSaveAt, save_state: SaveState) -> SaveState:
-        if subsaveat.ts is not None:
-            out_size = 1 if subsaveat.t0 else 0
-            out_size += 1 if subsaveat.t1 and not subsaveat.steps else 0
-            out_size += len(subsaveat.ts)
-
-            def _make_ys(out, old_outs):
-                outs = jnp.stack([out] * out_size)
-                if subsaveat.steps:
-                    outs = jnp.concatenate(
-                        [
-                            outs,
-                            jnp.full(
-                                (max_steps,) + out.shape, jnp.inf, dtype=out.dtype
-                            ),
-                        ]
-                    )
-                assert outs.shape == old_outs.shape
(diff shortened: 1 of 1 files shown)


dkweiss31:
I will investigate these test failures. Patrick can you confirm the adjoint failures are unrelated to this PR? I definitely need to address the event failures though.

patrick-kidger:
I think these changes are being made on top of an old version of Diffrax `main`. In particular I see the following warning:
```
/home/runner/work/diffrax/diffrax/diffrax/_integrate.py:1258: DeprecationWarning: shape requires ndarray or scalar arguments, got <class 'jax._src.api.ShapeDtypeStruct'> at position 0. In a future JAX release this will be an error.
1588
    (max_steps,) + jnp.shape(x), jnp.inf, dtype=x.dtype
```
Which does not reflect the current state or line number of that line:

https://github.com/patrick-kidger/diffrax/blob/2fafbc7591506834341822a869d9eefbdfefce82/diffrax/_integrate.py#L1268

That might explain the failures? (Which I can't reproduce locally.)

dkweiss31:
Hmm it seems I was wrong: those test failures do directly result from my changes (I do think I based my changes on the most recent version of main, I see commit 2fafbc7591506834341822a869d9eefbdfefce82 as the most recent change before mine. You might be referring to the below line?)
https://github.com/patrick-kidger/diffrax/blob/36c1d3e4b05456332afa1923340cd924f76da134/diffrax/_integrate.py#L1257-L1259 

In particular, commenting out the below section (which actually makes the replacement with the correct value of `y0` if `t0==t1`), the `test_adjoint.py` tests pass locally, whereas they do fail locally if I leave this section uncommented. I'm a little baffled by these errors, which don't seem to point to a particular line: maybe its upset that we are asking about `t0==t1` in `where`?
https://github.com/patrick-kidger/diffrax/blob/36c1d3e4b05456332afa1923340cd924f76da134/diffrax/_integrate.py#L1216-L1222

patrick-kidger:
Closing in favour of #618.

PR #618, opened 2025-04-11
Resolves speed issue of #606.

### LINKED ISSUE #606: Slow jit of diffeqsolve since v0.6.1

patrick-kidger:
Fixes #606.
Supersedes #614.

The main part of this change is to use a dynamic-update-slice on the `t0==t1` branch, rather than returning entirely new arrays. This is in-line with the way all of our other updates happen.

Whilst I'm here I've also slightly refactored some other saving code.

### DIFF
--- a/diffrax/_integrate.py
+++ b/diffrax/_integrate.py
@@ -240,14 +240,23 @@ def _save(
     args: PyTree,
     fn: Callable,
     save_state: SaveState,
+    repeat: int,
 ) -> SaveState:
     ts = save_state.ts
     ys = save_state.ys
     save_index = save_state.save_index
 
-    ts = ts.at[save_index].set(t)
-    ys = jtu.tree_map(lambda ys_, y_: ys_.at[save_index].set(y_), ys, fn(t, y, args))
-    save_index = save_index + 1
+    ts = lax.dynamic_update_slice_in_dim(
+        ts, jnp.broadcast_to(t, (repeat,)), save_index, axis=0
+    )
+    ys = jtu.tree_map(
+        lambda ys_, y_: lax.dynamic_update_slice_in_dim(
+            ys_, jnp.broadcast_to(y_, (repeat, *y_.shape)), save_index, axis=0
+        ),
+        ys,
+        fn(t, y, args),
+    )
+    save_index = save_index + repeat
 
     return eqx.tree_at(
         lambda s: [s.ts, s.ys, s.save_index], save_state, [ts, ys, save_index]
@@ -306,7 +315,9 @@ def loop(
 
     def save_t0(subsaveat: SubSaveAt, save_state: SaveState) -> SaveState:
         if subsaveat.t0:
-            save_state = _save(t0, init_state.y, args, subsaveat.fn, save_state)
+            save_state = _save(
+                t0, init_state.y, args, subsaveat.fn, save_state, repeat=1
+            )
         return save_state
 
     save_state = jtu.tree_map(
@@ -638,6 +649,7 @@ def loop(
     final_state = outer_while_loop(
         cond_fun, body_fun, init_state, max_steps=max_steps, buffers=_outer_buffers
     )
+    save_state = final_state.save_state
     result = final_state.result
 
     if event is None or event.root_finder is None:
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Tagging @dkweiss31 -- I've taken a crack at an alternative to #614. It handles the adjoint issue you bumped into it by happening inside the inner `adjoint.loop` call (rather than in the outer `diffeqsolve`). I realised our original strategy was probably simplest here, and it just needed expressing in a different way. LMK how this looks for you!

@HeSchatz -- I've tested that this gives good performance on the microbenchmark of https://github.com/patrick-kidger/diffrax/issues/606#issuecomment-2742752044 Can you give this a go on your actual problem?

HeSchatz:
I've tested with my actual problem and compile times are back down to ~1.5s with this commit. Great! In fact, it seems to be slightly faster now than with 0.6.0, most likely due to some other code improvements between 0.6.0 and now. Thank you both @patrick-kidger and @dkweiss31 for fixing this so fast!

PR #620, opened 2025-04-21
Updated to new doc stack. Phew, that took a long time!

patrick-kidger:
In particular this has been a great moment to do a doc refresh, and pick up many nitty issues where we've had missing docstrings, this kind of thing.

Tagging @johannahaffner here, as I've gone all-in on using jaxtyping annotations without further explanation. This was a point we debated a while ago, and I think you've convinced me :) (Some of them do get pretty unwieldy though, see `diffrax.adjoint_rms_seminorm`.)


docs/requirements.txt
johannahaffner, line 1: This needs to be updated in CONTRIBUTING.md :) It still has the old instructions for building documentation, which assume that the dependencies are listed in this file.

johannahaffner (commented):
Nice work! I'd love to build the docs to check this out :)

docs/requirements.txt
patrick-kidger: Good catch, ty! Just updated.

johannahaffner:
> (Some of them do get pretty unwieldy though, see `diffrax.adjoint_rms_seminorm`.)

I'm afraid I don't see that part! Local build vs. current online version: 

Screenshot 2025-04-22 at 20 53 49
Screenshot 2025-04-22 at 20 53 58

Nice color scheme and highlighting though! Even nicer to read. Also - why did we never link to the source code before this update?

Issue #625, opened 2025-05-02
Performance Issues with SubSaveAt


alessandrofasse:
# Description

I’m utilizing diffrax to simulate a system characterized by a substantial state vector, specifically `y = (20_000, 6)`, over extended durations, such as 120,000 timesteps. Storing the entire trajectory is impractical and unnecessary. Instead, I aim to record a reduced dataset derived from a matrix multiplication of the state. Precisely, I have a matrix `A` with dimensions `(85, 20_000)`, and I compute `A @ y[0]` to obtain weighted sums of the state.

To optimize performance, I only need to save these reduced outputs at intervals defined by dt_monitor, which is a multiple of the solver’s timestep dt. Drawing inspiration from issue #592, I implemented the following approach:

```
iterations = int(t / dt) + 1
ts = jnp.linspace(0, t, iterations)
if jnp.isclose(dt, self.delta_t):
    sub_monitors[name] = dfx.SubSaveAt(steps=True, fn=fn)  # Potentially includes info in the last step
else:
    n = int(dt_monitor / dt)
    sub_monitors[name] = dfx.SubSaveAt(ts=ts[1:][::n], fn=fn)
monitors = dfx.SaveAt(subs=sub_monitors)
sol = dfx.diffeqsolve(
    self.ode_term,
    dfx.Heun(),
    t0=0.0,
    t1=t,
    y0=y0,
    args=self.args,
    saveat=monitors,
    dt0=None,
    stepsize_controller=dfx.StepTo(ts=ts),
    max_steps=iterations,
)
```

# Issues Encountered
1.	**Performance Degradation with Reduced Sampling:** When setting dt_monitor = dt, the simulation runs approximately 1.5 times faster compared to cases where dt_monitor = n * dt (with n > 1). Intuitively, reducing the frequency of saved outputs should enhance performance due to decreased I/O operations. However, I observe the opposite. Additionally, GPU utilization drops to around 80% in the reduced sampling scenario, and there’s a noticeable delay before computation commences. In contrast, using steps=True maintains 100% GPU utilization throughout.
2.	**inf Values in Final State with steps=True:** Utilizing `steps=True` occasionally results in inf values in the final state at t1. This issue seems related to the handling of the final timestep during saving. A similar problem was discussed in issue #488, where intermediate saved values were sometimes inf when t0 == t1. Is this related?

# Suggestions
[...]

PR #626, opened 2025-05-02
Added steps=n logic for skipping steps

alessandrofasse:
[See https://github.com/patrick-kidger/diffrax/issues/](https://github.com/patrick-kidger/diffrax/issues/625)

### DIFF
--- a/test/test_saveat_solution.py
+++ b/test/test_saveat_solution.py
@@ -264,6 +264,12 @@ def test_saveat_solution_skip_vs_saveat():
 
     sol_skip = solve(saveat_skip)
     sol = solve(saveat)
+    assert sol is not None
+    assert sol.ts is not None
+    assert sol.ys is not None
+    assert sol_skip is not None
+    assert sol_skip.ts is not None
+    assert sol_skip.ys is not None
     assert jnp.allclose(sol_skip.ts, sol.ts)
     assert jnp.allclose(sol_skip.ys, sol.ys)
 


diffrax/_saveat.py
```diff
@@ -32,9 +32,29 @@ class SubSaveAt(eqx.Module):
     t0: bool = False
     t1: bool = False
     ts: Optional[Real[Array, " times"]] = eqx.field(default=None, converter=_convert_ts)
-    steps: bool = False
+    steps: int = 0
     fn: Callable = save_y
```
patrick-kidger, line 36: We can remove the default values here now that we have an explicit `__init__` method instead.

examples/neural_sde.ipynb
patrick-kidger, line 1: I think these changes to the examples are spurious?

test/test_citation.py
```diff
@@ -24,7 +24,7 @@ def test_adaptive_sde(capfd, getkey):
     dfx.citation(
         terms=terms,
         stepsize_controller=dfx.ClipStepSizeController(
-            stepsize_controller, store_rejected_steps=1
+            stepsize_controller, store_rejected_steps=True
```
patrick-kidger, line 27: I suspect this got caught a in a match-replace, it should still be a `1`.

test/test_saveat_solution.py
```diff
@@ -131,6 +131,27 @@ def test_saveat_solution():
     assert sol.stats["num_steps"] > 0
     assert sol.result == diffrax.RESULTS.successful
 
+    saveat = diffrax.SaveAt(steps=2)
+    sol = _integrate(saveat)
```
patrick-kidger, line 135: We should also add tests for `SaveAt(steps=n, t1=True)` and `SaveAt(steps=n, t1=False)` when the number of steps does not divide `n`. Then check that in the first case with e.g. `n=2` and 5 steps that we store `[0 2 4 5]` and in the second case we store `[0 2 4]`

diffrax/_integrate.py
```diff
@@ -1235,9 +1246,9 @@ def _allocate_output(subsaveat: SubSaveAt) -> SaveState:
             # maximum amount of steps we can possibly take.
             if max_steps is None:
                 raise ValueError(
-                    "`max_steps=None` is incompatible with saving at `steps=True`"
+                    "`max_steps=None` is incompatible with saving at `steps=n`"
                 )
-            out_size += max_steps
+            out_size += max_steps // subsaveat.steps
         if subsaveat.t1 and not subsaveat.steps:
```
patrick-kidger, line 1252: Note that we have quite a few usecases of `if subsaveat.steps` or `not subsaveat.steps` that should become `if subsaveat.steps != 0` and `if subsaveat.steps == 0`.

patrick-kidger: Although in this particular line I think the logic needs to be something like `if subsaveat.t1 and (max_steps % subsaveat.steps == 0)`, since it may be the case that we need to store an extra `t1` that does not occur in one of our saved steps.

diffrax/_integrate.py
```diff
@@ -487,14 +487,25 @@ def maybe_inplace(i, u, x):
             return eqxi.buffer_at_set(x, i, u, pred=keep_step)
 
         def save_steps(subsaveat: SubSaveAt, save_state: SaveState) -> SaveState:
-            if subsaveat.steps:
-                ts = maybe_inplace(save_state.save_index, tprev, save_state.ts)
-                ys = jtu.tree_map(
-                    ft.partial(maybe_inplace, save_state.save_index),
-                    subsaveat.fn(tprev, y, args),
-                    save_state.ys,
+            steps = int(subsaveat.steps)
```
patrick-kidger, line 490: I think the `int` should be unnecessary, this should already be an integer.

diffrax/_integrate.py
```diff
@@ -487,14 +487,25 @@ def maybe_inplace(i, u, x):
...
+            if steps:
```
patrick-kidger, line 491: I like to recommend against implicit-cast-to-bool, it's a really easy source of bugs.

<details><summary>Click to expand</summary>

Copy-pasting from a (nonpublic) style guide I've previously written:

Use
```python
if len(some_list) > 0:
    ...
else:
    ...
if x is None:
    ...
```
and not
```python
if some_list:
    ...
else:
    ...
if not x:
    ...
```
The former (a) clearly conveys intent and (b) avoids subtle bugs:
```python
def foo(seed: Optional[int]):
    if not seed:
        seed = generate_random_seed()
    ...
foo(seed=0)  # Secretly random!
```
I mean, what is this really:
```python
>>> all([])
True
>>> all([[]])
False
>>> all([[[]]])
True
```

</details>

diffrax/_integrate.py
```diff
@@ -487,14 +487,25 @@ def maybe_inplace(i, u, x):
...
+                save_step = (state.num_accepted_steps) % subsaveat.steps == 0
```
patrick-kidger, line 492: Nit, spurious brackets.

diffrax/_integrate.py
```diff
@@ -487,14 +487,25 @@ def maybe_inplace(i, u, x):
...
+                    new_ys = jtu.tree_map(
+                        lambda yi, ysi: maybe_inplace(save_state.save_index, yi, ysi),
+                        fn_result,
+                        save_state.ys,
+                    )
+                    return new_ts, new_ys
+
+                ts, ys = lax.cond(
+                    should_save, save_branch, lambda: (save_state.ts, save_state.ys)
                 )
```
patrick-kidger, line 507: Aha, you've bumped into an XLA compiler footgun! I'm guessing it looked like the logic I described in https://github.com/patrick-kidger/diffrax/issues/625#issuecomment-2846752579 could be simplified into just a single cond.

Unfortunately this now means that we have our large buffers `new_ts` and `new_ys` passing through control flow, and when combined with autodiff this means in some cases that XLA will spuriously copy the buffers on each branch, rather than doing a conditional write. (This is pretty obscure inside baseball, I totally appreciate.)

It's for this reason that when writing to these buffers, it's important to encode the conditional-ness of the write in the `pred=...` argument used inside `maybe_inplace`.

patrick-kidger (commented):
Thank you for the contribution! I like the look of this. I've left a first round of review comments. :)

patrick-kidger (commented):
Thanks for pushing this so far!

For `_save_t1`, then you're right, it gets a bit hairy! I think the idea for reasoning this through is as follows. We had some logic around whether to save the final value, which is a complicated combination of `.t1`, `.steps`, and the root-finding machinery. The important part of it, however, is when we check `subsaveat.steps`: at this point we were checking whether the value at `t1` is already going to be saved via `steps=True`. This is the logic we need to preserve: whether the value at `t1` is going to be saved anyway via `steps=True`.

In this PR, that corresponds to the boolean value `num_accepted_steps % subsaveat.steps == 0`. If this is `True` then we will have already saved the `t1` value via `.steps`. If it is `False` then we will not.

---

Armed with this intuition, let's see what that looks like in code. Now sadly the following will of course crash with a tracer error, but it highlights what I think the change in logic should be:
```diff
 def _save_t1(subsaveat, save_state):
     if event is None or event.root_finder is None:
-        if subsaveat.t1 and not subsaveat.steps:
+        if subsaveat.t1 and not (final_state.num_accepted_steps % subsaveat.steps == 0):
             # If subsaveat.steps then the final value is already saved.
             save_state = _save(
                 tfinal, yfinal, args, subsaveat.fn, save_state, repeat=1
             )
     else:
-        if subsaveat.t1 or subsaveat.steps:
+        if subsaveat.t1 or (final_state.num_accepted_steps % subsaveat.steps == 0):
             # In this branch we need to replace the last value with tfinal
             # and yfinal returned by the root finder also if subsaveat.steps
             # because we deleted the last value after the event time above.
             save_state = _save(
                 tfinal, yfinal, args, subsaveat.fn, save_state, repeat=1
             )
     return save_state
```
This hits a tracer error, as the value of the modulus might not be known at compile time, and we have to wait until runtime.

So at this point we then need to chase through some annoying JAX details, because our saving has now moved from 'definitely known at compile time' to 'maybe known at compile time but maybe only known at runtime'.

Untested, but I think that means rewriting the `_save_t1` function as follows:
```python
def _save_t1(subsaveat, save_state):
    if subsaveat.steps == 0:
        # We're not saving the final value via `steps`, so we might need to save it via `t1`.
        t1_saved_via_steps = False
    elif subsaveat.steps == 1:
        # We're definitely saving the final value via `steps`, so we can skip saving it via `t1`.
        t1_saved_via_steps = True
    else:
        # We might be saving the final value via `steps`, so we might need to save it via `t1`.
        t1_saved_via_steps = final_state.num_accepted_steps % subsaveat.steps == 0
    if event is None or event.root_finder is None:
        if type(t1_saved_via_steps) is bool:
            t1_not_saved_via_steps = not t1_saved_via_steps
        else:
            t1_not_saved_via_steps = jnp.logical_not(t1_saved_via_steps)
        pred = subsaveat.t1 & t1_not_saved_via_steps
    else:
        # If we're using an event with a root finder, and are saving steps, then we need to write
        # the final value here because we deleted the last value after the event time above.
        pred = subsaveat.t1 | t1_saved_via_steps
    if pred is not False:
        save_state = _save(
            tfinal, yfinal, args, subsaveat.fn, save_state, repeat=1, pred=pred
        )
    return save_state
```
along with the following small diff to `_save`:
```diff
 def _save(
     t: FloatScalarLike,
     y: PyTree[Array],
     args: PyTree,
     fn: Callable,
     save_state: SaveState,
     repeat: int,
     pred: bool,
 ) -> SaveState:
     ts = save_state.ts
     ys = save_state.ys
     save_index = save_state.save_index
 
     ts = lax.dynamic_update_slice_in_dim(
         ts, jnp.broadcast_to(t, (repeat,)), save_index, axis=0
     )
+    y_to_save = lax.cond(pred, lambda: fn(t, y, args), lambda: jtu.tree_map(lambda ys_: ys_[save_index], ys))
     ys = jtu.tree_map(
         lambda ys_, y_: lax.dynamic_update_slice_in_dim(
             ys_, jnp.broadcast_to(y_, (repeat, *y_.shape)), save_index, axis=0
         ),
         ys,
-        fn(t, y, args),
+        y_to_save,
     )
     save_index = save_index + jnp.where(pred, repeat, 0)
 
     return eqx.tree_at(
         lambda s: [s.ts, s.ys, s.save_index], save_state, [ts, ys, save_index]
     )
```
But please do check what I've written here looks correct, I haven't tested this at all! 😁

alessandrofasse:
Dear Patrick,

thanks for comments and explanations. This helped me a lot and I continued on the PR. Most of it worked and is good input, however, I again hit now two barriers:

1. `test_saveat_solution_skip_steps` here I get different sizes than the ones I expect (in the test). I am 80% sure that my results are correct but can you please confirm that my test should return what I assert?
2. I changed back the above `save_fn` to just return `subsaveat.fn(tprev, y, args)`, the problem is that I got errors with the `lax.cond` returning different shapes. I was also wondering, how do we know the shape of the output without evaluating it? Should we do somewhere a dummy call to extract the shape?

patrick-kidger:
Awesome! I've just taken a look at this and I think it's nearly there. On `test_saveat_solution_skip_steps`, I identified a few bugfixes: during the loop we were still using `maybe_inplace` which hardcodes the condition as `keep_step` rather than `keep_step & save_step`, and at t1 we were unconditionally saving `t` but not not `y`. So I've just pushed a commit that tackles these.

As I'm doing so, I've also squashed your commits and rebased the branch on latest `main`.

Anyway, `test_saveat_solution_skip_steps` now passes! Tell me what you think about these changes? I've not yet ran the full test suite, we'll see what the CI says.

alessandrofasse:
Thanks for checking and finding this issue! Very nice!

On the ci `test_saveat_solution_skip_vs_saveat` still fails. Did you push the lastest version that passed on your machine?

patrick-kidger:
I simply did not update that test as well! :) In particular if you take a look at the other `test_saveat_solution_skip_steps` test that I did update, then you can see that the logic I've selected results in picking different time points (due to an off-by-one change) than we had before. (This was what I found cleanest to implement.)

Can you take a try at updating the test and seeing if things work for your use-case as you expect them to?

alessandrofasse:
Thats a good idea! I changed now the final test from my side that now passes as expected. I did not run the full test suite, but I hope this will work now. Thanks again for your time and effort

patrick-kidger:
Looks like pre-commit is failing! Probably solveable with some tactical `assert sol.ts is not None` I think.

patrick-kidger:
Alright, merged! :D Thank you for your work getting this in. This is on the `dev` branch and will be part of the upcoming Diffrax release. :)

Issue #637, opened 2025-05-22
How to properly scale with multiple event conditions


LuggiStruggi:
Hi diffrax team,

I am working on a problem, where the amount of cond_fn scales with my state dimension. Unfortunately, this currently scales pretty badly timewise even when jitting the code. Do you have any idea what could be done to improve here?

Here as an example:

```math
\dot{y} + y = 0
```
```math
y(0) \sim [0.9, 1.0]
```
event condition (any state):
```math
y - 0.1 = 0
```
 Solving with Euler and a fixed stepsize of 0.005ms between 0 and 30ms (or until first event), it scales really badly with the state dimension:

```
   Dim |   Time (s)
--------------------
     1 |   0.001022
     2 |   0.000625
     4 |   0.000657
     8 |   0.001162
    16 |   0.001870
    32 |   0.004110
    64 |   0.010566
   128 |   0.044464
   200 |   0.218422
```

see for comparison, the timing where i just used the cond_fn on the first element of the state `y[0]`:

```
   Dim |   Time (s)
--------------------
     1 |   0.000723
     2 |   0.000612
     4 |   0.000595
     8 |   0.000615
    16 |   0.000596
    32 |   0.000716
    64 |   0.000903
   128 |   0.000724
   200 |   0.001835
```

Do you see any way to speed this up a reasonable amount?

Thanks for the help!

Here is the MWE i used (Single cond_fn commented out)

```python
import jax
import jax.numpy as jnp
import diffrax as dfx
import optimistix as optx
from timeit import default_timer as timer

def solve_once(y0, t0, t1, dt0, solver, stepsize_controller, event):
    term = dfx.ODETerm(lambda t, y, args: -y)
    return dfx.diffeqsolve(
        terms=term,
        solver=solver,
        t0=t0,
        t1=t1,
        dt0=dt0,
        y0=y0,
        args=None,
        stepsize_controller=stepsize_controller,
        event=event,
        saveat=dfx.SaveAt(t0=False, t1=True, steps=False),
        max_steps=3000000,
        throw=True,
    )

solve_once = jax.jit(solve_once, static_argnames=["solver", "stepsize_controller", "event"])

def benchmark_diffeqsolve_with_event():
    key = jax.random.PRNGKey(0)
    t0, t1, dt0 = 0.0, 30.0, 0.005
    solver = dfx.Euler()
    controller = dfx.ConstantStepSize()
[...]

PR #643, opened 2025-05-26
Introduction of bidirectional vs. unidirectional triggering of events

LuggiStruggi:
Changed Event to support flags for unidirectional vs. bidirectional triggering.

### DIFF
--- a/diffrax/_event.py
+++ b/diffrax/_event.py
@@ -19,8 +19,38 @@ class Event(eqx.Module):
     [`diffrax.diffeqsolve`][].
     """
 
-    cond_fn: PyTree[Callable[..., BoolScalarLike | RealScalarLike]]
-    root_finder: optx.AbstractRootFinder | None = None
+    cond_fn: PyTree[Callable[..., Union[BoolScalarLike, RealScalarLike]]]
+    trig_dir: PyTree[None | bool]
+    root_finder: Optional[optx.AbstractRootFinder]
+
+    def __init__(
+        self,
+        cond_fn,
+        root_finder: Optional[optx.AbstractRootFinder] = None,
+        trig_dir: Union[None | bool, PyTree[None | bool]] = None,
+    ):
+        vals_cond, treedef_cond = flatten(cond_fn)
+
+        if isinstance(trig_dir, bool):
+            vals_trig = [trig_dir] * len(vals_cond)
+            treedef_trig = treedef_cond
+        else:
+            vals_trig, treedef_trig = flatten(trig_dir, is_leaf=lambda x: x is None)
+            print(vals_trig, treedef_trig)
+
+        if treedef_cond != treedef_trig:
+            raise ValueError("Missmatch in the structure of cond_fn and trigger_dir")
+
+        if not all(x is None or isinstance(x, bool) for x in vals_trig):
+            raise ValueError(
+                "`trig_dir` must be a None, bool or a PyTree of None | bools"
+                + " with the same structure as cond_fn"
+            )
+
+        trig_tree = unflatten(treedef_cond, vals_trig)
+        self.cond_fn = cond_fn
+        self.root_finder = root_finder
+        self.trig_dir = trig_tree
 
 
 Event.__init__.__doc__ = """**Arguments:**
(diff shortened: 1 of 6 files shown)


diffrax/_event.py
```diff
@@ -21,7 +22,32 @@ class Event(eqx.Module):
...
+        root_finder: Optional[optx.AbstractRootFinder] = None,
+        bidirect: Union[bool, PyTree[bool]] = True,
+    ):
+        vals, treedef = flatten(cond_fn)
+        n = len(vals)
+
+        if isinstance(bidirect, bool):
+            bidir_list = [bidirect] * n
+        else:
+            bidir_list = list(bidirect)
```
patrick-kidger, line 40: I think the logic here does not reflect the pytree structure of `cond_fn`. It seems to operate on the flattened representation.

Can we add a test for this case?

diffrax/_event.py
```diff
@@ -21,7 +22,32 @@ class Event(eqx.Module):
...
+    def __init__(
+        self,
+        cond_fn,
+        root_finder: Optional[optx.AbstractRootFinder] = None,
+        bidirect: Union[bool, PyTree[bool]] = True,
+    ):
+        vals, treedef = flatten(cond_fn)
+        n = len(vals)
+
+        if isinstance(bidirect, bool):
```
patrick-kidger, line 37: I'm inclined not to support this case and require exactly matching pytrees. I expect it might be a bit unusual to have multiple events that all have the same crossing requirements, and this seems like an easy footgun. WDYT?

diffrax/_event.py
```diff
@@ -21,7 +22,32 @@ class Event(eqx.Module):
...
     cond_fn: PyTree[Callable[..., Union[BoolScalarLike, RealScalarLike]]]
-    root_finder: Optional[optx.AbstractRootFinder] = None
+    bidirect: PyTree[bool]
+    root_finder: Optional[optx.AbstractRootFinder]
+
+    def __init__(
+        self,
+        cond_fn,
+        root_finder: Optional[optx.AbstractRootFinder] = None,
+        bidirect: Union[bool, PyTree[bool]] = True,
```
patrick-kidger, line 32: Perhaps this should be a `PyTree[None | bool]`: handling the cases of upcrossing-only, downcrossing-only, or either crossing. WDYT? (No strong feelings from me, since one crossing can easily be constructed from the other.)

patrick-kidger (commented):
Very minor comments only, this LGTM!

diffrax/_event.py
LuggiStruggi: I feel inclined to keep it since at least in my case (if not using a single cond_fn for a whole vector as discussed in https://github.com/patrick-kidger/diffrax/issues/637#issuecomment-2923113557), all cond_fns have the same crossing direction. I feel like it could be helpful but you definitely have a bigger overview for what people might use this library and therefore what might be confusing vs helpful.

diffrax/_event.py
```diff
@@ -21,7 +22,37 @@ class Event(eqx.Module):
...
+
+    def __init__(
+        self,
+        cond_fn,
+        root_finder: Optional[optx.AbstractRootFinder] = None,
+        trig_dir: Union[None | bool, PyTree[None | bool]] = None,
+    ):
+        vals_cond, treedef_cond = flatten(cond_fn)
+
+        if isinstance(trig_dir, bool):
```
patrick-kidger, line 36: Should this also include the case of `trig_dir = None`?

diffrax/_event.py
```diff
@@ -21,7 +22,37 @@ class Event(eqx.Module):
...
+            vals_trig = [trig_dir] * len(vals_cond)
+            treedef_trig = treedef_cond
+        else:
+            vals_trig, treedef_trig = flatten(trig_dir, is_leaf=lambda x: x is None)
+            print(vals_trig, treedef_trig)
```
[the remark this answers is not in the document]
LuggiStruggi: oh sorry ^^! will fix

LuggiStruggi:
I think i can close this, since it has been merged through https://github.com/patrick-kidger/diffrax/pull/655.

PR #644, opened 2025-05-27
start root finder at time of previous ode_solver_step

LuggiStruggi:
Initialize root finder closer to earlier roots at t_prev instead of t_next, when multiple solutions exist.


LuggiStruggi:
Needs to be looked at again

PR #647, opened 2025-05-29
fix paths to example notebooks

johannahaffner:
The links were broken and lead to a 404 not found error - I think the recent update of the documentation moved the examples? Anyway, it's now fixed for all of them.


patrick-kidger:
Aha, thank you!
I think there are also some other spurious chnanges here?

johannahaffner:
Ah damn, those are recreated (but not actually changed) encodings for the figures in the top markdown cell, bit of a Jupyter quirk.
I always executed this cell to see if the new link renders correctly and can be opened.

I can see if this can be removed!

PR #655, opened 2025-06-18
Introduction of bidirectional vs. unidirectional triggering of events

patrick-kidger:
Supersedes #643.

@LuggiStruggi - in getting ready to merge this feature, I have:

- Squashed all your commits together.
- Fixed a crash when using pytree-valued condition functions.

WDYT? If all looks good and the CI passes then I think we can merge this.

### DIFF
--- a/diffrax/_event.py
+++ b/diffrax/_event.py
@@ -2,8 +2,8 @@ import abc
 from collections.abc import Callable
 
 import equinox as eqx
+import jax.tree_util as jtu
 import optimistix as optx
-from jax.tree import flatten, unflatten
 from jaxtyping import Array, PyTree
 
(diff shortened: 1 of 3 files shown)


LuggiStruggi:
Looks great ^^Thanks! Only a little error caused by `direction in {None, True False}` which can be fixed by using a tuple instead.
I added the fix to https://github.com/patrick-kidger/diffrax/pull/643.

patrick-kidger:
Ah, yes. Thank you! I've just pushed that fix here.
(I'd happily use #643 but I think you might be merging in rather than squashing/rebasing, so the commit history that github is showing gets a bit unwieldy – I try to keep a clean history in what we merge.)

patrick-kidger:
Aaaand merged!
Thank you for putting this together @LuggiStruggi, this is awesome. :D

Issue #657, opened 2025-06-20
Odd behaviour of StepTo controller for small time steps


aidancrilly:
I have found some odd behaviour of the StepTo controller for small time steps, the result is also floating point precision dependent.

Below is a minimum working example. For single precision, the time jumps straight to t1 after 1 time step and the final integration result is incorrect.

``` python 3
import diffrax
import jax.numpy as jnp
import jax

t0 = 0.0
t1 = 1e-9

dt = 2e-12
dt_max = 1e-11
step_t = t0
step_ts = [step_t]
while(step_t < t1):
    step_t += dt
    step_ts.append(step_t)
    dt = min(2*dt,dt_max)

step_ts[-1] = t1
print(step_ts)

stepsize_controller = diffrax.StepTo(ts=step_ts)

def f(t,y,args):
    jax.debug.print('{t}',t=t)
    return -y/t1

sol = diffrax.diffeqsolve(terms=diffrax.ODETerm(f),
                    solver=diffrax.Euler(),
                    t0=t0,
                    t1=t1,
                    dt0=None,
                    y0=jnp.ones(1),
                    stepsize_controller=stepsize_controller)

print(sol.ts,sol.ys)
```

If I run in double using:
``` python3
from jax import config

config.update("jax_enable_x64", True)
```

Things are marginally better but the jax.debug.print still suggests that the final few times are clipped to t1.

If I swap t1 to 1.0 and change dt and dt_max accordingly, the correct behaviour is restored. I would however like to run with time steps of this order (nanoseconds) for physics simulation. Cheers!

patrick-kidger:
Hi there! So this also came up recently in #632, which also discusses the reasoning why this happens. 

This isn't ideal for use-cases like yours, however, and I'd definitely welcome thoughts(/PRs) on how this might be improved.

aidancrilly:
Ah apologies for missing that very similar issue (and from someone in my same field!).

Reasons make sense. For now, I'll probably go the quick route of just externally rescale the time axis to 0 to 1 and apply a correction to the ode term to compensate for this before passing to diffrax.

I don't know if it would be possible to, internally to diffeqsolve, linearly rescale the time axis to [0,1] for the integration and interpolation. This would seem easy (in principle! Seems daunting to me when I glance into the code) for ODEs but I'm not familiar with SDEs or CDEs. Does this rescaling sort these issues for all use cases?

patrick-kidger:
[...]

PR #660, opened 2025-06-28
Use 100 ULP's to clip timesteps close to t1

### LINKED ISSUE #632: Issue with small time steps
First off, thanks for this (and your other) libraries! Very useful for scientific programming with JAX.

I'm having an odd issue with the time-stepping. If the time step is too small, it looks like there is some kind of numerical error that creeps in. Here's a simple MWE with the ODE being exponential decay 

```python 

from diffrax import diffeqsolve, Dopri5, Euler,ODETerm, SaveAt, ConstantStepSize
import jax 
vector_field = lambda t, y, args: -y / args[0]
term = ODETerm(vector_field)
solver = Euler()
saveat = SaveAt(t0=True, t1=True)
stepsize_controller = ConstantStepSize()

tau = 1e-11
dt0 = tau / 1000
sol = diffeqsolve(term, solver, t0=0, t1=tau, dt0=dt0, y0=1.0, saveat=saveat,
                  args=(tau,),
                  stepsize_controller=stepsize_controller, max_steps=10000)

print(sol.ts[0:10]) 
print(sol.ys[0:10])  

```

It's set up so that when you set the decay time `tau`, it'll run for 1000 steps for one decay time, so the solution will always be 1/e. 

For tau over about 1e-9, you get the right answer. If tau < 1e-9, you start getting some pretty bad numerical error. For instance, tau = 1e-10 gives the answer as 1e-4. 

If I switch to `Bosh3`, the issue is less - but still, there is a `tau` dependence on the error 

Image

The same thing run through `scipy.integrate.solve_ivp` gives no `tau` dependence on the solution error (though of course, `diffrax` is like two orders of magnitude faster!). I jitted the input to `scipy.integrate.solve_ivp` and it made no difference - so I guess the issue is somewhere in `diffrax`? 

Image

Can I reparametrize the ODE I'm working with to avoid small numbers? Yes, but it does make me worry about numerical stability. Any advice?

jpbrodrick89:
This allows timesteps within 100 ULP's of t1 for `ConstantStepSize` which can be smaller than the previously fixed absolute tolerance of `1e-10` for double precision. This fixes #632 and #657 as discussed.

Using `for _ in range(100)` rather than multiplying by `100.0` takes about 7µs rather than 3µs which would be undesirable if we had to perform the computation at every timestep. Fortunately, this is not necessary and we just compute `t1_clip_floor` at the beginning of `_integrate.loop`.

I am ambivalent with whether a multiplier of 100 is optimal here, 1000 would be more than fine for double precision but would limit total number of steps to about 1000 in single precision which seems overly restrictive.

Happy to follow up with a `ConstantStepSize` PR once this is merged.

### DIFF
--- a/diffrax/_integrate.py
+++ b/diffrax/_integrate.py
@@ -258,15 +258,11 @@ def _save(
     )
 
 
-def _clip_to_end(tprev, tnext, t1, keep_step):
-    # The tolerance means that we don't end up with too-small intervals for
-    # dense output, which then gives numerically unstable answers due to floating
+def _clip_to_end(tprev, tnext, t1, t1_clip_floor, keep_step):
+    # The tolerance of ~100 ULP's means that we don't end up with too-small intervals
+    # for dense output, which then gives numerically unstable answers due to floating
     # point errors.
-    if tnext.dtype is jnp.dtype("float64"):
-        tol = 1e-10
-    else:
-        tol = 1e-6
-    clip = tnext > t1 - tol
+    clip = tnext > t1_clip_floor
     tclip = jnp.where(keep_step, t1, tprev + 0.5 * (t1 - tprev))
     return jnp.where(clip, tclip, tnext)
 
@@ -303,6 +299,11 @@ def loop(
     outer_while_loop,
     progress_meter,
 ):
+    # Calculate in advance t1 - 100 ULP's: the threshold at which to round tnext to t1
+    t1_clip_floor = t1
+    for _ in range(100):
+        t1_clip_floor = eqxi.prevbefore(t1_clip_floor)
+
     if saveat.dense:
         dense_ts = init_state.dense_ts
         dense_ts = dense_ts.at[0].set(t0)
@@ -392,7 +393,7 @@ def loop(
         #
 
         tprev = jnp.minimum(tprev, t1)
-        tnext = _clip_to_end(tprev, tnext, t1, keep_step)
+        tnext = _clip_to_end(tprev, tnext, t1, t1_clip_floor, keep_step)
 
         progress_meter_state = progress_meter.step(
             state.progress_meter_state, linear_rescale(t0, tprev, t1)
--- a/diffrax/_integrate.py
+++ b/diffrax/_integrate.py
@@ -303,6 +303,11 @@ def loop(
     t1_clip_floor = t1
     for _ in range(100):
         t1_clip_floor = eqxi.prevbefore(t1_clip_floor)
+    if jnp.sign(t1_clip_floor - t0) != jnp.sign(t1 - t0):
+        raise ValueError(
+            "t0 and t1 are within 100 units of least precision "
+            "implying a high risk of floating point error. Aborting."
+        )
 
     if saveat.dense:
         dense_ts = init_state.dense_ts


jpbrodrick89:
Open to better ideas for error checking of traced values, but hopefully the possibility of having `t0` and `t1` is so vanishing rare we don't have to test for it.

patrick-kidger:
Nice, this LGTM!

> I am ambivalent with whether a multiplier of 100 is optimal here, 1000 would be more than fine for double precision but would limit total number of steps to about 1000 in single precision which seems overly restrictive.

I'm not sure I understand the concern in single precision here?

> but hopefully the possibility of having t0 and t1 is so vanishing rare we don't have to test for it.

I'm also not sure I understand this concern, can you expand?

---

CC @aidancrilly @varchasgopalaswamy could you both give this PR a go? It'd be good to know if this fixes your problems!

jpbrodrick89:
Sorry I got the single precision comment a bit wrong,1 ULP is essentially a rtol of about 6e-8 (I was thinking 6e-6 as I misread my own comments) meaning that going with 1000 ULP would limit us to 16000 timesteps, that probably seems as good as anyone could expect for single precision. Do you think I should bump up to 1000 ULPs?

My second comment was if t1-t0 was within tolerance then the clipped t1 would be on the other side of t0 (i.e. in most cases below) and results will be very confusing. Just as if it would have been nice if we could have raised an error or warning if dt0 was less than 1e-10 with the old implementation, I thought I'd try warn her but that doesn't seem possible. However as our tolerance (at least with double precision) is incredibly low I don't think it's really worth worrying about.

patrick-kidger:
> meaning that going with 1000 ULP would limit us to 16000 timesteps

Hmm, it sounds like you're computing `(1 / 6e-8) / 1000`, i.e. you're considering specifically a solve over the time interval `[0, 1]`? If so then bear in mind that (a) it's only the last 1000 ULPs that matter (so we need to subtract by that, not divide) and that (b) 1 ULP is a not a uniform size over the time interval. Brute-forcing the total number of ULPs between 0 and 1 (and entirely ignoring subnormal numbers):
```python
import equinox.internal as eqxi
import jax
import jax.lax as lax
import jax.numpy as jnp

x = jnp.array(1.)

@jax.jit
def run(start):
    def cond(carry):
        _, value = carry
        return value > 0
    def body(carry):
        counter, value = carry
        return counter + 1, eqxi.prevbefore(value)
    counter, _ = lax.while_loop(cond, body, (0, start))
    return counter

print(run(x))  # 1056964609
```

> I thought I'd try warn

Right, so if we did want this then I think this actually is possible. We could definitely precompute the t-clip value at the start of `diffeqsolve`, and then use an `eqx.error_if` to handle the case that `(t0 > tclip) & (t0 == t1)`.

But as a practical matter I think I like the current approach which basically just enforces that we take a single step. (We could add a test for this case, perhaps? Not super important.)

aidancrilly:
Hi guys,

I can confirm that this fix works in my MWE in #657 as well as in the main code of interest.

Plotting out the results from MWE as a function of t1 from this PR vs 0.7.0 release:

image

image

Cheers Jonathan!

jpbrodrick89:
With regards to the warning, yes, I think the current behaviour should be fine (taking a single step to `tclip`) I think we can leave that.

With regards to the maximum number of steps, I meant maximum number of steps with a constant step size such that there is never any clipping. Clipping is a sometimes surprising effect when using a constant step size (although this might be avoided by the constant step size PR I'll draft after this) and it would be good to think carefully when the surprise is warranted. In such a case, `dt0` must be smaller than `t1 - t1_clip_floor` giving the maximum number of such steps as `(t1 - t0)/(t1 - t1_clip_floor)` which would be approximately 160k for 100 ULP's, and 16k for 1000 ULP's.

I've plotted the maximum number of constant steps for 100 ULP's and 1000 ULP's as a function of `t1-t0` for `t0 = 0.0` and `t0 = 1.0`:

image

image

My gut feel is that expecting more than 10k steps with single precision without any surprises is asking too much and that we can probably bump to 1000 ULP's. WDYT?

patrick-kidger:
> With regards to the maximum number of steps, I meant maximum number of steps with a constant step size such that there is never any clipping.

Ah, right, I get you now.

I think my feeling is that ideally we'd determine the number based on what gives us enough range to avoid instability in dense interpolation calculations. I could see 100 being too small but perhaps 1000 is too large?

Actually, that does make me now think: maybe the 'correct' fix would be to (a) adjust `ConstantStepSize` as discussed; (b) adjust the dense interpolation routines to be robust to short intervals; (c) remove clipping altogether? (This is certainly more work though.)

WDYT?

jpbrodrick89:
Would `StepTo` also need a dedicated fix with that proposal?

Do you feel we are introducing any additional instability compared to the status quo with 1000 ULP's? If not, could we implement this (or 200/500 ULP's if you prefer) as a temporary fix and then work on the long term fix after that? As I said I would be happy to help with part (a).

patrick-kidger:
> Would StepTo also need a dedicated fix with that proposal?

I don't think so. This already steps to precise locations, so for this the clipping is honestly a bit weird.

> Do you feel we are introducing any additional instability compared to the status quo with 1000 ULP's? If not, could we implement this (or 200/500 ULP's if you prefer) as a temporary fix and then work on the long term fix after that?

I don't think so. I agree, let's merge this (with 100 ULPs) as it's certainly at least an improvement on what we had before! I've just triggered CI, to merge once it (hopefully) passes.

jpbrodrick89:
Looks like root finder no longer requires pyright ignores for some reason, I have removed the suggested ignore comments as pyright passes on my side. (Pre-commit did run fine so not sure whether this is because it's checking more files or due to an upstream change.)

jpbrodrick89:
I'm guessing, the failing test is a compounded rounding error from the `+ dt` with `ConstantStepSize`. Increasing `max_steps` by 1 or the number of ULP's to 500 ~~removes~~ DOES NOT remove the failure.

I'm going to see if I can push the constant step size fix first in a separate PR.

jpbrodrick89:
Changed to target dev and merged recent changes. Tests seem to pass except for Jax bug. Hopefully good to go!

patrick-kidger:
Awesome! LGTM, and merged :D

PR #675, opened 2025-08-05
Key bug in loss function for CNF example

homerjed:
Hi Patrick,

Looked at the nice CNF example again and noticed that in the loss function you are splitting `key` defined before training  instead of `train_key` split from the `loss_key` passed into the loss function. This is a bug right?

Cheers
Jed


patrick-kidger:
Ty, good spot! This looks good to me.

Issue #650, opened 2025-06-03
Recommendation for saving auxilary data


jpbrodrick89:
Often times we compute auxiliary variables each timestep within `vector_field` that we would like to save as part of processing for future reference/sanity checking. We would like to avoid having to manually recalculate these within postprocessing (for confident debugging). What is the best practice way of achieving this? One option is to have a custom stepper where the auxilary data is not "stepped"/incremented but instead updated/replaced but doesn't feel ideal. Ideally, something like a `has_aux` like in the standard `jax.vjp` could be supported. Any other suggestions? Thanks 🙂

lockwo:
While it might involve some re-computation, this sounds like it could be done with a `SaveAt(fn=)`

patrick-kidger:
Yup, I think @lockwo's approach is probably best.

In general we don't support saving aux output from a vector field because it's not really clear how this is defined in general: for example when making multiple vector field calls inside the implicit solve or a stiff solver, or when time progresses nonmonotonically due to step rejection.

This means that aux output is best thought of as a function of `(t, y, args)` and handled via `SaveAt(fn=...)`, even if that just involves calling the vector field -- rather than as an output of the vector field directly.

PR #699, opened 2025-10-25
Adding PostProcessSolver Wrapper Class

LennartGevers:
I've been using the same kind of trick as suggested in #200  , for a kind of similar reason.

In my case, I'm integrating angular velocities by representing rotations as unit-quaternions (due to: numerical stability, gimbal-lock, ...). For this I had to add a quaternion _renormalization_ which simply divides the quaternion that is returned by the solver by its norm, so that the trajectory "stays" on $\mathrm{SU}(2)$.

As this pattern of _post-processing_ the solvers output seems to be useful in various situations I've opened this PR to propose a simple WrapperClass similar to `HalfStep` that allows to inject some post-processing functionality.

If this feature seems worth adding to diffrax for you, I could also setup an example usecase for unit-testing or maybe also for a small example in the docs.

There might be also some other tricks, that might work with a Wrapper class, such as:

- Ability to extend the `DenseInfo` Struct, so that auxiliary data that is not exposed to `save_fn` can be saved after each step (#650).
- Ability to extend the `SolverState` Struct, so that data from the previous iteration can be passed to the args parameter of the next iteration.

### DIFF
--- a/diffrax/_solver/base.py
+++ b/diffrax/_solver/base.py
@@ -352,27 +352,7 @@ HalfSolver.__init__.__doc__ = """**Arguments:**
 """
 
 
-class PostProcessSolver(
-    AbstractAdaptiveSolver[_SolverState], AbstractWrappedSolver[_SolverState]
-):
-    """Wraps another solver and postprocesses every step it returns. This lets you
-    tweak the outputs of an existing solver – for example adjusting error estimates,
-    projecting onto constraints, or populating additional dense output fields –
-    without reimplementing the solver itself.
-
-    Each call to `step` runs the wrapped solver once, then forwards the tuple
-    `(y1, y_error, dense_info, solver_state, result)` to `postprocess_fn`. The return
-    value of that function becomes the value returned to the caller, so you can modify
-    or augment any part of the solver's outputs. Aside from whatever work your
-    post-processing does, no extra solver evaluations are required.
-
-    !!! tip
-
-        This is a lightweight way to customise off-the-shelf solvers – use it to add
-        safety checks, change convergence diagnostics, or integrate with external
-        bookkeeping, all while keeping the solver implementation untouched.
-    """
-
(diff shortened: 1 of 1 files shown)


patrick-kidger:
So I think the general case you're considering here is that of solving an ODE on a manifold / solving a semi-explicit DAE. The way of framing this feature that I've been considering is something like:
```python
# Define a manifold as points at which a function equals zero.
surface = lambda x: x.conj().T @ x - jnp.eye(x.shape[0])
# Define a way to project to the surface.
projection = optimistix.Newton()
diffeqsolve(..., constraint=Constraint(surface, projection))
```
This hasn't been something I've yet tackled, principally because I'd like to be sure that this is doing the right thing for the general case of DAEs.

Whilst the above is the API I have in mind, as an implementation I suspect it would look a lot like what you have! (Actually, it's often possible to do slightly better: you can prepend the projection – your postprocess function – to the vector field, so that even the evaluations happening within the solver are strictly on the desired manifold.)

For that reason I'm thinking that this probably isn't something I'd like to merge, to leave room for the above API at a later date.

LennartGevers:
Alright, thanks for the point about DAEs, @patrick-kidger  — I actually hadn’t looked at it from that angle yet.
So far, I’ve mostly treated the renormalization as a numerical side note (that’s the physicist in me, yk…).
But seeing it as a DAE with the quaternion renormalization as basically a projection operator onto the unit sphere makes way more sense. Thanks for that!

Supporting DAEs in Diffrax sounds super interesting, though I imagine it would need some deeper changes. For example, in curved (non-Euclidean) spaces like the unit-quaternion manifold, the way the PIDController currently computes error isn’t quite geometrically accurate. In my case, I wrapped the PIDController and pre-process the error using the quaternion log-map, so it can still assume Euclidean structure internally. (The same problems would also arise for interpolation and I guess also steady_state_event).

Issue #703, opened 2025-11-05
Infinite loop in PID stepsize controller due to float precision error


philipwijesinghe:
There is a special case that leads to the same step in PID being recalculated infinitely.

This arises in PIDController.adapt_step_size() when t0 is high, dt is clipped to dtmin, and force_dtmin=True

dt sets next_t1 as: `next_t1 = next_t0 + dt`
this is then read back next loop and prev_dt is recalculated as: `prev_dt = t1 - t0`

Due to float precision error, sometimes prev_dt != dt, and `(prev_dt <= self.dtmin)` ends up False

So, keep_step stays False, but dt never drops below dtmin.

See below example:

```python
import jax
import jax.numpy as jnp

jax.config.update('jax_enable_x64', True)

DTYPE = jnp.float64

# Case:
# dt is clipped to dtmin, and we wish to continue solver (force_dtmin=True)
# let's say dtmin = 0.2
dt = jnp.asarray(0.2, dtype=DTYPE)

# Let's say we are at an arbitrary time point>>dtmin in the solver
next_t0 = jnp.asarray(967.1378944377053, dtype=DTYPE)
next_t1 = next_t0 + dt

# next_t0, next_t1 are returned
# an ODE step is taken
# adapt_step_size() is called again with t0=next_t0, t1=next_t1

# We calculate the actual dt taken
# prev_dt = t1 - t0
prev_dt = next_t1 - next_t0
print(f"desired dt: {dt}; actual dt: {prev_dt}")

# Because t0 >> dt, there is a chance for float error such that prev_dt > dt (and dtmin)
# in keep_step = keep_step | (prev_dt <= self.dtmin)
print(prev_dt <= dt)  # is FALSE!

# then, keep_step=False
# factor < 1
# dt is clipped to dtmin again
# ...
# Infinite loop!
```

My solution was to add a flag in the `controller_state` (called `keep_next_step`)

```python
if self.dtmin is not None:
    keep_step = keep_step | keep_next_step
```
```python
if self.dtmin is not None:
    if not self.force_dtmin:
        result = RESULTS.where(dt < self.dtmin, RESULTS.dt_min_reached, result)
    # flag next step to be kept if dtmin is reached
    # or if it was reached previously and dt is unchanged
    keep_next_step = (dt <= self.dtmin) | (keep_next_step & (factor == 1))
    dt = jnp.maximum(dt, self.dtmin)
```

Which works for my stuff, but unsure if it breaks things down the line..

[...]

PR #704, opened 2025-11-06
Fix for PIDController infinite loop at dtmin

philipwijesinghe:
A proposed fix for an unintended infinite loop encountered in PIDController, issue: #703 

## issue

This situation is encountered in `pid.PIDController.adapt_step_size()` when we set a `dtmin` and `force_dtmin=True`.

When the previous step is performed at `dt=dtmin`, the intention was for `keep_step` to be set to `True`
```python
if self.dtmin is not None:
    keep_step = keep_step | (prev_dt <= self.dtmin)
```
However, `prev_dt` is recalculated from `prev_dt = t1 - t0`, when in the previous step `next_t1 = next_t0 + dt`.
The intention is that `prev_dt`  equals `dt` from the last step, but when the exponents of `t0`, `t1` floats are higher than of `dt`, we loose float precision in `prev_dt`, and there is a good chance of  `(prev_dt <= self.dtmin)` being False.

In this situation, no further values are changed and steps loop and fail infinitely.

## solution

Instead of `(prev_dt <= self.dtmin)` we add a bool flag `at_dtmin` which we pass via the `controller_state`.

We set this to True when `dt <= self.dtmin`. Importantly, we also want to keep it as True if we are at `dtmin` already, and we want to avoid an accumulation of precision errors if we stay at `dtmin` for a long time. So, in this case, we keep resetting `dt` to the desired `dtmin`.
```python
if self.dtmin is not None:
    if not self.force_dtmin:
        result = RESULTS.where(dt < self.dtmin, RESULTS.dt_min_reached, result)
    # if we are already at dtmin and dt is unchanged (factor == 1),
    # reset dt to dtmin to avoid accumulating float precision errors
    dt = jnp.where(at_dtmin & (factor == 1), self.dtmin, dt)
    # this flags the next loop to accept step
    at_dtmin = dt <= self.dtmin
    dt = jnp.maximum(dt, self.dtmin)
```

### DIFF
--- a/diffrax/_step_size_controller/pid.py
+++ b/diffrax/_step_size_controller/pid.py
@@ -81,7 +81,7 @@ def _select_initial_step(
     return jnp.minimum(100 * h0, h1)
 
 
-# _PidState = (prev_inv_scaled_error, prev_prev_inv_scaled_error, keep_next_step)
+# _PidState = (prev_inv_scaled_error, prev_prev_inv_scaled_error, at_dtmin)
 _PidState = tuple[RealScalarLike, RealScalarLike, BoolScalarLike]
 
 
@@ -470,7 +470,7 @@ class PIDController(
         (
             prev_inv_scaled_error,
             prev_prev_inv_scaled_error,
-            keep_next_step,
+            at_dtmin,
         ) = controller_state
         error_order = self._get_error_order(error_order)
         prev_dt = t1 - t0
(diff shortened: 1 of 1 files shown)


johannahaffner:
Thanks! I will take a look as soon as I'm able :)

diffrax/_step_size_controller/pid.py
```diff
@@ -545,6 +547,11 @@ def _scale(_y0, _y1_candidate, _y_error):
         if self.dtmin is not None:
             if not self.force_dtmin:
                 result = RESULTS.where(dt < self.dtmin, RESULTS.dt_min_reached, result)
+            # if we are already at dtmin and dt is unchanged (factor == 1),
+            # reset dt to dtmin to avoid accumulating float precision errors
+            dt = jnp.where(at_dtmin & (factor == 1), self.dtmin, dt)
```
patrick-kidger: I think `dt = prev_dt * 1` should imply `dt = prev_dt` exactly regardless of precision errors / floating point weirdness. Did you find this line necessary?

patrick-kidger (commented):
One comment aside, this LGTM!

patrick-kidger:
Okay, merged! Thank you for fixing this :D

Issue #705, opened 2025-11-07
Obscure error for passing EulerHeun instead of EulerHeun()


rdaems:
Hi,

I've wasted some time because I made a small mistake, and the error was quite obscure.
I passed `EulerHeun` instead of `EulerHeun()` as the solver argument.
Minimal example:

```
import jax.random as jr
from diffrax import diffeqsolve, ControlTerm, EulerHeun, MultiTerm, ODETerm, SaveAt, VirtualBrownianTree

t0, t1 = 0, 3
drift = lambda t, y, args: -y
diffusion = lambda t, y, args: 0.1 * t
brownian_motion = VirtualBrownianTree(t0, t1, tol=1e-3, shape=(), key=jr.PRNGKey(0))
terms = MultiTerm(ODETerm(drift), ControlTerm(diffusion, brownian_motion))
solver = EulerHeun
saveat = SaveAt(dense=True)

sol = diffeqsolve(terms, solver, t0, t1, dt0=0.05, y0=1.0, saveat=saveat)
print(sol.evaluate(1.1))  # DeviceArray(0.89436394)
```

This returns
```
--> 137 assert type(term_contr_kwargs) is tuple
```

For the Euler solver, the error is different:
```
TypeError: functools.partial() argument after ** must be a mapping, not property
```
Maybe this can be improved? I can take a look later if I have some time.

johannahaffner:
Hi! Thanks for the report. 

A simple `isinstance(solver, dfx.AbstractSolver)` should detect if the solver hasn't been instantiated at compile time. Given that this would be a one-liner inside `diffeqsolve`, we can either easily implement it, or there is a good reason we haven't already. I can take a look to make up my mind which it is! @lockwo do you have insight here?

Type-checking should also catch this, in case you need another tool in your arsenal to find errata like this faster. This would look like

```python
import diffrax as dfx 

from jaxtyping import jaxtyped
from beartype import beartype as typechecker

# Other code

sol = jaxtyped(typechecker=typechecker)(dfx.diffeqsolve)(...)
```

lockwo:
[...]

PR #706, opened 2025-11-09
Add an informative error message if the solver has not been instantiated

johannahaffner:
Solves https://github.com/patrick-kidger/diffrax/issues/705.

### DIFF
--- a/diffrax/_integrate.py
+++ b/diffrax/_integrate.py
@@ -1084,6 +1084,22 @@ def diffeqsolve(
     y0 = jtu.tree_map(_promote, y0)
     del timelikes
 
+    # Check if the solver is an instance of AbstractSolver and provide an informative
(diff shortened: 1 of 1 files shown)


diffrax/_integrate.py
```diff
@@ -1084,6 +1084,22 @@ def _promote(yi):
     y0 = jtu.tree_map(_promote, y0)
     del timelikes
 
+    # Check if the solver is an instance of AbstractSolver and provide an informative
+    # error if it is not. Addresses https://github.com/patrick-kidger/diffrax/issues/705
+    if not isinstance(solver, AbstractSolver):
+        if issubclass(solver, AbstractSolver):
```
patrick-kidger: This will error out if `solver` is not a class. We need `isinstance(solver, type) and issubclass(solver, AbstractSolver)`.

PR #708, opened 2025-11-13
test: add tracer-leak check

nstarman:


nstarman:
@patrick-kidger in testing quaxify(diffrax) I'm seeing errors that I think originate from having the tracer checks on. This means that diffrax isn't compatible with `JAX_CHECK_TRACER_LEAKS=1` and this PR should demonstrate the issue.

patrick-kidger:
Hmm, bother.

I think fixing the underlying issue comes under "I'd be happy to take a PR on that". We could then add a test at the same time. (Ideally just a single test, to avoid duplicating out our test suite by too much.)

On the other CI changes here - heads-up that we have a dev branch with this updated.

PR #728, opened 2026-02-07
Add example for neural gru-ode

LoSpiri:
Hi, I added an example in the drowpdown "Neural Differential Equations" for neural gru-ode, if it can be interesting you can add it to the documentation. Also modified the mkdocs.yml.
Thanks in advance and obviously I'm open to feedback


patrick-kidger:
Thanks for writing this! However I'm afraid I think this particular models is fairly niche, so I'm inclined to leave it out of the documentation :)

PR #734, opened 2026-02-20
Combine bits via xor when bitcasting from larger to smaller type

### LINKED ISSUE #733: UnsafeBrownianPath.evaluate() bitcasts to int32, which results in unexpected results when using float64 timesteps
In this code snippet:

https://github.com/patrick-kidger/diffrax/blob/b3a188513f91d2969746e92f2b352f04763ed80f/diffrax/_brownian/path.py#L133-L134

When t0 and t1 are of dtype `float64` , this bitcasting to `int32`  can result in different values of `t0` and `t1` being casted to identical `int32`'s. Consequently, this produces noise that is identical in subsequent timesteps, which is unexpected behaviour. I suppose this could be fixed either by always casting to `int64` if `jax_enable_x64` is `True`, or by checking the dtype of `t0` and `t1` and casting to `int64` if the dtype is `int64` or `float64`. I'd be happy to make a PR for this, just let me know whether which fix is preferred! [Update: this would not work, see comment]

Note that this doesn't occur for all values of `t0` and `t1`, but presumably only those that can not be accurately represented as `float32`, but I'd have to look into how `lax.bitcast_convert_type` works to be sure. 

Here is a MWE:
```
from jax import numpy as jnp
from jax import random as jr
import diffrax as dfx

t0 = jnp.float64(4096)
t1 = jnp.float64(4096 + 1e-4)
t2 = jnp.float64(4096 + 2e-4)

path = dfx.UnsafeBrownianPath(shape=(1,), key=jr.PRNGKey(0))
noise_1 = path.evaluate(t0, t1)
noise_2 = path.evaluate(t1, t2)
print("Noise 1:", noise_1)
print("Noise 2:", noise_2)
assert not jnp.isclose(noise_1, noise_2), "Noise values for different time intervals should not be the same"
```
Output:
```
Noise 1: [-0.00833833]
Noise 2: [-0.00833833]
AssertionError: Noise values for different time intervals should not be the same
```

alexander-de-ranitz:
Fixes #733. 

Previously, when downcasting to a type with fewer bits, the larger type would be converted to a smaller intermediate dtype, which was subsequently bitcasted to the requested type. The casting to a smaller intermediate type loses precision, and can result in distinct values being cast to identical values. Since these values are used as keys to generate random sequences, this is problematic, as it results in identical noise being generated in subsequent timesteps. This commit fixes this by not throwing away bits when the input type is larger than the requested output type. Instead, the larger number is bitcast to multiple values in the smaller type, which are then combined using xor.

### DIFF
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -9,3 +9,22 @@ def test_fill_forward():
     out_ = jnp.array([jnp.nan, 0.0, 1.0, 1.0, 1.0, 2.0, 2.0])
     fill_in = diffrax._misc.fill_forward(in_[:, None])
     assert tree_allclose(fill_in, out_[:, None], equal_nan=True)
+
+
+def test_force_bitcast_convert_type():
+    val_1 = jnp.float64(1e6)
+    val_2 = jnp.float64(1e6 + 1e-4)
+
+    # Val_1 and val_2 are different as float64,
+    # but would be the same if naively downcast to float32.
+    assert val_1 != val_2
+    assert val_1.astype(jnp.int32) == val_2.astype(jnp.int32)
+
+    val_1_cast = diffrax._misc.force_bitcast_convert_type(val_1, jnp.int32)
+    val_2_cast = diffrax._misc.force_bitcast_convert_type(val_2, jnp.int32)
+
+    assert val_1_cast.dtype == jnp.int32
+    assert val_2_cast.dtype == jnp.int32
+
+    # Bitcasted values should be different in the smaller type
+    assert val_1_cast != val_2_cast


patrick-kidger:
Looks reasonable to me! (Though I think the pre-commits are failing, see [`CONTRIBUTING.md`](https://github.com/patrick-kidger/diffrax/blob/main/CONTRIBUTING.md).) It'd also be great to see tests that the new functionality is correct but no worries if you're up for that?

diffrax/_misc.py
```diff
@@ -12,23 +12,17 @@
...
-    val = val.astype(intermediate_type)
-    return lax.bitcast_convert_type(val, new_type)
+    result = lax.bitcast_convert_type(val, new_type)
+    
+    # If downcasting (larger -> smaller type), bitcast returns multiple values.
+    # Combine them via XOR to ensure nearby input values map to different outputs.
+    if result.shape != val.shape:
+        result = jnp.bitwise_xor.reduce(result, axis=-1)
+    
+    return result
```
patrick-kidger: Maybe an `assert result.shape == val.shape` afterwards to be sure that this xor has done its job?

alexander-de-ranitz:
Added two assertions (shape and dtype) and a simple test which fails with the previous implementation, but passes with these changes. The formatting issues have also been fixed. LMK if any further additions are necessary!

patrick-kidger:
Looks perfect! Merged 🎉 Thank you for the fix.

Issue #598, opened 2025-03-09
Solving backwards-in-time fails for AbstractSRK's


sammccallum:
### Description
When solving from `t1` to `t0` with `-dt0`, `AbstractSRK` produces an incorrect solution.

Here is an example:
```python
def drift(t, y, args):
    a, b = args
    return -a * y


def diffusion(t, y, args):
    a, b = args
    return b


t0 = 0
t1 = 1.0
dt0 = 0.001
y0 = jnp.array([1.0])
args = (1.0, 1.0)

W = dfx.VirtualBrownianTree(
    t0, t1, tol=1e-6, shape=(), key=jr.PRNGKey(2), levy_area=dfx.SpaceTimeLevyArea
)
terms = dfx.MultiTerm(dfx.ODETerm(drift), dfx.ControlTerm(diffusion, W))

print("Forwards-in-time")
solver = dfx.Heun()
sol_heun = dfx.diffeqsolve(terms, solver, t0, t1, dt0, y0, args)
print(f"Heun: {sol_heun.ys[-1][0]}")

solver = dfx.SlowRK()
sol_shark = dfx.diffeqsolve(terms, solver, t0, t1, dt0, y0, args)
print(f"SRK: {sol_shark.ys[-1][0]}")


print("Backwards-in-time")
solver = dfx.Heun()
sol_heun = dfx.diffeqsolve(terms, solver, t1, t0, -dt0, y0, args)
print(f"Heun: {sol_heun.ys[-1][0]}")

solver = dfx.SlowRK()
sol_shark = dfx.diffeqsolve(terms, solver, t1, t0, -dt0, y0, args)
print(f"SRK: {sol_shark.ys[-1][0]}")
```

Output:
```bash
Forwards-in-time
Heun: 1.2419108152389526
SRK: 1.2419666051864624
Backwards-in-time
Heun: 0.34244588017463684
SRK: -0.14933258295059204
```

### Fix
This is due to this line in `AbstractSRK`:
https://github.com/patrick-kidger/diffrax/blob/14baa1edddcacf27c0483962b3c9cf2e86e6e5b6/diffrax/_solver/srk.py#L355-L356
Here, `h` has the wrong sign. Instead, we should use the `dt` from `VirtualBrownianTree`:
```python
# Brownian increment (and space-time Lévy area)
bm_inc = diffusion.contr(t0, t1, use_levy=True)

# time increment
h = bm_inc.dt
```

With this change, the output becomes:
```bash
[...]

PR #746, opened 2026-04-10
Fix backwards-in-time solve for SRKs

sammccallum:
The `AbstractSRK` methods produced incorrect solutions when solving backwards-in-time. The fix proposed in #598 and #599 got the methods to converge but with incorrect strong orders. This PR gets the methods back to their correct strong orders.

 The two key points are:
 - `AbstractSRK` methods have two internal times: unsigned `h` and signed return from `drift.contr(t0, t1)`
 - `WrapTerm` multiplies the Levy areas by `direction`, but the space-time Levy area `H` should remain unchanged
 
The `test_sde1.py` test has been edited to now check strong order for both forwards and backwards in time.

### DIFF
--- a/diffrax/_solver/srk.py
+++ b/diffrax/_solver/srk.py
@@ -360,8 +360,15 @@ class AbstractSRK(AbstractSolver[_SolverState]):
         else:
             ignore_stage_g = jnp.array(self.tableau.ignore_stage_g)
 
-        # time increment
+        # The internal time step, h, is always positive. The direction is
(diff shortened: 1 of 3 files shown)


sammccallum:
The dtype promotion was needed to two more places. Sorry about that, Patrick

patrick-kidger:
No worries! And merged 🎉 

Thanks for identifying the problem and the solution :) I appreciate you make this happen!

Issue #752, opened 2026-05-04
Nonlinear max steps reached error on implicit solvers with PID step controller


zhou-haoyang:
Hi!

I was using implicit solvers with a PID step size controller for solving stiff ODEs and I got the error of `equinox.EquinoxRuntimeError: The maximum number of steps was reached in the nonlinear solver. The problem may not be solveable (e.g., a root-find on a function that has no roots), or you may need to increase max_steps.` If I understand correctly, the PID controller should instead reject the step and try with a smaller step in this case.

Checking the code shown below, the integrator doesn't seem to check the `solver_result` returned by the solver, and the PID controller relies on `y_error=inf` as a sign of a failed implicit step. In my case, the root finder returns `max_steps_reached` but with a non-infinite `y_error`. This is probably the reason why the error got propagated to the final state.

https://github.com/patrick-kidger/diffrax/blob/7f0001c7959bd8cbc267809c112b4403fa416e35/diffrax/_integrate.py#L373-L405

https://github.com/patrick-kidger/diffrax/blob/7f0001c7959bd8cbc267809c112b4403fa416e35/diffrax/_step_size_controller/pid.py#L533-L538

In my case, applying this simple fix after `solver.step` seems to solve the problem, but I'm not sure if this is as intended.

```python
y_error = jtu.tree_map(lambda x: jnp.where(is_okay(solver_result), x, jnp.inf), y_error)
```

patrick-kidger:
So at least for Runge-Kutta methods then exactly the fix you have in mind is performed here:

https://github.com/patrick-kidger/diffrax/blob/7f0001c7959bd8cbc267809c112b4403fa416e35/diffrax/_solver/runge_kutta.py#L1194-L1197

Which solver are you using? It might be that we left this out somewhere else!

zhou-haoyang:
Ah that explains it. I was using implicit Euler, and it's missing there. I didn't run into this issue with the Runge-Kutta methods.

I can open a PR with a small patch if that sounds good to you :)

patrick-kidger:
Yes please! :)

zhou-haoyang:
Fixed in #753

PR #753, opened 2026-05-08
Fix y_error returned from implicit Euler step when non-linear solve fails

zhou-haoyang:
As discussed in #752, `ImplicitEuler` solver doesn't check the result returned from the non-linear solver and fails to return an infinite `y_error`, which prevents the stepsize controller from trying a smaller step. This PR fixes this issue by implementing the same logic in the `AbstractRungeKutta`.

If necessary, we could add a test on this logic for all implicit solvers.

### DIFF
--- a/diffrax/_solver/implicit_euler.py
+++ b/diffrax/_solver/implicit_euler.py
@@ -1,10 +1,10 @@
 from collections.abc import Callable
 from typing import ClassVar, TypeAlias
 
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Awesome, thank you for the fix. Merged! 🎉

zhou-haoyang:
Perfect, thank you as well!

PR #754, opened 2026-05-13
fix reporting nonconvergence as divergence for non-Cauchy termination

jpbrodrick89:
Fixes https://github.com/patrick-kidger/optimistix/issues/221, also see https://github.com/patrick-kidger/optimistix/pull/230. Note the step counts only reduce from 845 to 242 rather than 238 simply because at_least_two is now doing what it should.

Note that if we manage to progress https://github.com/patrick-kidger/optimistix/pull/214 at some stage (quite an undertaking) we should be able to remove the terminate copy paste issue between VeryChord and AbstractNewtonChord.


patrick-kidger:
LGTM! :)

PR #755, opened 2026-05-13
Enable setting of Jacobian tags in ODETerm

jpbrodrick89:
In light of the new colouring methods in lineax I really wanted to see if there was an unfootgunnable way for users to provide structural tags for their syste, and I think I might have found it.

**The challenge**

Providing tags to implicit solvers requires user's knowledge of the exact impicit relation being solved and perhaps careful consideration of whether the solve is fully-implicit (in which case a tridiagonal ODE term may be become pentadiagonal for a FIRK2) or direct implicit.Fortunately though, diffrax currently only has DIRK solvers (Implicit Euler is just a flavour of DIRK).

**The observation**

Each iteration of the nonlinear solver solves a linearised form of the implicit relation—this can be represented in terms of lineax operators and our existing tag propagation logic.

**The solution**

Users provide tags for the _ODE system_ Jacobian directly on to ODETerm, this should represent the structure of `jacobian(vf)(y)` and is completely **agnostic** of solver internals. Implicit solvers use lineax tag propagation to determine which tags remain relevant. For example, if a user declares a system to be negative semi-definite (e.g. a diffusion equation) then the current DIRK and implicit Euler solvers will recognised this as positive/negative semi-definite and pick Cholesky, however if the system is positive semi-definite (anti-diffusion) the implicit solvers will realise that residual jacobian is not guaranteed to have either form of semi-definiteness for arbitrary timestep and LU will be used instead.

**Remaining footguns**

The inherent footgun is not necessary completely removed, but the burden of complexity now lays solely at the feet of power users developing their own implicit solvers rather than the typical diffeqsolve users who can just plug in well designed implicit solvers and have it "just work". The main responsibility of implicit solver writers is to provide `implicit_op_relation` mirroring `implicit_relation`. Furthermore, this handling is opt-in from implicit solver designers, no mandatory (or optional) arguments are introduced into implicit solver design.

It would be really nice if we can get this in before the AD workshop and provide an easy heat equation example but completely understand if this is too complicated to find the time for in that window.

### DIFF
--- a/diffrax/_adjoint.py
+++ b/diffrax/_adjoint.py
@@ -1,7 +1,7 @@
 import abc
 import functools as ft
 import warnings
-from collections.abc import Callable, Iterable
+from collections.abc import Callable
 from typing import Any, cast
 
 import equinox as eqx
@@ -15,6 +15,7 @@ import optimistix.internal as optxi
 from equinox.internal import ω
 
 from ._heuristics import is_sde, is_unsafe_sde
+from ._misc import _frozenset
 from ._saveat import save_y, SaveAt, SubSaveAt
 from ._solver import (
     AbstractItoSolver,
@@ -450,15 +451,6 @@ if _vf.__globals__["__name__"].startswith("jaxtyping"):
     _vf = _vf.__wrapped__  # pyright: ignore[reportFunctionMemberAccess]
 if _solve.__globals__["__name__"].startswith("jaxtyping"):
     _solve = _solve.__wrapped__  # pyright: ignore[reportFunctionMemberAccess]
-
-
-def _frozenset(x: object | Iterable[object]) -> frozenset[object]:
-    try:
-        iter_x = iter(x)  # pyright: ignore
-    except TypeError:
-        return frozenset([x])
-    else:
-        return frozenset(iter_x)
 
 
 class ImplicitAdjoint(AbstractAdjoint):
--- a/diffrax/_solver/implicit_euler.py
+++ b/diffrax/_solver/implicit_euler.py
@@ -1,8 +1,10 @@
 from collections.abc import Callable
 from typing import ClassVar, TypeAlias
 
+import jax
 import jax.numpy as jnp
 import jax.tree_util as jtu
+import lineax as lx
 import optimistix as optx
 from equinox.internal import ω
 
@@ -11,7 +13,7 @@ from .._heuristics import is_sde
 from .._local_interpolation import LocalLinearInterpolation
 from .._root_finder import with_stepsize_controller_tols
 from .._solution import is_okay, RESULTS
-from .._term import AbstractTerm
+from .._term import AbstractTerm, WrapTerm
 from .base import AbstractAdaptiveSolver, AbstractImplicitSolver
 
 
@@ -22,6 +24,12 @@ def _implicit_relation(z1, nonlinear_solve_args):
     vf_prod, t1, y0, args, control = nonlinear_solve_args
     diff = (vf_prod(t1, (y0**ω + z1**ω).ω, args, control) ** ω - z1**ω).ω
     return diff
+
+
+def _implicit_op_relation(J: lx.AbstractLinearOperator) -> lx.AbstractLinearOperator:
+    # Mirrors `_implicit_relation`: residual(z) ≈ h·vf(y0 + z) - z, so the residual
+    # Jacobian is h·J - I. The scalar h is omitted (tag inference is sign-only).
+    return J - lx.IdentityLinearOperator(J.in_structure())
 
 
 class ImplicitEuler(AbstractImplicitSolver, AbstractAdaptiveSolver):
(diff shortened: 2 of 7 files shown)


jpbrodrick89:
@patrick-kidger just a friendly ping to see if you have had/might have a chance to think about this before the workshop next week. I realise this is an impactful change that might take some digestion, so no worries if not.

Note we can offload even more complexity and maintenance to lineax now with the new tags_from_checks helper. I could document that and do another lineax release and refactor this accordingly?

diffrax/_solver/implicit_euler.py
```diff
@@ -82,6 +90,13 @@ def step(
         # (C.f. `AbstractRungeKutta.step`.)
         # If we wanted FSAL then really the correct thing to do would just be to
         # write out a `ButcherTableau` and use `AbstractSDIRK`.
+        _inner = terms.term if isinstance(terms, WrapTerm) else terms
```
patrick-kidger, line 93: I don't think this is okay. Kind of the point of having an ABC is that we don't generally care exactly what we get, and can abstract over them.

(As a practical matter we do pierce this veil in a couple of places, but almost always just as special-case performance optimizations, not things that affect functional correctness.)

diffrax/_solver/implicit_euler.py
```diff
@@ -82,6 +90,13 @@ def step(
...
+        y_struct = jax.tree_util.tree_map(
+            lambda x: jax.ShapeDtypeStruct(x.shape, x.dtype), y0
+        )
+        residual_tags = self._residual_tags(
+            getattr(_inner, "tags", frozenset()), y_struct, _implicit_op_relation
```
patrick-kidger, line 98: I consider `getattr` to be pretty much banned in all code I write! 😄 The better choice is usually to set up an abstractmethod on the appropriate ABC, or to use an `isinstance` (depending on whether you want type or consumer to own the implementation).

diffrax/_adjoint.py
```diff
@@ -15,6 +15,7 @@
 from equinox.internal import ω
 
 from ._heuristics import is_sde, is_unsafe_sde
+from ._misc import _frozenset
```
patrick-kidger, line 18: nit, underscore indicates privacy to the file it's created in. (Is the convention we use in the Equinox ecosystem.)

patrick-kidger (commented):
I like the basic idea of deferring the tag-propagation problem to Lineax. That sounds like the right place for this problem to live, I think.

I am concerned that this is started to get a bit of a thorny stack to reason over: whilst many users can indeed just provide some tags, this really starts to raise the bar for what is required to be a power user. More and more subsystems that need addressing. (I see this a lot first-hand: I almost always take a look at the source code of the libraries I use, and many of them start to accrue a lot of special-cases that look a lot like this, that are largely inscrutable to the non-expert or non-author.)

I wonder if we should instead make it possible for a user to just say something like 'just use a Cholesky solver, trust me it'll work'.

(Honestly the whole tags system is never something I've been a fan of anyway...)

jpbrodrick89:
I'm tempted to for a both/and option by going with this PLUS a lineax environment LINEAX_CHECK_TAGS that defaults to True (for error checking and backwards compatibility) and allowing users to disable this if they can't be bothered to mess around with tags.  The trickiest thing though is Cholesky requires awareness of whether the Jacobian is nsd or psd and this would require the user diving into the internals (it depends whether they're using DIRK or implicit Euler) to check our sign conventions which is also a footgun that tag propagation would solve for them. Can you think of any clever solution to this? Possibly the easiest is to change the sign convention of Euler and then default Cholesky to assuming psd rather than nsd when LINEAX_CHECK_TAGS is false but there might be another non-breaking change.

jpbrodrick89:
I've been thinking more about this, I think I'd still prefer a both/and approach but I think the solution might be to offer a parallel scipy-like `assume_` framework provided to a solver at instatiation. I know this breaks your pattern about solvers not having knowledge of operator structure but I think this is the cleanest way. The neatest thing about this is we could have `assume_(semi)definite` and then Cholesky could calculate one (or more if necessary) inner product to probe its sign so we don't need to worry so much about tracking sign conventions. This would be quite a big lineax feature but think it may pay dividends, especially those coming from a scipy background, wdyt?

patrick-kidger:
So I think a strong preference from me to have just one system. (Even at the cost of a breaking change.) Having multiple really increases the difficulty of using the library, and forever invites discussions around the 'right' way to do something.

Fwiw if you really wanted to go the solver approach then this would be possible with something like what your describe: `AssumeTagged(solver, tags)`, that wraps an incoming operator into a `TaggedLinearOperator`, and then calls the underlying solver.

I think you could use the same trick for `options` actually too.

jpbrodrick89:
The `AssumeTagged` seems like a clean solution if we introduce it with an sign ambivalent `is_(semi)definite` tag.

Do you mind if I just probe a bit more what you mean by "a strong preference from me to have just one system", do you mean we should decide between ONLY using this PR's approach vs `AssumeTagged` vs a lineax overhaul? Personally my gut feel would be to do both of the first two (this PR + `AssumeTagged`) but appreciate the concern about avoiding discussions on the "right" way to do something (my soft pushback is `AssumeTagged` is good when you're confident the implicit solver mechanics guarantee this tag but this PR is better when you're not so sure). However, if you're really trying to say we should strongly consider a complete lineax overhaul perhaps gutting the tag system that is another matter entirely. I'm nervous and confused about this, but also open minded, so if this is what you're trying to say I'd appreciate hearing more of your case before we buy too strongly into going even deeper into the tag system in lineax (not necessarily by a huge amount, but to me it seems a nice way to encode properties about an operator at compile time that a solver could use).

patrick-kidger:
> Personally my gut feel would be to do both of the first two (this PR + AssumeTagged) but appreciate the concern about avoiding discussions on the "right" way to do something (my soft pushback is AssumeTagged is good when you're confident the implicit solver mechanics guarantee this tag but this PR is better when you're not so sure).

Yeah, I think that's reasonable. My pushback was really around the idea of having approaches like `LINEAX_CHECK_TAGS`, which I think is more a footgun than it is an escape hatch.

(I would note that the contents of this PR do seem a bit fragile, as per my commentary above.

As a second point, do you think it would be possible to offer a set of tag-propagation mechanisms in Lineax, corresponding to adding operators, etc. – also meaning that we don't need to update Diffrax if there are new tags.)

Issue #85, opened 2022-03-29
Diffrax can't seem to find a solution despite increasing max_steps


ma-sadeghi:
So far, I've tried using `float64`, and smaller `atol`/`rtol` values, but I get the same error. I also increased `max_steps` up to 16^6, but no luck.

To give a little bit of context, the system of ODEs I'm trying to solve here is method of lines discretization of the heat equation in 1D, initially at y = 1, with Dirichlet BCs imposed at the two boundaries.

To make sure that the problem is not ill-defined, I tried to solve it using `scipy`'s `solve_ivp` and it worked fine.

Would really appreciate it if you could shed some light.

```python
# Imports
import diffrax
import jax
import jax.numpy as jnp
jax.config.update("jax_enable_x64", False)

# Parameters
N = 1000
h = 1/(N+1)
y0 = jnp.ones(N, dtype="float32")    # IC: @t=0 ; y=1
y0 = y0.at[0].set(0.0)               # BC: @x=0 ; y=0
y0 = y0.at[0].set(0.0)               # BC: @x=L ; y=0
tspan = (0, 0.05)
saveat = jnp.linspace(*tspan, 10).tolist()
rtol = 1e-3
atol = 1e-5

# Define system of ODEs
def func(t, y, dy):
    idx = jnp.arange(1, N-1)
    dy = dy.at[idx].set((y[idx-1] - 2*y[idx] + y[idx+1]) / h**2)
    return dy

# Solve diffeq system using Diffrax
dy = jnp.zeros(N)
odefunc = lambda t, y, args: func(t, y, dy)
eqs = diffrax.ODETerm(odefunc)
solver = diffrax.Dopri5()
stepsize_controller = diffrax.PIDController(rtol=rtol, atol=atol)
t = diffrax.SaveAt(ts=saveat)

@jax.jit
def solve_diffrax(y0):
    return diffrax.diffeqsolve(eqs, solver, t0=tspan[0], t1=tspan[1], dt0=None,
                               y0=y0, saveat=t, max_steps=4*16**4,
                               stepsize_controller=stepsize_controller)

sol = solve_diffrax(y0)
```

```python
WARNING:absl:No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)
ERROR:absl:Outside call <function _call.<locals>.<lambda> at 0x7f5810336e50> threw exception The maximum number of solver steps was reached. Try increasing `max_steps`..
[...]

PR #758, opened 2026-06-02
stop clipping rejected steps to t1

### LINKED ISSUE #756: Infinite final-step rejection loop due to conflicting endpoint clipping and adaptative RK step rejection
Hi :)

I am using Diffrax through the Dynamiqs quantum simulation library, and I think I encountered an endpoint-handling issue in Diffrax.

This issue is similar but not the same mechanism as : [Issue #703](https://github.com/patrick-kidger/diffrax/issues/703#issue-3591397040) 

### Issue

I was investigating the stability of Tsit5() on a difficult quantum simulation where the PID controller converges to small time steps compared with the final integration time. The simulation runs fine for almost the whole interval, but at the very last step the solver gets stuck in what looks like an infinite rejection loop until max_steps is reached.

### Diagnosis of the issue 

Diffrax has a mechanism to avoid taking a very small last time step. If the proposed tnext is within the last 100 * ULP(t1) before the final time, then Diffrax treats this as close enough to the end and clips toward t1:

```
def _clip_to_end(tprev, tnext, t1, t1_clip_floor, keep_step):
    clip = tnext > t1_clip_floor
    tclip = jnp.where(keep_step, t1, tprev + 0.5 * (t1 - tprev))
    return jnp.where(clip, tclip, tnext)
```

I understand the motivation: if the solver is at

`tprev = t1 - epsilon`

with very small epsilon, then taking a tiny final step to exactly t1 may be numerically unstable, especially for dense output.

In my case, with `float32` and `t1 = 600.0`:

```
ULP(t1) ≈ 6.1e-5
100 * ULP(t1) ≈ 6.1e-3
PID-selected dt ≈ 3e-3
```

So the PID controller wants steps smaller than the final clipping window.

For example, suppose we are at:

`tprev ≈ 599.993
`
and the PID controller naturally proposes:

`tnext ≈ 599.996
`
This is already inside the interval:

`[t1 - 100 * ULP(t1), t1]
`
so Diffrax treats it as a near-final step. If it tries to go to t1, the effective final step becomes roughly:

`599.993 -> 600.0
`
which is about `0.007`, larger than what the PID controller wanted. The step is rejected.

Then Diffrax proposes the midpoint:

`tnext = 0.5 * (tprev + t1)
`
which is around:

`599.9965
`
but this is still inside the final 100-ULP window. So Diffrax again treats it as a near-final step, tries to finish, rejects, proposes another point still inside the window, and the loop repeats until max_steps is reached.

### Suggested fix

A possible fix, if you agree with the diagnosis, would be to make endpoint handling an option. The current behavior could remain the default, but another mode could be “overshoot and interpolate”:

**Instead of forcing the last numerical step to land exactly on t1, allow the solver to take one normal accepted step past t1, then evaluate the existing dense/local interpolation at t1.**

This would treat the final time like any other saved intermediate time. For example:

`9.9975 -> 9.9985 -> 9.9995 -> 10.005`

and then return:

`y(10.0)`

by evaluating the local interpolation on the accepted step `[9.9995, 10.005]`.

Would this endpoint mode make sense for Diffrax? If so, I would be happy to try opening a PR.

jpbrodrick89:
- Fixes #756 

`_clip_to_end` previously overrode the PID controller on a rejected step with a fixed midpoint that is also rejected; on rejection tprev doesn't advance, so the midpoint is constant, and the same too-large step gets retried forever.

The midpoint override's only justification was floating-point stability of dense interpolation against the old fixed 1e-6 tolerance. #660 replaced that with 100 * ULP(t1), which is directly related to the float resolution boundary. This PR just drops that branch and trusts the controller's shrunken proposal on rejection. The #85 fix (only snap-to-t1 on accepted steps) is preserved.

Regression coverage:
- _clip_to_end unit test pinning the three branches directly.
- End-to-end test: harmonic oscillator at t1=600 in float32 with rtol=1e-4 deterministically hangs at the clip floor (final_t ≈ 599.9937) on main and finishes in ~30 steps with this patch.

### DIFF
--- a/diffrax/_integrate.py
+++ b/diffrax/_integrate.py
@@ -275,13 +275,13 @@ def _save(
     )
 
 
-def _clip_to_end(tprev, tnext, t1, t1_clip_floor, keep_step):
+def _clip_to_end(tnext, t1, t1_clip_floor, keep_step):
     # The tolerance of ~100 ULP's means that we don't end up with too-small intervals
     # for dense output, which then gives numerically unstable answers due to floating
     # point errors.
-    clip = tnext > t1_clip_floor
-    tclip = jnp.where(keep_step, t1, tprev + 0.5 * (t1 - tprev))
-    return jnp.where(clip, tclip, tnext)
+    # Only clip on accepted steps: on a rejection the controller has just shrunk its
+    # proposal, so overriding it can cause an infinite rejection loop (see #756).
+    return jnp.where(keep_step & (tnext > t1_clip_floor), t1, tnext)
 
 
 def _maybe_static(static_x: ArrayLike | None, x: ArrayLike) -> ArrayLike:
@@ -410,7 +410,7 @@ def loop(
         #
 
         tprev = jnp.minimum(tprev, t1)
-        tnext = _clip_to_end(tprev, tnext, t1, t1_clip_floor, keep_step)
+        tnext = _clip_to_end(tnext, t1, t1_clip_floor, keep_step)
 
         progress_meter_state = progress_meter.step(
             state.progress_meter_state, linear_rescale(t0, tprev, t1)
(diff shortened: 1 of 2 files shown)


patrick-kidger:
LGTM! Merged 🎉

Issue #558, opened 2025-01-04
Adjoints question 


lockwo:
I was putting together some tests, when I realized that I'm not sure I fully understand the error bounds on the non recursive (which is what I almost always use) adjoints. Specifically, for a simple diagonal noise SDE, I actually encounter errors for the other adjoints, are these expected or am I doing something wrong?

```python
import jax
import jax.numpy as jnp
import jax.random as jr
import equinox as eqx
import diffrax
import lineax as lx

key = jax.random.PRNGKey(42)
key, subkey = jax.random.split(key)
driftkey, diffusionkey, ykey = jr.split(subkey, 3)
drift_mlp = eqx.nn.MLP(
    in_size=3,
    out_size=3,
    width_size=8,
    depth=2,
    activation=jax.nn.swish,
    final_activation=jnp.tanh,
    key=driftkey,
)
diffusion_mlp = eqx.nn.MLP(
    in_size=3,
    out_size=3,
    width_size=8,
    depth=2,
    activation=jax.nn.swish,
    final_activation=jnp.tanh,
    key=diffusionkey,
)
y0 = jr.normal(ykey, (3,))

k1, k2, k3 = jax.random.split(key, 3)
vbt = diffrax.VirtualBrownianTree(0.3, 9.5, 1e-4, (3,), k1, levy_area=diffrax.SpaceTimeLevyArea)
vbt_terms = diffrax.MultiTerm(
    diffrax.ODETerm(lambda t, y, args: drift_mlp(y)),
    diffrax.ControlTerm(lambda t, y, args: lx.DiagonalLinearOperator(diffusion_mlp(y)), vbt),
)
solver = diffrax.GeneralShARK()
y0_args_term0 = (y0, None, vbt_terms)

def _run(y0__args__term, saveat, adjoint):
    y0_, args, term = y0__args__term
    ys = diffrax.diffeqsolve(
        term,
        solver,
        0.3,
        9.5,
        0.1,
        y0_,
        args,
        saveat=saveat,
        adjoint=adjoint,
    ).ys
    return jnp.sum(ys)


t0 = True
t1 = True
ts = None
y0__args__term = y0_args_term0

saveat = diffrax.SaveAt(t0=t0, t1=t1, ts=ts)

inexact, static = eqx.partition(y0__args__term, eqx.is_inexact_array)

def _run_inexact(inexact, saveat_, adjoint_):
    return _run(eqx.combine(inexact, static), saveat_, adjoint_)

[...]

PR #761, opened 2026-06-20
Backsolve error

lockwo:
https://github.com/patrick-kidger/diffrax/issues/558

GeneralShark and spark will also error here even though they can technically work on general noise terms, so maybe this isn't the optimal way of catching it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ site/
 .idea/
 .venv/
 uv.lock
+.vscode/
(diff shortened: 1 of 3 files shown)


patrick-kidger:
In principle a `MultiTerm(ODETerm, ODETerm)` should be valid here, so agreed - ideally we'd have a better way of catching this.

Checking the original issue, IIUC an error does appear anyway? It's just not a very friendly one?

lockwo:
That's true, an error does appear, in theory the multiterm should be allowed, but in practice multiterm solver requirements cause a problem, maybe the better fix is to make sure those work to help streamline the error

Issue #438, opened 2024-06-09
Support for uncertainty propagation?


nstarman:
It would be amazing to be able to specify uncertainties on ICs and have that propagate through to the solutions.
I know this is a very difficult problem in general. Off the top of my head some challenges are:

1. Gaussian uncertainties are hard, but operations on arbitrary (e.g. non-symmetric) distributions often don't even have an analytic form. This could be approximated to 1st order...
2. What's the new API look like?
3. How to support dense solutions?

Thankfully I think at least point 2 has a workable solution. @patrick-kidger, you've written `quax` to allow for array-ish objects in JAX. My suggestion would be to make a `diffeqsolve(y0=)` accept `quax` classes that handle the distribution and its propagation.

Point 1 still remains hard, but there's a still-useful starting point.  The simplest "uncertainty" to support isn't even Gaussian but a simple lower and upper bound interval. That would be a good proof of concept but still useful! I know that the same result could be accomplished by doing `diffeqsolve` twice, but a) the unified API would be a convenience and b) we could hopefully subsequently implement Gaussian and more complex distributions. 

To use the opening example from https://docs.kidger.site/diffrax/usage/getting-started/

```python
from diffrax import diffeqsolve, Dopri5, ODETerm, SaveAt, PIDController
from diffrax import Interval

vector_field = lambda t, y, args: -y
term = ODETerm(vector_field)
solver = Dopri5()
saveat = SaveAt(ts=[0., 1., 2., 3.])
stepsize_controller = PIDController(rtol=1e-5, atol=1e-5)

sol = diffeqsolve(term, solver, t0=0, t1=3, dt0=0.1, y0=Interval(0.9, 1.1),  # note the Interval
                             saveat=saveat, stepsize_controller=stepsize_controller)

print(sol.ts)  # DeviceArray([0.   , 1.   , 2.   , 3.    ])
print(sol.ys)  # Interval(...)  # IDK about the internals
```

patrick-kidger:
I think this'd be really cool!

So as you've highlighted, Quax is the way to go about implementing something like this. The way this would work is to provide overloads for all primitives that `diffeqsolve` uses, and then call `quax.quaxify(diffrax.diffeqsolve)(y0=Interval(...), ...)`.

In particular the point of a library like Quax is that this shouldn't require changing Diffrax at all.

Providing all of those overloads is probably fairly ambitious (and might stress-test just how load-bearing Quax really is 😅), but if you do something like that I'd love to see it!

[...]

PR #767, opened 2026-08-19
Allocate the `SaveAt` buffer from the saved value, not its shape

nstarman:
Allocates the `SaveAt` output buffer *via* an operation on the value being saved, rather than conjuring it from that value's shape and dtype.

### The problem

`_allocate_output` builds the buffer with `jnp.full`:

```python
struct = eqx.filter_eval_shape(subsaveat.fn, t0, y0, args)
ys = jtu.tree_map(lambda y: jnp.full((out_size,) + y.shape, jnp.inf, dtype=y.dtype), struct)
```

`jnp.full` takes no operand. So when `y0` is an array-ish abstraction — a `quax.Value` carrying units, an uncertainty, a sparsity pattern — there is nothing here for Quax to dispatch on, and the buffer comes out as a plain array. `_save` then puts a slice of that plain buffer opposite the carried value in a `lax.cond`:

```python
y_to_save = lax.cond(
    pred,
    lambda: fn(t, y, args),                                  # the wrapped value
    lambda: jtu.tree_map(lambda ys_: ys_[save_index], ys),   # a plain buffer slice
)
```

The two branches disagree on structure, `cond_p` needs one, and the wrapper is dropped. `sol.ys` comes back as a bare array; a type that declines to materialise (units, say — a dimensionless metre has no meaning) raises instead.

### Why this one can't be fixed from outside

In #438 the conclusion was that Quax shouldn't require changing Diffrax, and for essentially all of `diffeqsolve` that holds — Quax intercepts primitives, and Diffrax is written in primitives. This site is the exception, and structurally so: a rule can only fire where there is an operand to dispatch on, and an array conjured from a shape has none. No set of Quax rules can reach it.

(The workaround in #438 — a `default` rule permissive enough to re-box *everything* — only appears to work because it also boxes the solver's integer loop counters, at which point almost every primitive has a boxed operand and the buffer gets re-boxed on first write. That is also why `sol.ts` came back wrapped there.)

### The change

Keep the `inf` fill exactly, but route it through a `select` whose predicate is a compile-time constant, so `y` is an operand:

```python
y_init = subsaveat.fn(t0, y0, args)
ys = jtu.tree_map(
    lambda y: jnp.where(
        True,
        jnp.full((out_size,) + jnp.shape(y), jnp.inf, dtype=jnp.result_type(y)),
        lax.stop_gradient(jnp.broadcast_to(y, (out_size,) + jnp.shape(y))),
    ),
    y_init,
)
```

XLA folds the select away and DCEs the dead branch — including the `subsaveat.fn` call feeding it.

The `stop_gradient` is load-bearing. Without it the buffer acquires a tangent path back to `y0` — zero-valued, but structurally present — and `BacksolveAdjoint` trips its own `nondifferentiable` guard: `RuntimeError: Unexpected tangent`. That showed up as 4 failures in `test_adjoint.py`. With `stop_gradient` all adjoints produce gradients identical to `main`.

### Cost

XLA cost analysis, 64-dim state, `Tsit5` + `PIDController(rtol=1e-6, atol=1e-8)`, `max_steps=4096`:

| `SaveAt` | flops before → after | bytes accessed | temp memory |
|---|---|---|---|
| `t1=True` | 3,267 → 3,263 | 14,113 → 14,113 | 3,392 → 3,392 |
| `steps=True` | 3,440 → 3,436 | 1,080,575 → 1,080,575 | 1,065,888 → 1,065,888 |
| `steps=True, fn=sin(y)**3+cos(y)` | 3,632 → 3,628 | 1,080,575 → 1,080,575 | 1,065,888 → 1,065,888 |

Four fewer flops, identical memory traffic. The non-trivial `fn` row is there to show the extra `subsaveat.fn` call costs nothing once the dead branch is eliminated.

### Behaviour

Unchanged. Saved values are bit-identical and unwritten slots still read back as `inf` — including `test_saveat_solution.py`'s explicit `jnp.allclose(ys1, jnp.array([0.25, 0.09375, jnp.inf]))`.

One caveat worth naming: `subsaveat.fn` is now invoked once during allocation. It was already invoked inside the solve, so this introduces no new *kind* of behaviour, but a `fn` with a side effect (a `jax.debug.print`, say) would fire one additional time, since DCE removes the pure computation but not the effect.

Two simpler forms that I tried and discarded, in case they come to mind on review:

- `jnp.broadcast_to(y_init, ...)` with no select at all. Changes the fill from `inf` to `y0`, which contradicts the assertions above, and separately broke `test_adjoint.py::test_against`.
- The select without the `stop_gradient`, which breaks `BacksolveAdjoint` as described.

### Test

`test/test_quax.py` asserts `sol.ys` keeps its wrapper across `t1` / `t0,t1` / `ts` / `steps`, that `sol.ts` does *not* pick one up, and that the `inf` fill survives. Verified against this branch: 4 failures before the change, all green after.

It is gated on `quax>=0.5.0` and so currently skips — `quaxify(diffeqsolve)` also needs Quax's `custom_vjp` support (Diffrax's buffered loops route through `jax.custom_vjp`), which lands in that release. Happy to drop the test and the test-group dependency if you would rather not carry a skipped test until then.

### Not fixed here

With the default `throw=True`, the `Solution` is passed through `equinox.internal.error_if`, which allocates its error branch from `jax.eval_shape` + `pure_callback` — the same shape-not-operand pattern, one library over. That erases the type again, so the test uses `throw=False`. Filed as patrick-kidger/equinox#1257, with a candidate patch.

### DIFF
--- a/diffrax/_integrate.py
+++ b/diffrax/_integrate.py
@@ -1294,10 +1294,21 @@ def diffeqsolve(
         saveat_ts_index = 0
         save_index = 0
         ts = jnp.full(out_size, direction * jnp.inf, dtype=time_dtype)
-        struct = eqx.filter_eval_shape(subsaveat.fn, t0, y0, args)
-        ys = jtu.tree_map(
-            lambda y: jnp.full((out_size,) + y.shape, jnp.inf, dtype=y.dtype), struct
-        )
+
+        # Allocate the buffer *via* an operation on the saved value, not from its
+        # shape: `jnp.full` conjures an array with no operand, so an array-ish
+        # abstraction wrapping `y0` (a `quax.Value`) is erased as soon as a saved
+        # value round-trips through it (`_save` slices it opposite `y` in a `cond`).
+        # The select's predicate is a compile-time constant, so XLA folds it away
+        # and DCEs the `subsaveat.fn` call feeding the dead branch. `stop_gradient`
+        # is required: without it the buffer gains a tangent path back to `y0` and
+        # `BacksolveAdjoint` trips its `nondifferentiable` guard.
+        def _alloc(y):
+            shape = (out_size,) + jnp.shape(y)
+            fill = jnp.full(shape, jnp.inf, dtype=jnp.result_type(y))
+            return jnp.where(True, fill, lax.stop_gradient(jnp.broadcast_to(y, shape)))
+
+        ys = jtu.tree_map(_alloc, subsaveat.fn(t0, y0, args))
         return SaveState(
             ts=ts, ys=ys, save_index=save_index, saveat_ts_index=saveat_ts_index
         )
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ tests = [
   "jaxlib",
   "optax",
   "pytest",
+  "quax",
   "scipy",
   "tqdm"
 ]
--- a/test/test_quax.py
+++ b/test/test_quax.py
@@ -1,0 +1,89 @@
+"""`diffeqsolve` under `quax.quaxify`.
+
+Quax dispatches on *operations*. An array conjured from a shape rather than
+from an operand -- `jnp.full(shape, ...)` -- gives it nothing to dispatch on,
+so a wrapped `y0` is silently erased the moment a saved value round-trips
+through the `SaveAt` buffer. These tests pin the buffer as an operation on
+the value being saved.
+"""
+
+from collections.abc import Sequence
+
+import diffrax
+import equinox as eqx
+import jax
+import jax.numpy as jnp
+import pytest
+from jax.core import ShapedArray
+from jax.extend.core import Primitive
+
+
+# `quaxify(diffeqsolve)` additionally needs quax's `custom_vjp` support, which
+# diffrax's buffered loops go through; that landed in quax 0.5.0.
+quax = pytest.importorskip("quax", minversion="0.5.0")
+
+
+class Boxed(quax.ArrayValue):
+    """Propagates itself through any primitive it is an operand of.
+
+    Boxes *inexact* outputs only. Boxing integers too would put the box on the
+    solver's own loop counters and save indices, at which point essentially
+    every primitive has a boxed operand and the buffer gets re-boxed on its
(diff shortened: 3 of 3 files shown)


patrick-kidger:
Sorry for the delay, gosh.

Just so that I understand this and https://github.com/patrick-kidger/equinox/issues/1257 correctly: is your goal to introduce a `lax.select_n_p.bind(predicate, raw_array, quax_value)`, which then rules like 

https://github.com/nstarman/quax/blob/9ae62578fc877c7098c068d1d316982eea2346f4/src/quax/examples/unitful/_core.py#L125-L126

can catch on? And then you write your Quax rule to be okay with receiving an array/quax combo, and to create its output as a Quax Value? Let me know if I've understood that correctly or if I'm misunderstanding.

nstarman:
Yes, that's the gist.
As a note, `jnp.where(True, fill, y)` lowers to `select_n(pred, y, fill)`, so the Value is case 0 and the buffer is case 1. And we also need to dispatch for `broadcast_in_dim` and `stop_gradient`.

On whether rules can catch it: the linked `select_n_unitful(which, *cases: Unitful)` doesn't match `(Unitful, Array)`, so it falls to `Value.default` -> `materialise` -> `Unitful` raises. With these two, the alloc comes back as `Unitful(inf-buffer, {m: 1})`:

```python
@quax.register(jax.lax.stop_gradient_p)
def _(x: Unitful, **kw) -> Unitful:
    return Unitful(jax.lax.stop_gradient_p.bind(x.array, **kw), x.units)

@quax.register(jax.lax.select_n_p)
def _(which: ArrayLike, case0: Unitful, case1: ArrayLike, **kw) -> Unitful:
    return Unitful(jax.lax.select_n_p.bind(which, case0.array, case1, **kw), case0.units)
```

So yeah it's array/Value combo in, Value out.

The part I'm least happy about is that these dispatches are ill-typed on its face. What we actually want is a *construction* but quax dispatches on operations only, so this isn't possible in `diffrax` without some creative use of `map` operations. Currently, a select with a constant predicate is the smallest way to set up a construction as an operation. An alternative is a `full_like` hook quax side, which I think diffrax would have to use.

PR #768, opened 2026-09-02
Fix docstring formatting

simonaltrogge:
Add missing hyphen before `store_rejected_steps` argument in docstring of `ClipStepSizeController`'s constructor.


patrick-kidger:
Thank you!

Issue #738, opened 2026-02-26
Support for Sharding 2.0 2.0


wittenator:
As a follow up to #735 I wanted to report/pose another issue with Jax sharding and diffrax. While passing the mesh to the shard_map and setting `check_vma=False` works for the example, automatically inferring the mesh and unsetting the VMA check fails with the following error:

```python
import functools as ft

import jax
import jax.numpy as jnp
from diffrax import ODETerm, Tsit5, diffeqsolve
from jax import shard_map, sharding
from jax.experimental import mesh_utils

mesh = sharding.Mesh(mesh_utils.create_device_mesh((2,)), ["i"])
spec = sharding.PartitionSpec("i")

with jax.set_mesh(mesh):

    @jax.jit
    @ft.partial(shard_map, mesh=mesh, in_specs=spec, out_specs=spec, check_vma=False)
    @jax.vmap
    def run(y0):
        term = ODETerm(lambda t, y, args: -y)
        solver = Tsit5()
        t0 = 0
        t1 = 1
        dt0 = 0.1
        sol = diffeqsolve(term, solver, t0, t1, dt0, y0)
        return sol.ys

    y0 = jnp.array([10.0, 10.0])
    y0 = jax.device_put(y0, spec)
    out = run(y0)
    print(out)
    jax.debug.visualize_array_sharding(out)
```

The stack trace looks as follows:
```
Traceback (most recent call last):
  File "/workspaces/gfdm-jax/scripts/test.py", line 28, in <module>
    out = run(y0)
          ^^^^^^^
  File "/workspaces/gfdm-jax/scripts/test.py", line 23, in run
    sol = diffeqsolve(term, solver, t0, t1, dt0, y0)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/gfdm-jax/.venv/lib/python3.12/site-packages/diffrax/_integrate.py", line 1456, in diffeqsolve
    final_state, aux_stats = adjoint.loop(
                             ^^^^^^^^^^^^^
  File "/workspaces/gfdm-jax/.venv/lib/python3.12/site-packages/equinox/_module/_prebuilt.py", line 34, in __call__
    return self.__func__(self.__self__, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #770, opened 2026-09-19
remove ensure_compile_time_eval because hlo shows it's already optimized

spraharsh:
Addresses #738 

The main issue comes up in the evaluation of
```python
with jax.ensure_compile_time_eval():
    return is_successful(result) | is_event(result)
```
This also shows up in https://github.com/jax-ml/jax/discussions/31461 where `jax.ensure_compile_time_eval` doesn't play well with sharding.

In this particular case however `ensure_compile_time_eval()` makes no difference to optimized code. to reproduce that, run,

```python

# check_hlo.py
import functools
import re
import sys
from contextlib import nullcontext
from pathlib import Path

import diffrax as dfx
import jax
import numpy as np
from jax.sharding import Mesh, NamedSharding, PartitionSpec as P

label = sys.argv[1]  # before or after
assert label in ("before", "after")

mesh = Mesh(np.array(jax.devices("cpu")[:2]), ("i",))
spec = P("i")


@jax.jit
@functools.partial(
    jax.shard_map,
    mesh=mesh,
    in_specs=spec,
    out_specs=spec,
    check_vma=False,
)
@jax.vmap
def run(y0):
    return dfx.diffeqsolve(
        dfx.ODETerm(lambda t, y, args: -y),
        dfx.Tsit5(),
        t0=0,
        t1=1,
        dt0=0.1,
        y0=y0,
    ).ys


# The original fails with the ambient context, so use its working baseline.
context = jax.set_mesh(mesh) if label == "after" else nullcontext()
with context:
    y0 = jax.device_put(
        np.array([10, 20], dtype=np.float32),
        NamedSharding(mesh, spec),
    )
    compiled = run.lower(y0).compile()
    hlo = compiled.as_text()

Path(f"{label}.raw-hlo.txt").write_text(hlo)

# Retain all computation bodies; exclude header/source-location tables.
start = re.search(r"^(?:%|ENTRY %)", hlo, re.MULTILINE).start()
instructions = hlo[start:]

# Remove instruction debug metadata; retain all other attributes.
instructions = re.sub(r", metadata=\{[^\n]*?\}", "", instructions)
Path(f"{label}.instructions.txt").write_text(instructions)

```

with 

`JAX_PLATFORMS=cpu JAX_NUM_CPU_DEVICES=2 PYTHONPATH=. python check_hlo.py before`

before the patch is applied (Note that before does not have a mesh set, setting a mesh leads to the compilation error described). Then run

`JAX_PLATFORMS=cpu JAX_NUM_CPU_DEVICES=2 PYTHONPATH=. python check_hlo.py after
diff -u before.instructions.txt after.instructions.txt`

after the patch is applied. The instructions show no difference


spraharsh:
The original reason `ensure compile_time_eval` was added was to optimize code for "benchmarks/against_scan.py" in [this commit ](https://github.com/patrick-kidger/diffrax/commit/020942de32ab919182ca8a85351cd7a207af4021). I double checked that removing it does not change emitted HLO

Run

```python
import re
import sys
from pathlib import Path

import jax

label = sys.argv[1]  # before or after

# Load the benchmark definitions without running its speed tests.
path = Path("benchmarks/against_scan.py")
source = "\n".join(
    line for line in path.read_text().splitlines()
    if not line.startswith("speedtest(")
)
benchmark = {}
exec(compile(source, str(path), "exec"), benchmark)

compiled = jax.jit(benchmark["dfx_fn"]).lower(
    benchmark["fields"], benchmark["ts"]
).compile()

hlo = compiled.as_text()
Path(f"{label}.raw-hlo.txt").write_text(hlo)

# Exclude module header/source tables and instruction debug metadata.
start = re.search(r"^(?:%|ENTRY %)", hlo, re.MULTILINE).start()
instructions = re.sub(r", metadata=\{[^\n]*?\}", "", hlo[start:])
Path(f"{label}.instructions.txt").write_text(instructions)
```

before the patch with

```
JAX_PLATFORMS=cuda XLA_PYTHON_CLIENT_PREALLOCATE=false PYTHONPATH=. \
python check_benchmark_hlo.py before
```

. Then run

```
JAX_PLATFORMS=cuda XLA_PYTHON_CLIENT_PREALLOCATE=false PYTHONPATH=. \
python check_benchmark_hlo.py after

diff -u before.instructions.txt after.instructions.txt
```

There should be no difference in instructions

